PR #2, opened 2015-04-15
Correct the stdio redirection example

lydell:
Actually use stidin redirection instead of piping. Think of the kittens!


ben-eb:
That's pretty nice. But can we keep the whitespace around `>` and `<`?

```
cssnano < test.css > test.min.css
```

lydell:
Done.

ben-eb:
Thanks. :smiley:

PR #10, opened 2015-05-02
refactor index, closes #9

### LINKED ISSUE #9: refactor index.js
what about this refactor, it looks a lot better and cleaner.
also fixes possible thrown `TypeError: Cannot read property 'inline' of undefined` from [L57](https://github.com/ben-eb/cssnano/blob/master/index.js#L57)

```js
'use strict';

var Postcss = require('postcss');

var processors = {
  'postcss-discard-comments': 'comments',
  'postcss-zindex': null,
  'postcss-discard-empty': null,
  'postcss-minify-font-weight': null,
  'postcss-convert-values': null,
  'postcss-calc': 'calc',
  'postcss-colormin': null,
  'postcss-pseudoelements': null,
  './lib/filterOptimiser': null,
  './lib/longhandOptimiser': null,
  'postcss-minify-selectors': null,
  'postcss-single-charset': null,
  // font-family should be run before discard-font-face
  'postcss-font-family': null,
  'postcss-discard-font-face': null,
  'postcss-normalize-url': 'urls',
  'postcss-reduce-idents': 'idents',
  './lib/core': null,
  // Optimisations after this are sensitive to previous optimisations in
  // the pipe, such as whitespace normalising/selector re-ordering
  './lib/borderOptimiser': null,
  'postcss-discard-duplicates': null,
  './lib/functionOptimiser': null,
  'postcss-merge-rules': 'merge'
};
// var namespaces = {
//   'postcss-discard-comments': 'comments',
//   'postcss-zindex': 'zindex',
//   'postcss-calc': 'calc',
//   'postcss-normalize-url': 'urls',
//   'postcss-reduce-idents': 'idents',
//   'postcss-merge-rules': 'merge'
// };

module.exports = function cssnano(css, options) {
  if (typeof css === 'object') {
    options = css;
    css = null;
  }
  options = options || {};
  options = typeof options === 'object' ? options : {};
  var postcss = Postcss();
  var plugins = Object.keys(processors);
  var len = plugins.length;
  var i = 0;

  while (i < len) {
    var plugin = plugins[i++];
    var ns = processors[plugin];
    options = ns && options[ns] ? options[ns] : options;

    if (options.disable) {
      continue;
    }
    postcss.use(require(plugin)(options));
  }

  if (typeof css === 'string') {
    var result = postcss.process(css, options);
    // return a css string if inline/no sourcemap.
    if (options.map && options.map.inline || !options.map) {
      return result.css;
    }
    // otherwise return an object of css & map
    return result;
  }

  return postcss;
};
```

tunnckoCore:
closes #9 
dammit, continue to fails... and this tape + tap-spec :laughing:

can you investigate why? there's no logic for that when all other tests work.

i think it isn't from `postcss-conver-values` as you mentioned, cuz i review its new commits.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,67 +1,75 @@
 'use strict';
 
-var postcss = require('postcss');
+var Postcss = require('postcss');
 
-function createProcessor (module, namespace) {
-    return function (options) {
-        var opts = options && namespace ? options[namespace] : opts;
-        var ns = require(module)(opts);
-        ns._disabled = opts === false;
-        return ns;
-    };
-}
-
-var pipeline = {
-    discardComments: createProcessor('postcss-discard-comments', 'comments'),
-    zindex: createProcessor('postcss-zindex', 'zindex'),
-    discardEmpty: createProcessor('postcss-discard-empty'),
-    minifyFontWeight: createProcessor('postcss-minify-font-weight'),
-    convertValues: createProcessor('postcss-convert-values'),
-    calc: createProcessor('postcss-calc', 'calc'),
-    colormin: createProcessor('postcss-colormin'),
-    pseudoElements: createProcessor('postcss-pseudoelements'),
-    filterOptimiser: createProcessor('./lib/filterOptimiser'),
-    longhandOptimiser: createProcessor('./lib/longhandOptimiser'),
-    minifySelectors: createProcessor('postcss-minify-selectors'),
-    singleCharset: createProcessor('postcss-single-charset'),
-    // font-family should be run before discard-font-face
-    fontFamily: createProcessor('postcss-font-family'),
-    discardFontFace: createProcessor('postcss-discard-font-face'),
-    normalizeUrl: createProcessor('postcss-normalize-url', 'urls'),
-    reduceIdents: createProcessor('postcss-reduce-idents', 'idents'),
-    core: createProcessor('./lib/core'),
-    // Optimisations after this are sensitive to previous optimisations in
-    // the pipe, such as whitespace normalising/selector re-ordering
-    borderOptimiser: createProcessor('./lib/borderOptimiser'),
-    discardDuplicates: createProcessor('postcss-discard-duplicates'),
-    functionOptimiser: createProcessor('./lib/functionOptimiser'),
-    mergeRules: createProcessor('postcss-merge-rules', 'merge')
+var processors = {
+  'postcss-discard-comments': 'comments',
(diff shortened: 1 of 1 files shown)


tunnckoCore:
also exposed new option `disable` for each of the components.

```js
// disable calc
cssnano(input, {calc: false});

// disable calc
cssnano(input, {calc: {
  // some other options
  disable: true
}})
```

ben-eb:
I need to come up with a decent way of rebuilding the integration tests. For now you could use this snippet (just change the test name), but I can do a CLI tool eventually when I get some time.

```js
var nano = require('../');
var directory = require('fs').readdirSync;
var file = require('fs').readFileSync;
var formatter = require('./util/formatter');
var path = require('path');
var base = path.join(__dirname, 'integrations');

function formatted (css) {
    return postcss().use(nano()).use(formatter).process(css).css;
}

console.log(formatted(file(base + '/skeleton.fixture.css')));
```

The reason the tests were failing is because npm installed the newer versions of cssnano's modules, so it actually performed better minification as a result. This means the tests fail because the output has changed. :smile:

ben-eb:
I just realised you used 2 space indentation, in future could you respect the coding style and use 4?

https://github.com/ben-eb/cssnano/blob/master/.editorconfig

tunnckoCore:
oh yea, yea i forgot it. this is because i playing with refactor outside of fork, just in one simple dir ;d

Can I ask why 4 spaces? 90% of the community is 2 spaces and etc.

tunnckoCore:
beside that, thanks for merge :)

ben-eb:
It's a personal preference, I think that 4 spaces is more readable than 2. :smile:

tunnckoCore:
>  is more readable

absolutely not. its awful and anger people when reading. :palm_tree:

ben-eb:
Let's not have an argument about coding style. We could be here for hours. :wink:

tunnckoCore:
:+1:

TrySound:
@ben-eb I like tabs! :)

tunnckoCore:
yea yea 4 tabs is awesome awfulness :+1: :laughing:

tunnckoCore:
- single quotes
- 2 spaces
- return earlier (no else's or else-if's)

so easy and pretty rules

TrySound:
@tunnckoCore One tab. It could be 2 and 4 width. Everybody use own width.

PR #35, opened 2015-07-22
Refactor opts

HMUDesign:
I needed the ability to disable particular processors (specifically, convertValues was breaking some things). I refactored the opts usage to support plugin-specific options overriding namespace-specific options.

### DIFF
--- a/index.js
+++ b/index.js
@@ -41,23 +41,37 @@ var cssnano = Postcss.plugin('cssnano', function (options) {
     var plugins = Object.keys(processors);
     var len = plugins.length;
     var i = 0;
-
+    
     while (i < len) {
         var plugin = plugins[i++];
         var processor = processors[plugin];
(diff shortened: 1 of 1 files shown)


ben-eb:
So this would enable using the plugin's name as a value in the options config?

HMUDesign:
Exactly. Specifying the plugins name would take precedence over specifying
the namespace.

ben-eb:
I must admit this is a nicer solution going forward; we can perhaps drop the namespace altogether so that it is easier to document, in future. But could you attach a test case to the pull request please?

ben-eb:
Decided to go a different way on this. https://github.com/ben-eb/cssnano/commit/4c3553a14b6b26d1df9ae7a2ef4d26d61fcfd799

Thanks for providing the base of the new processor handling. :+1:

PR #62, opened 2015-09-18
Fix value parsing and striping spaces around every slash

TrySound:


lib/functionOptimiser.js
```diff
@@ -17,15 +17,29 @@ var functions = [
     'var'
 ];
 
+function walkTrim(nodes) {
```
ben-eb:
```js
function walkTrim (nodes) {
```

TrySound: Add eslint please))

PR #66, opened 2015-09-21
Tidy up

TrySound:


TrySound:
@ben-eb I do not use `0 0` value

TrySound:
@ben-eb I just replaced `none` with `0`

TrySound:
@ben-eb In future we need to move it to merge-longhand maybe

ben-eb:
Yeah, none to 0 is what I was talking about. :smile: 

https://github.com/jakubpawlowicz/clean-css/issues/226

TrySound:
@ben-eb But why you do the same with outline?

ben-eb:
Thanks! :+1:

PR #73, opened 2015-10-05
Add postcss-reduce-transforms. Fixes #72.

### LINKED ISSUE #72: Add postcss-reduce-transforms.
https://github.com/ben-eb/postcss-reduce-transforms

ben-eb:


TrySound:
@ben-eb Is transformZ equal transform3D as hack?

TrySound:
@ben-eb How much this save?

ben-eb:
My understanding is that `translate3d(0, 0, tz)` is equivalent to `translateZ(tz)`, therefore the `translate3d(0,0,0)` trigger for hardware acceleration also works with `translateZ(0)`.

Further details:

* http://blog.teamtreehouse.com/increase-your-sites-performance-with-hardware-accelerated-css
* https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function

PR #85, opened 2015-10-31
Resolve a crash when using the safe option via wrapper plugin.

ben-eb:
Ref: https://github.com/ben-eb/cssnano/issues/84

I think it might be fine to re-assign the property internally, instead of changing the name.

### DIFF
--- a/index.js
+++ b/index.js
@@ -65,14 +65,19 @@ var safeOptions = {
         keyframes: false
     },
     postcssZindex: {
-      disable: true
+        disable: true
     }
 };
 
 var cssnano = postcss.plugin('cssnano', function (options) {
+    if (options && options.safe) {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -65,14 +65,19 @@ var safeOptions = {
...
 var cssnano = postcss.plugin('cssnano', function (options) {
+    if (options && options.safe) {
+        options.safeMode = options.safe;
+        options.safe = null;
+    }
+
     options = options || {};
 
-    var safe = options.safe === true;
+    var safe = options.safeMode === true;
```
MadLittleMods: `isSafe`

ben-eb: :+1:

I prefer `isSafe` too.

PR #92, opened 2015-12-06
Add postcss-minify-gradients

ben-eb:
New optimisation to reduce gradient parameters where possible. :smile:

https://github.com/ben-eb/postcss-minify-gradients

/cc @TrySound


TrySound:
lgtm

PR #185, opened 2016-05-11
Added: Reduce transition timing functions.

ben-eb:
Further reading: https://drafts.csswg.org/css-transitions/#transition-timing-function-property


marcustisater:
Agreed, this should definitely go into the release.

PR #188, opened 2016-05-16
Rework reduce-positions code.

ben-eb:


ben-eb:
Note that this file is 100% covered, so it is a reduction in the overall lines of code in the codebase that is causing the code coverage to fail.

PR #258, opened 2016-08-29
Update autoprefixer to ^6.4.0

jesstelford:
See [`autoprefixer`'s changelog](https://github.com/postcss/autoprefixer/blob/master/CHANGELOG.md#64-hic-et-ubique-terrarum) for more info.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
   ],
   "license": "MIT",
   "dependencies": {
-    "autoprefixer": "^6.3.1",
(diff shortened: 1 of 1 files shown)


TrySound:
@jesstelford caret (`^`) includes the latest minor version. Just reinstall your node_modules

jesstelford:
Right you are! I was having problems getting in the latest version, turns out I had poisoned my local `node_modules` folder with an older hard coded version. Thanks for taking the time :)

Issue #268, opened 2016-09-02
Safe option not persistent for multiple files


techmatt101:
I'm using the gulp-cssnano tool and have found that passing safe flag as true only persists for the first file that is streamed through. Noticed it after updating to v3.7.4 and have tracked down the change to cda2edbbe54d906be86c157fed5dceabd256e24c commit. Looks to me that the `options.safe = null;` is then clearing the safe flag option for the next file that is piped through cssnano. 

Any chance this change could be reverted to fix multiple files not all using safe option?
I hopefully plan to write a test for this scenario tomorrow.

ben-eb:
This behaviour was initially introduced in https://github.com/ben-eb/cssnano/releases/tag/v3.3.2, as PostCSS will throw if you pass a `safe` flag (it's to do with the removal of the old `safe` option in PostCSS, which has been extracted out into postcss-safe-parser). We already have a [test case](https://github.com/ben-eb/cssnano/blob/cb1cf571c2464d31f1ecf8d85917664822f1c9b8/src/__tests__/api.js#L57) for this.

So we can't remove setting of this parameter to `null` but we could use another workaround; a test would be really helpful. 😄

ben-eb:
Closed by https://github.com/ben-eb/cssnano/issues/277.

PR #277, opened 2016-09-17
fixed safe option not persisting for multiple files

techmatt101:
When processing multiple files with safe flag set as true, the first time it will set the safe options as null due to PostCSS throwing an error if passed in. This then means the following files will not process with the safe flag as enabled. Instead it will now set isSafe as true on the options config for the next processed files to persist the the safe option.

Fix for issue: #268


ben-eb:
Thanks! 👍

PR #313, opened 2016-12-08
`inline list-item` -> `inline-list-item`

mattbasta:
Seems to be a typo. The tests are based on the mappings in this file, so it was uncaught. This is per the CSS3 display module spec:

https://drafts.csswg.org/css-display/#the-display-properties


ben-eb:
Hi @mattbasta,

Looks like a typo indeed! Nice catch, thanks! 👍

PR #331, opened 2016-12-31
Add a global browsers option.

ben-eb:
Ref: https://github.com/ben-eb/cssnano/issues/297

@ai What do you think of this implementation? This sets the stage for each optimisation to take a list of supported browsers, then they can use caniuse-db if they need to provide different logic for different browser versions.

### DIFF
--- a/docs/content/options.md
+++ b/docs/content/options.md
@@ -7,6 +7,22 @@ layout: BasicPage
 
 See the [optimisations](/optimisations/) section.
 
+## options.browsers (array|string)
+
+Set a browser scope which will be used by cssnano to determine which
+optimisations are safe to perform, using [browserslist]. You can use this option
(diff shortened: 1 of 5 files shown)


ben-eb:
Hmm, that might be kind of problematic due to the way that cssnano works right now; it doesn't actually have access to the `Result` because it returns a separate `Processor` - https://github.com/ben-eb/cssnano/blob/fa0580d5df958ccc8ffb33576a73f9c8610ab0bf/src/index.js#L132-L174

ben-eb:
After having a chat with @ai about the way Browserslist works, I think its better to not have a `browsers` option at all. Instead, this config will be loaded from the environment, either via a config file or environment variable, which is a much better way of handling it and is consistent with the other tools that use Browserslist.

But, we should probably document this somewhere. 😄

PR #336, opened 2017-01-09
Update index.html

Ne-Ne:
Incorrect npm reference?


ben-eb:
Hi @Ne-Ne,

No, `apm` is Atom's package manager. https://github.com/atom/apm

PR #368, opened 2017-05-18
Update postcss-discard-duplicates to latest

ismail-syed:
v2.0.2 of [`postcss-discard-duplicates`](https://github.com/ben-eb/postcss-discard-duplicates/blob/master/CHANGELOG.md#202) comes with a major performance boost. Since then `v2.1.0` has been released, so I've just updated `postcss-discard-duplicates` to the latest. 

This will subsequently provide performance boosts for consuming project, such as [`css-loader`](https://github.com/webpack-contrib/css-loader/blob/master/package.json#L17).

Also see https://github.com/angular-fullstack/generator-angular-fullstack/issues/2342#issuecomment-260291291

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "postcss-colormin": "^2.1.8",
     "postcss-convert-values": "^2.3.4",
     "postcss-discard-comments": "^2.0.4",
-    "postcss-discard-duplicates": "^2.0.1",
+    "postcss-discard-duplicates": "^2.1.0",
     "postcss-discard-empty": "^2.0.1",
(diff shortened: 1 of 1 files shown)


ben-eb:
The `^` (semver caret) means that a fresh install will pull down the latest version, so thanks for the contribution but it isn't necessary. 😄

ismail-syed:
Oh yes, that's right. 😓 
My `yarn.lock` seems to be expecting older versions. Could be another consuming module explicitly setting the versions.

ben-eb:
All going well the next major should be out within a couple of weeks anyway. 👍

PR #382, opened 2017-07-01
Fix resolveConfig for Root without source

darthmaim:
`fromFile` was trying to access the `Root` nodes `source` property, which is not always set (for example when creating it with `postcss.root()`.

This just adds an additional check for `css.source` in `fromFile`.

### DIFF
--- a/packages/cssnano/src/index.js
+++ b/packages/cssnano/src/index.js
@@ -31,7 +31,7 @@ function initializePlugin (plugin, css, result) {
 }
 
 function fromFile (css, result) {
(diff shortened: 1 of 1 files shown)


tivac:
This is causing issues for `modular-css` when using `cssnano@4` 😞

ben-eb:
@tivac Yep, planning to cut alpha 2 when I've done a bit more for postcss-selector-parser this week, this will be included.

tivac:
Thanks @ben-eb! :taco:

PR #384, opened 2017-07-12
Add css-declaration-sorter

Siilwyn:
Finally got the time to look into this, I see cssnano 4.0 is on it's way and the monorepo is looking nice!

Think the sorting is safe to put in the default preset, the only conflict I can think of is two identical declaration properties that are in a particular order for fallback. Since the module uses stable sorting identical properties will keep their order in regard to each other so that shouldn't be a problem.

Can't find how external modules are tested, should there be a new test?


andyjansson:
Considering that we do some optimisations by replacing the declarations, maybe sorting the declarations should be the last thing we do, rather than the first?

Siilwyn:
Good idea, I naively added it in alphabetical order. :)

ben-eb:
I would prefer that the options key would be consistent with the others, e.g. remove postcss-/cssnano- prefix and camel case the rest. I know it's a slightly different case because it's a custom name but it would be much more preferable to follow the existing format:

http://cssnano.co/guides/presets/#options-syntax

Also I think the tests might be failing due to the new logic so you might need to rebuild them. Sorry, no docs on that right now, I'm thinking that these should be converted over to snapshot tests in time. But you can rebuild by:

```
babel-node util/rebuildFrameworks.js
```

Siilwyn:
Discovered an unhandled promise rejection, it seems like it expects a function where there is none. My module returns a PostCSS plugin that returns a function, just like the other packages by the looks of it.

Getting this error a dozen of times when running `util/rebuildFrameworks.js`:
```
TypeError: plugin is not a function
    at initializePlugin (/home/selwyn/codeground/cssnano/packages/cssnano/dist/index.js:44:32)
```
which is: `return Promise.resolve(plugin()(css, result));` in `initializePlugin`.

Ideas?

packages/cssnano-preset-default/src/index.js
```diff
@@ -82,6 +82,7 @@ export default function defaultPreset (opts = {}) {
         [postcssMergeRules, options.mergeRules],
         [postcssDiscardEmpty, options.discardEmpty],
         [postcssUniqueSelectors, options.uniqueSelectors],
+        [cssDeclarationSorter, options.cssDeclarationSorter]
```
ben-eb: There's no trailing comment here. 😉

Siilwyn: Sharp!

Siilwyn:
Got the plugin is not a function errors away. :)

ben-eb:
Well, your plugin affects declaration order within a rule and because the tests are expecting declarations to be in a certain order, you can see why that would fail. Some of the tests do involve testing a plugin with all other plugins loaded. So I think yes, you will have to correct fixtures as I mentioned with the rebuild function, some of them unfortunately are manual, but I wrote them when snapshot testing wasn't available in AVA. I don't really have the patience anymore to convert them.

Siilwyn:
Thanks for the confirmation @ben-eb! The only two failing tests now are the preset framework tests, running `util/rebuildFrameworks.js` doesn't do anything. Something I missed?

ben-eb:
From the failure it looks like a problem still with the integration tests, could you rebase off master and try running `babel-node util/rebuildFrameworks.js` again please? 😄

Siilwyn:
@ben-eb for some reason that worked! ^^

ben-eb:
@Siilwyn Looks like a random failure from AppVeyor, I tried restarting the build and it worked but we have a merge conflict. Could you rebase again please? Then I can merge and ship rc 2 with this & the other PRs. 😄

Siilwyn:
Roger that, rebased again @ben-eb

ben-eb:
Great. Thanks! 👍

Siilwyn:
🎉

PR #390, opened 2017-07-18
Fix CI module caching for npm versions older than 2.6.1

Siilwyn:
As of npm@2.6.1, the npm update will only inspect top-level packages.
Prior versions of npm would also recursively inspect all dependencies.
The old recursive behaviour causes version mismatches.

See https://github.com/ben-eb/cssnano/pull/388#issuecomment-315902885 for more context.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,8 +17,7 @@ addons:
       - g++-4.8
 
 install:
-  - npm update
-  - npm run postinstall
+  - npm install
   - npm prune
 
 cache:


Siilwyn:
@ben-eb could you [clear the cache](https://docs.travis-ci.com/user/caching/#Clearing-Caches)?

ben-eb:
Done, & restarted the build but no joy. 😟

ben-eb:
Will try merging now that the tests are passing. 😄

ben-eb:
@Siilwyn Still a no-go I'm afraid.

ben-eb:
Latest master is passing! Thanks for resolving the issue! 👍

Siilwyn:
@ben-eb Oh sorry, I was still investigating things should have commented.

Just using `npm install` will cause the CI to use the cached modules if they are in the semver range which is different from the 'no cache' behaviour. That's why I was using `npm update`. This has the benefit of not breaking tests 'randomly' when a module does a breaking change in a minor or patch bump but this also means the CI won't 'alarm' you when this happens. Thoughts?

ben-eb:
I think in 99.9% of cases that won't be an issue, I often use the CI as a "double check" for my local tests; modules are reinstalled from scratch on every publish also, thanks to https://github.com/sindresorhus/np.

PR #396, opened 2017-08-01
Feature(postcss-svgo): optimize svg in base64.

evilebottnawi:
I will be glad to hear feedback.

Some notes:
1. Some integration tests become less.
2. Use `/i` flag for dataUri regex (https://tools.ietf.org/html/rfc2045#section-2) 
> All numeric and octet values are given in decimal notation in this
   set of documents. All media type values, subtype values, and
   parameter names as defined are case-insensitive.  However, parameter
   values are case-sensitive unless otherwise specified for the specific
   parameter.

But charset value is case case-insensitive.
3. Add more tests.


ben-eb:
Looks great, thanks!

PR #397, opened 2017-08-01
Fixed(stylehacks): prefixed value now ignored.

evilebottnawi:
Ref: https://github.com/ben-eb/cssnano/issues/395


ben-eb:
Looks great, thanks! 👍

PR #403, opened 2017-08-22
Preserve invalid input on border ordered values -  fixes #402

### LINKED ISSUE #402: Wrong minification on invalid input
Hi,

I have custom strings inside my css, 
```css
body { border: "unit(--borderWidth, px)" solid "color(color-1)"; }
```
CSS-Nano removes the strings after minification, and it becomes
```css
body{border:solid}
```

I`ve tried, and if the input is only:
```css
body { border-width: "unit(--borderWidth, px)"; }
```
The output will remain that string.
```css
body{border-width:"unit(--borderWidth, px)"}
```

Expected result, those css strings should remain at the minified result.

felixmosh:
Given invalid values to border, ordered-values should not remove them.

### DIFF
--- a/packages/postcss-ordered-values/src/__tests__/index.js
+++ b/packages/postcss-ordered-values/src/__tests__/index.js
@@ -19,6 +19,20 @@ test(
 );
 
 test(
+    'should preserve unknown values in border with order',
+    processCSS,
+    'h1{border:"unknown1" 1px "unknown2"}',
+    'h1{border:1px "unknown1" "unknown2"}'
+);
+
+test(
+    'should preserve unknown values in border with order with two properties',
+    processCSS,
+    'h1{border:solid "unknown1"}',
+    'h1{border:"unknown1" solid}'
+);
+
+test(
     'should order border with color functions',
     processCSS,
     'h1{border:rgba(255,255,255,0.5) dashed thick}',
--- a/packages/postcss-ordered-values/src/rules/border.js
+++ b/packages/postcss-ordered-values/src/rules/border.js
@@ -48,5 +48,31 @@ export default function normalizeBorder (decl, border) {
             return false;
         }
     });
-    decl.value = `${order.width} ${order.style} ${order.color}`.trim();
+
+    const parts = getBorderParts(border, decl.value);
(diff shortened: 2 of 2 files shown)


packages/postcss-ordered-values/src/__tests__/index.js
```diff
@@ -19,6 +19,20 @@ test(
 );
 
 test(
+    'should preserve unknown values in border with order',
+    processCSS,
+    'h1{border:"unknown1" 1px "unknown2"}',
+    'h1{border:1px "unknown1" "unknown2"}'
```
andyjansson: I'm not sure about this. 

@ben-eb What do you think?

ben-eb:
What's the use case for keeping them?

felixmosh:
The use case is [custom run time css processor](https://github.com/wix/wix-style-processor), that support syntax like that:
```css
body { border: "unit(--borderWidth, px)" solid "color(color-1)"; }
```

At runtime the processor replaces it with real values.
Any way, I think that plugin with the name ordered-values should not remove values that it doesn't recognize, it is weird behaviour, it suppose to reorder the values, and if it not recognize them, it should spread them with the original order.

felixmosh:
Any update?

Siilwyn:
Note I'm not a core contributor but just wanted to say I would prefer to keep the original implementation. This brings in complexity for a niche usecase. Especially since the ordering values role can be disabled.

ben-eb:
Hey, apologies for the late response but my OSS time is super-limited nowadays.

As far as I can tell there isn't anything special that is being done via this run-time style processor that couldn't be handled natively via CSS variables, no custom syntax extensions required. With that in mind I don't think it is sensible to start down a path of validating custom syntax as it adds logic that won't be useful for the large majority of our users, and then a maintenance cost to an already large codebase. I think that although the behaviour on the invalid CSS seems weird, it's better to act this way (with a minor inconvenience to turn off the transform) than to add some more logic to work around it, thereby adding more complexity.

So on that basis I will have to decline your contribution; thanks for your patience & understanding. 🙂

felixmosh:
Ok, thank you.

PR #423, opened 2017-11-12
Update gh-pages to version 1.0.0

Siilwyn:
Update the gh-pages dependency to a stable range and gets rid of
deprecation warnings during npm dependency installation.

:warning: seems like I can not test this locally because the `docs` script fails to run.


evilebottnawi:
Already done, thanks!

Siilwyn:
Good to see activity on this project again! :heart:

PR #433, opened 2018-01-31
Bugfix/add moz extension

fvoordeckers:
Fixes issue with moz-extension protocols:
https://github.com/ben-eb/cssnano/issues/414


evilebottnawi:
Thanks for PR, already done in master

Issue #430, opened 2017-12-28
removeQuotes error for minifyFontValues.


septs:
Steps to reproduce
------------------

```javascript
const postcss = require("postcss");
async function main() {
  const plugins = [require("postcss-minify-font-values")];
  const input = 'body { font-family: "STHeiti Light [STXihei]" }';
  console.log("/* Input: */\n%s", input);
  const output = await postcss(plugins).process(input);
  console.log("/* Output: */\n%s", output.css);
}
main();
```

Actual behavior
-----------------

```css
/* Input: */
body { font-family: "STHeiti Light [STXihei]" }
/* Output: */
body { font-family: STHeiti Light\ [STXihei] }
```

Output is **Invaild property value**

Expected behavior
---------------

```css
/* Input: */
body { font-family: "STHeiti Light [STXihei]" }
/* Output: */
body { font-family: STHeiti Light \[STXihei\] }
```

Reference
---------------

1. https://drafts.csswg.org/css-fonts-3/#font-family-prop
2. https://www.w3.org/TR/CSS22/syndata.html#strings
3. https://www.w3.org/TR/CSS22/syndata.html#escaped-characters
4. https://www.w3.org/TR/CSS22/syndata.html#tokenization 
5. https://github.com/ben-eb/cssnano/blob/77a8889fb500c1579740421d48d72bd2549b0f40/packages/postcss-minify-font-values/src/lib/minify-family.js

septs:
CC @ben-eb @chriseppstein

PR #435, opened 2018-02-05
Don't remove quotes from font-family by default

### LINKED ISSUE #434: Removing quotes of font-family will make it invalid
Remove the quotes will cause "invalid property value" error in chrome (of version 63.0.3239.132, may be in any version)
image

tamagokun:
This can cause "invalid property value" for certain font stacks. Closes #434 #430


evilebottnawi:
@tamagokun we should fix problem with quotes, don't disable their

tamagokun:
Totally agree, here's the commit that fixes this for good:

https://github.com/ben-eb/cssnano/commit/fbb1385fa8eb88ff8865b4be106769596c1afb65

PR #446, opened 2018-02-27
docs: dev dependency

zalog:
I think we should mark this as a dev dependency.
What do you think?


andyjansson:
None of the packages are marked with this in their `README.md`. Frankly, I think it's a non-issue. Closing this for now.

PR #456, opened 2018-03-30
fix: improve `font-family` handling

### LINKED ISSUE #430: removeQuotes error for minifyFontValues.

### LINKED ISSUE #434: Removing quotes of font-family will make it invalid

evilebottnawi:
fixes #430 fixes #434
Also fix buggy tests. Can add more tests.

### DIFF
--- a/packages/postcss-minify-font-values/src/__tests__/index.js
+++ b/packages/postcss-minify-font-values/src/__tests__/index.js
@@ -26,10 +26,79 @@ test(
 );
 
 test(
+    'should unquote font names with one character name',
+    processCSS,
+    'h1{font-family:"A"}',
+    'h1{font-family:A}'
+);
+
+test(
+    'should unquote font names with space at the start',
+    processCSS,
+    'h1{font-family:" Helvetica Neue"}',
+    'h1{font-family:\\ Helvetica Neue}'
+);
+
+test(
+    'should unquote font names with space at the end',
+    processCSS,
+    'h1{font-family:"Helvetica Neue "}',
+    'h1{font-family:Helvetica Neue\\ }'
+);
+
+test(
     'should unquote and join identifiers with a slash, if numeric',
     processCSS,
     'h1{font-family:"Bond 007"}',
     'h1{font-family:Bond\\ 007}'
+);
+
+test(
+    'should not unquote font name contains generic font family word at start',
+    passthroughCSS,
+    'h1{font-family:"serif custom font"}'
+);
+
+test(
+    'should not unquote font name contains generic font family word at middle',
+    passthroughCSS,
+    'h1{font-family:"custom serif font"}'
+);
+
+test(
+    'should not unquote font name contains generic font family word at end',
+    passthroughCSS,
+    'h1{font-family:"custom font serif"}'
+);
+
+test(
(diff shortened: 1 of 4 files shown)


packages/postcss-minify-font-values/src/__tests__/index.js
```diff
@@ -116,6 +158,55 @@ test(
     'h1{font-family:\\$42}'
 );
 
+test(
+    'should correctly escape special characters at the end',
+    processCSS,
```
andyjansson: You can use `passthroughCSS` for any test where the output is the same as the input.

andyjansson:
Maybe add `monospace` to `keywords`? I don't think the comment applies anymore. Also, I think we can also address #434 easily by making any checks against `keywords` case insensitive.

evilebottnawi:
@andyjansson let's do it :+1:

evilebottnawi:
FYI: for test https://mothereff.in/font-family

> Maybe add monospace to keywords

In code we have comment:
```
// Note that monospace is missing intentionally from this list; we should not
// remove instances of duplicated monospace keywords, it causes the font to be
// rendered smaller in Chrome.
```

But i will fix it anyway :+1:

evilebottnawi:
@andyjansson can you give me invite i close some invalid issues (with tests if needed)

andyjansson:
>In code we have comment:
>```
>// Note that monospace is missing intentionally from this list; we should not
>// remove instances of duplicated monospace keywords, it causes the font to be
>// rendered smaller in Chrome.
>```

Yeah, I know. I don't think the comment applies anymore. I modified your code to include `monospace` in the keywords collection and the `should not dedupe monospace` and `should not dedupe monospace (2)` tests pass just the same.

evilebottnawi:
@andyjansson hm, let's do it

PR #462, opened 2018-04-19
fix: add extra space after hex color and function

### LINKED ISSUE #413: Invalid css generated through colormin
First of all thank you for your excellent work, we use cssnano a lot.

We had the a failure of styling of a component imported from vue-mdl after css minification.  We tracked it down to where there is an rgb color and image url both specified.

To reproduce, using the cssnano-cli:
`echo .foo{background:rgb(1,2,3)url(bar.png) | cssnano` which gives the output as `.foo{background:#010203url(bar.png)}`.  

Leaving no space between `rgb()` and `url()` is valid syntax, unfortunately between `#...` and `url()` it is not.

evilebottnawi:
fixes #413

### DIFF
--- a/packages/postcss-colormin/src/__tests__/index.js
+++ b/packages/postcss-colormin/src/__tests__/index.js
@@ -13,10 +13,17 @@ const {
 } = processCSSWithPresetFactory('default');
 
 test(
-    'should minify color values',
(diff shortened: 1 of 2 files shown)


evilebottnawi:
@andyjansson easy read?

PR #466, opened 2018-04-19
fix: skip convert empty urls

### LINKED ISSUE #444: Maybe bug with empty url
`url("")` => `url(.)`, i think it is invalid modification

evilebottnawi:
fixes #444

### DIFF
--- a/packages/postcss-normalize-url/src/index.js
+++ b/packages/postcss-normalize-url/src/index.js
@@ -48,7 +48,7 @@ function transformDecl (decl, opts) {
             return false;
         }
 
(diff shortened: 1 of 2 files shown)


packages/postcss-normalize-url/src/index.js
```diff
@@ -52,7 +52,7 @@ function transformDecl (decl, opts) {
             return false;
         }
 
-        if (!~url.value.indexOf('chrome-extension')) {
+        if (url.value.length !== 0 && !~url.value.indexOf('chrome-extension')) {
```
andyjansson: Given #463, this should take into account other extensions as well.

PR #469, opened 2018-04-19
fix: handle uppercase property for normalize repeat

evilebottnawi:


packages/postcss-normalize-repeat-style/src/index.js
```diff
@@ -13,6 +13,7 @@ const repeatKeywords = mappings.map((mapping) => mapping[0]);
 const getMatch = getMatchFactory(mappings);
 
 function transform (decl) {
+    decl.prop = decl.prop.toLowerCase();
```
andyjansson: This kind of transformation doesn't really belong in this plugin. If properties should be normalized, a new plugin is warranted. Something along the lines of `postcss-normalize-props`.

evilebottnawi: @andyjansson It is very interesting, need test what gzip better and release new plugin :+1: Let's do it in future, not high priority.

evilebottnawi:
Close in favor https://github.com/ben-eb/cssnano/pull/500

PR #471, opened 2018-04-19
fix: handle uppercase gradients

evilebottnawi:


evilebottnawi:
Already done in master

PR #473, opened 2018-04-19
fix: handle uppercase transform property and values

evilebottnawi:


evilebottnawi:
Close in favor https://github.com/ben-eb/cssnano/pull/488

PR #474, opened 2018-04-19
fix: handle uppercase initial transform

evilebottnawi:


evilebottnawi:
Close in favor https://github.com/ben-eb/cssnano/pull/489

PR #480, opened 2018-05-29
fix(postcss-merge-rules): check for ms vendor

### LINKED ISSUE #419: mergeRules incorrectly merges placeholder selectors
For example, consider this CSS file:

```css
input:-ms-input-placeholder {
  font-size: 14px;
}
input::-ms-input-placeholder {
  font-size: 14px;
}
```

If you run cssnano against this file with the default preset, you get:

```css
input:-ms-input-placeholder,input::-ms-input-placeholder{font-size:14px}
```

Unfortunately, this won't work with Internet Explorer. These rules cannot be safely merged. When I disable the `mergeRules` optimization, this issue is resolved.

This issue may be related to https://github.com/ben-eb/cssnano/issues/297

evless:
fixes #419

### DIFF
--- a/packages/postcss-merge-rules/src/index.js
+++ b/packages/postcss-merge-rules/src/index.js
@@ -16,7 +16,7 @@ function intersect (a, b, not) {
 
 // Internet Explorer use :-ms-input-placeholder.
 // Microsoft Edge use ::-ms-input-placeholder.
-const findMsInputPlaceholder = selector => ~selector.indexOf('-ms-input-placeholder');
+const findMsInputPlaceholder = selector => ~selector.search(/-ms-input-placeholder/i);
 const different = (a, b) => intersect(a, b, true).concat(intersect(b, a, true));
 const filterPrefixes = selector => intersect(prefixes, selector);
 
--- a/packages/postcss-merge-rules/src/index.js
+++ b/packages/postcss-merge-rules/src/index.js
@@ -13,12 +13,15 @@ function intersect (a, b, not) {
     });
 }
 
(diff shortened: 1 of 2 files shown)


ai:
Maybe we should check not just prefix but whole selector? To be sure that it is placeholder selector?

evilebottnawi:
Yep, looks like better use `-ms-input-placeholder` whole selector and add comment to future developers:
```
Internet Explorer use :-ms-input-placeholder.
Microsoft Edge use ::-ms-input-placeholder.
```

packages/postcss-merge-rules/src/index.js
```diff
@@ -13,15 +13,17 @@ function intersect (a, b, not) {
     });
 }
 
-const checkMS = (a, b) => !!(~a.indexOf('ms') && ~b.indexOf('ms'));
+
+// Internet Explorer use :-ms-input-placeholder.
+// Microsoft Edge use ::-ms-input-placeholder.
+const findMsInputPlaceholder = selector => ~selector.indexOf('-ms-input-placeholder');
```
evilebottnawi: Better use `regex` with `i` flag, because css is case insensitivity language

evless: Oh, well, I did not think about it

evilebottnawi:
@evless Thanks for work!

evless:
@evilebottnawi I apologize, but can I make a commit with the addition of a list of contributors? :)

evilebottnawi:
@evless yep, feel free to send PR :+1:

PR #486, opened 2018-06-02
fix ben-eb/cssnano/448 don't remove fallbacks for custom css properties

### LINKED ISSUE #448: mergeLonghand is not always a safe transformation
The [postcss-custom-properties](https://github.com/postcss/postcss-custom-properties) plugin sets fallbacks for properties that use native CSS custom properties. Unfortunately, I noticed cssnano was stripping these fallbacks out, and I traced it back to the mergeLonghand plugin.

Here's an example of what `postcss-custom-properties` does that gets stripped by mergeLonghand:

```css
:root {
  --gutter: 1em;
}

.component {
  padding: var(--gutter);
}

/* becomes */

:root {
  --gutter: 1em;
}

.component {
  padding: 1em;  
  padding: var(--gutter);
}
```

But since cssnano sees two `padding` declarations, it tries to combine them to a single, longhand version. However, when doing that the fallback is lost.

karapetyan:
@ai could you check, please. 
Thanks.

fixes #448


packages/cssnano/src/__tests__/postcss-merge-longhand.js
```diff
@@ -154,3 +154,38 @@ test(
...
+    processCss,
+    'h1{padding-top:var(--variable);padding-right:var(--variable);padding-bottom:var(--variable);padding-left:var(--variable)}',
+    'h1{padding:var(--variable)}'
+);
+
+test(
+    'should not break unmergeable fallbacks with custom props (2)',
+    processCss,
+    'h1{padding:10em;padding:var(--variable)}',
+    'h1{padding:var(--variable);padding:10em}'
```
andyjansson: This looks wrong. Surely the output should be the same as the input..?

packages/cssnano/src/__tests__/postcss-merge-longhand.js
```diff
@@ -154,3 +154,38 @@ test(
     'h1{margin:10px 20px 30px 40px}',
     'h1{margin:10px 20px 30px 40px}',
 );
+
+test(   
+    'should merge custom props and dont remove fallbacks',
+    processCss,
+    'h1{padding-top:10px;padding-right:15px;padding-bottom:20px;padding-left:25px;padding-top:var(--variable);padding-right:var(--variable);padding-bottom:var(--variable);padding-left:var(--variable)}',
+    'h1{padding:var(--variable);padding:10px 15px 20px 25px}'
```
andyjansson: Shouldn't it be: 
```
 h1{padding:10px 15px 20px 25px;padding:var(--variable)}
```
?

packages/cssnano/src/__tests__/postcss-merge-longhand.js
karapetyan: As described in issue https://github.com/ben-eb/cssnano/issues/448
cssnano shouldn't remove fallbacks for custom CSS properties.

As far i understand that exactly what fix should do.

andyjansson: Yes, but the correct order should be: 

```
h1{padding:10em;padding:var(--variable)}
```

`var(--variable)` wouldn't be understood by browsers not supporting it, thus falling back on `10em`. In your case, you overwrite `var(--variable) ` with `10em` in all cases.

evilebottnawi: @karapetyan yep, we should have `h1{padding:10em;padding:var(--variable)}` here

packages/cssnano/src/__tests__/postcss-merge-longhand.js
karapetyan: If we look at many other tests that contain more than one unmegable rule, they have similar behavior. 

Here is some examples from same test file:
```javascript
test(
    'should not merge padding values with mixed !important',
    processCss,
    'h1{padding-top:10px!important;padding-right:20px;padding-bottom:30px!important;padding-left:40px}',
    'h1{padding-bottom:30px!important;padding-left:40px;padding-right:20px;padding-top:10px!important}',
);

test(
    'should not merge identical border values with mixed !important',
    processCss,
    'h1{border-top:1px solid #000;border-bottom:1px solid #000;border-left:1px solid #000!important;border-right:1px solid #000!important}',
    'h1{border-bottom:1px solid #000;border-left:1px solid #000!important;border-right:1px solid #000!important;border-top:1px solid #000}',
);
```
so I guess this is correct behavior

andyjansson: I disagree, the semantics have fundamentally changed. See my other comment.

packages/cssnano/src/__tests__/postcss-merge-longhand.js
karapetyan: @andyjansson @evilebottnawi understood, thanks for review :) I’ll fix it soon.

packages/postcss-merge-longhand/src/lib/getLastNode.js
```diff
@@ -1,3 +1,30 @@
...
+        ).filter(Boolean)
+    );
+}
+
+export function getRulesWithAndWithoutCustomProps (props, properties) {
+    return new Array(
+        getRulesWithoutCustomProps(props, properties),
+        getRulesWithCustomProps(props, properties)
+    );
+}
```
evilebottnawi: Looks we should rename this file to `utils.js`

packages/cssnano/src/__tests__/postcss-merge-longhand.js
karapetyan: @andyjansson done, thanks for help!

packages/cssnano/src/__tests__/postcss-merge-longhand.js
karapetyan: @evilebottnawi friendly ping

andyjansson:
Can you add a test where the properties are mixed up to demonstrate its correctness? E.g.:

```
h1{
  padding-top:10px;
  padding-right:var(--variable);
  padding-right:15px;
  padding-bottom:var(--variable);
  padding-bottom:20px;
  padding-left:25px;
  padding-top:var(--variable);
  padding-left:var(--variable)
}
```

karapetyan:
@andyjansson done

andyjansson:
@karapetyan okay, looking at the final test, I don't think the implementation is correct. `padding-right` and `padding-bottom` should be overwritten by concrete values, while the rest should not.

evilebottnawi:
@andyjansson can you explain what tests we should add more?

andyjansson:
The correct minification should be something like: 

```
h1{
  padding: 10px 15px 20px 25px;
  padding-left:var(--variable)
  padding-top:var(--variable);
}
```

karapetyan:
@andyjansson previous implementation was overcoded and awful, thank you for review. What you think about new one?

andyjansson:
I'll be off this weekend, so I won't be able to properly review this until Monday. But at first glance, it seems good.

evilebottnawi:
@andyjansson :+1: i will close other problems for `cssnano@4` and Monday after review we can merge this and maybe release `rc.3`, when two-three week and release stable 4 version

andyjansson:
Alright, the implementation looks good. I was a bit confused by the odd ordering of the properties in the tests until I realised I was looking at integration tests (the reordering is due to css-declaration-sorter). 

I'd like some unit tests where we test the plugin in isolation as well, if you don't mind (under `/packages/postcss-merge-longhand/src/__tests__`). The other thing that caught my eye was `isCustomProp` which could possibly be written so that false-positives are less likely to happen. 

Apart from that, it looks really good. Good job!

karapetyan:
@andyjansson @evilebottnawi great review, thank you guys. It was review driven development, actually)

packages/postcss-merge-longhand/src/lib/isCustomProp.js
```diff
@@ -0,0 +1,2 @@
+export default node => 
+    ~node.value.search(/var(\s+\(|\()(\s+--|--)/i);
```
andyjansson: this could be simplified a bit. E.g.:

```js
~node.value.search(/var\s*\(\s*--/i);
```
Last nitpick, I promise!

evilebottnawi:
@karapetyan Thanks!

YozhikM:
```bash
/home/stanislav/projects/cssnano/packages/postcss-merge-longhand/src/__tests__/borders.js
  316:5  error  'processCss' is not defined  no-undef
  323:5  error  'processCss' is not defined  no-undef
  330:5  error  'processCss' is not defined  no-undef
```

YozhikM:
@karapetyan @evilebottnawi 
I actually came for someone else, I think it's necessary to fix it, but not in my PR

evilebottnawi:
@YozhikM PR welcome :+1:

PR #491, opened 2018-06-04
chore(postcss-filter-plugins): Fix vulnerable downstream

kontrollanten:
https://github.com/postcss/postcss-filter-plugins/commit/86501b247fc33dd1386033111aacbf3e43fa148b

I've checked out a separate branch from the commit of `3.10.0` version. So you should be able to publish an update from this branch to `3.10.1`. Thanks!

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "postcss-discard-empty": "^2.0.1",
     "postcss-discard-overridden": "^0.1.1",
     "postcss-discard-unused": "^2.2.1",
-    "postcss-filter-plugins": "^2.0.0",
+    "postcss-filter-plugins": "^2.0.3",
     "postcss-merge-idents": "^2.1.5",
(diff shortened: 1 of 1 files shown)


kontrollanten:
Yes, that's true. But this way we can trigger cssnano's dependants to update cssnano and with a newer `post-css-plugins` version. Hopefully this will make the upgrade chain faster.

Feel free to close if you think this is unnecessary. I have a dependency that depends on cssnano, and that's why I created this PR.

evilebottnawi:
@kontrollanten just update deps, no need patch release, also in near future we will release new `cssnano@4`

kontrollanten:
I know, but I don't own the module who needs to update deps so I can't do that. The patch can work as a notice sent to all the dependants, cause I don't think everyone knows about this vulnerability.

But I'll contact the owner and ask him to do a dep update.

Issue #515, opened 2018-06-10
postcss-normalize-casing


YozhikM:
I suggest using the plugin to bring the values to the lower case.

#508

evilebottnawi:
Thanks i will review this in near

andyjansson:
I see you nabbed the package name. That's unfortunate.

andyjansson:
Closing this in favour of  #197

PR #516, opened 2018-06-10
Postcss normalize casing

YozhikM:
#515 
/cc @evilebottnawi


andyjansson:
The implementation is a bit narrow and doesn't seem very safe. I'd rather see something written using parsers, doing contextual transformation based on the node type. Scaling this implementation up to that level wouldn't be maintainable and a bit slow.

YozhikM:
@andyjansson I agree, I ran the tests with integration and collected several cases. I'll try the best myself and correct it. If cssnano is not needed in this plugin, I can close PR. What do you think?

YozhikM:
@andyjansson How do you like this?
[link](https://github.com/YozhikM/postcss-normalize-casing/blob/master/lib/index.js)

andyjansson:
At first glance, it's a great improvement. Could you refactor the code to communicate intent a bit better?

PR #522, opened 2018-06-19
fix: update initial database

evilebottnawi:
/cc @andyjansson


evilebottnawi:
@andyjansson yep, i think it can be bad, because some of their dropped, other can change behavior, also many of them have non standard initial

PR #525, opened 2018-06-26
fix: postcss-normalize-unicode causes bugs in IE11 and Edge

vitaliymaz:


vitaliymaz:
@evilebottnawi Do you mean it should works in lowercase by default and in uppercase if ie11 exists in the browser list? Correct? 
I will make that changes today.

evilebottnawi:
@vitaliymaz yes, only if ie11 exists in browserslist

PR #530, opened 2018-07-03
feat: update `browserslist` to latest version

### LINKED ISSUE #526: Move to Browserslist 4
The main difference that `browserslist()` now could return `node x.x.x`

evilebottnawi:
fixes #526

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,14 +20,14 @@
     "node": ">=6.9.0"
   },
   "devDependencies": {
-    "all-contributors-cli": "5.0.0",
+    "all-contributors-cli": "^5.0.0",
     "ava": "^0.25.0",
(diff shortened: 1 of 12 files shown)


ai:
Are you ready to find `node x.x.x` in Browserslist results? 😏

evilebottnawi:
@ai hm, what you mean? we just get browsers version in some plugins

ai:
How do you use this browser versions? I think you get some Can I Use data for this versions. You should check that code will not throw a error if Can I Use data will not find browser versions.

evilebottnawi:
@ai can you describe more in detail, I do not understand. Example how we use this https://github.com/cssnano/cssnano/blob/master/packages/postcss-minify-params/src/index.js#L92

PR #542, opened 2018-07-11
chore(deps): update `caniuse-api` to `3.0.0` version

evilebottnawi:
/cc @ai 

Should we use `ignoreUnkownVersions` in each `browserslist`?

### DIFF
--- a/packages/postcss-merge-rules/package.json
+++ b/packages/postcss-merge-rules/package.json
@@ -32,7 +32,7 @@
   "repository": "cssnano/cssnano",
   "dependencies": {
     "browserslist": "^4.0.0",
-    "caniuse-api": "^2.0.0",
+    "caniuse-api": "^3.0.0",
     "cssnano-util-same-parent": "^4.0.0",
     "postcss": "^6.0.0",
     "postcss-selector-parser": "^3.0.0",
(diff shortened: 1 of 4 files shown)


ai:
Only for tools, where we send `browsers` option (instead of queries).

evilebottnawi:
@ai here `const initialSupport = isSupported('css-initial-value', browsers);`, right?

ai:
Yeap, you need to pass `ignoreUnkownVersions` to `isSupported`

evilebottnawi:
@ai i can't find any docs about this feature in repo :confused:

ai:
They just enable it https://github.com/Nyalab/caniuse-api/blob/master/src/index.js#L47

Not the best way, but whatever :D

evilebottnawi:
@ai i.e. we can merge this and release patch version?

ai:
Yeap, there is no changes, since we didn’t pass queries to `caniuse-api`

evilebottnawi:
Need patch release `postcss-merge-rules` and `postcss-reduce-initial`

@ai @andyjansson 
Should we do patch release `cssnano`?

andyjansson:
>Should we do patch release `cssnano`?

Yes, I believe so. It's not a breaking change or a new feature. A patch is very much appropriate.

ai:
+1 for patch

PR #546, opened 2018-07-12
fix(postcss-colormin): Fix bug where color value is mutated

### LINKED ISSUE #545: `colormin` does change the actual color when transforming to hsl
I did find a bug with `postcss-colormin` that causes the underlying color value to be different after transforming it to hsl(a).

### Setup
Let's say the input for `postcss-colormin` is`rgb(221, 221, 221)`. The `hex` representation of it is `#DDDDDD`. 

### Expected Result
The correct `hsl` representation of this would be `hsl(0, 0%, 86.7%)` or `hsl(0, 0%, 86.6666667%)` to be more precise. Both `hsl` declarations would resolve in the expected hex value `#DDDDDD`.

### Actual Result
The result of colormin for this setup is `hsl(0, 0%, 87%)` which resolves into an unexpected hex color value of `#DEDEDE`.

### The Fix
`postcss-colormin` does round the result after transforming it `hsl`. Which causes a shift in the actual underlying color value. To fix this `postcss-colormin` shouldn't call `round()` anymore. Luckily calling `color.hsl().string` does already output `hsl(0, 0%, 86.7%)` instead of `hsl(0, 0%, 86.6666667%)` to give it the right amount of specificity.

I'm well aware the the visual difference between `#DDDDDD` and `#DEDEDE` is probably not noticeable by human eye but I expect `cssnano` to optimize without mutating the actual values.

weaintplastic:
Fix bug where color value is mutated when tranforming to hsl

Fixes #545


weaintplastic:
@evilebottnawi I've update the existing test case for rgba to hsla to cover that use-case plus updated the expected result of two others that change along with this fix.

evilebottnawi:
@weaintplastic thanks :+1:

packages/postcss-colormin/src/__tests__/index.js
```diff
@@ -6,7 +6,7 @@ import {
     processCSSWithPresetFactory,
 } from '../../../../util/testHelpers';
 
-const {passthroughCSS, processCSS} = processCSSFactory(plugin);
+const { passthroughCSS, processCSS } = processCSSFactory(plugin);
```
andyjansson: This is causing issues with our linting settings.

weaintplastic: @andyjansson Sorry, I've missed that and fixed it with the next commit.

weaintplastic:
@andyjansson integration tests have been updated and CI check is green. Thanks for your direction. 🙌🏻

evilebottnawi:
@andyjansson looks good for me, merge, feel free to feedback

PR #549, opened 2018-07-14
Remove v4 notice on homepage

andyjansson:
The website has the following notice on the homepage which no longer applies since v4 is released: 

>This documentation is for the upcoming version 4 release, which can be installed with npm install cssnano@next. Please test it out and report any bugs that you may find!


evilebottnawi:
@andyjansson feel free to merge and update site

Issue #554, opened 2018-07-17
mergeLonghand removes things it shouldn't remove


hudochenkov:
`mergeLonghand` removes CSS in `cssnano@4.0.2`.

Input:

```css
.b {
    border-top-color: rgba(85, 85, 85, 0.95);
    border-bottom: none;
}
```

Output:

```css
.b {
  border-bottom: none;
}
```

vseguin:
@hudochenkov Do you have any workaround? Did you downgrade meanwhile?

hudochenkov:
@vseguin I disabled this rule:

```js
require('cssnano')({
    preset: [
        'default',
        {
            mergeLonghand: false,
        },
    ],
}),
```

vseguin:
@hudochenkov  worked for me, thank you so much!

andyjansson:
Looking into this.

evilebottnawi:
Done in https://github.com/cssnano/cssnano/pull/555

evilebottnawi:
Fixed in `cssnano@4.0.3`

hudochenkov:
Thank you for a quick fix!

PR #555, opened 2018-07-18
fix(postcss-merge-longhand): Should not mangle border output

### LINKED ISSUE #553: mergeLonghand breaks border-color
`mergeLonghand` breaks CSS in `cssnano@4.0.2`.

Input:

```css
.a {
    border: solid transparent;
    border-width: 0 8px 16px;
    border-bottom-color: #eee;
}
```

Output:

```css
.a {
  border: solid transparent;
  border-color: currentColor currentColor #eee;
  border-width: 0 8px 16px;
}
```

`currentColor` is black by default, so instead of transparent, we have black.

andyjansson:
Fixes #553 and #554. 

Incidentally, the fix also revealed some issues in the integration suite, as well as provided slightly better compression.


evilebottnawi:
Today i will do patch release

PR #563, opened 2018-07-23
Refactor postcss-merge-longhand

andyjansson:
Added more precise parsing of `width` `style` `color`, which in turn has opened the door to parse borders with custom properties. Also added an additional border optimisation, increasing compression.

### DIFF
--- a/packages/postcss-merge-longhand/src/lib/decl/borders.js
+++ b/packages/postcss-merge-longhand/src/lib/decl/borders.js
@@ -8,12 +8,15 @@ import getRules from '../getRules';
 import getValue from '../getValue';
 import mergeRules from '../mergeRules';
 import minifyTrbl from '../minifyTrbl';
+import minifyWsc from '../minifyWsc';
 import canMerge from '../canMerge';
 import remove from '../remove';
 import trbl from '../trbl';
 import isCustomProp from '../isCustomProp';
 import canExplode from '../canExplode';
 import getLastNode from '../getLastNode';
+import parseWsc from '../parseWsc';
+import {isValidWsc} from '../validateWsc';
 
 const wsc = ['width', 'style', 'color'];
 const defaults = ['medium', 'none', 'currentColor'];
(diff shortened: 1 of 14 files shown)


packages/postcss-merge-longhand/src/lib/validateWsc.js
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    if (/#([0-9a-zA-Z]{6}|[0-9a-zA-Z]{3})/.test(value)) {
+        return true;
+    }
+    if (value === 'transparent') {
+        return true;
+    }
+
+    if (value === 'currentColor') {
```
evilebottnawi: All values from list can be in UPPERCASE format, css is case insensitivity (in many cases), don't forget this

packages/postcss-merge-longhand/src/lib/parseWsc.js
```diff
@@ -0,0 +1,44 @@
+import {list} from 'postcss';
+import {isWidth, isStyle, isColor} from './validateWsc';
+
+export default function parseWsc (value) {
+    if (value === 'none' || value === 'none none' || value === 'none none currentColor') {
```
evilebottnawi: Also can be case insensitivity

packages/postcss-merge-longhand/src/lib/decl/borders.js
```diff
@@ -47,49 +50,18 @@ function getLevel (prop) {
     }
 }
 
-const widths = ['thin', 'medium', 'thick'];
-const styles = ['none', 'hidden', 'dotted', 'dashed', 'solid', 'double', 'groove', 'ridge', 'inset', 'outset'];
+const isValueCustomProp = value => value.search(/var\s*\(\s*--/i);
```
evilebottnawi: var can be case insensitivity, also i think we can avoid check on `--` in regex

evilebottnawi:
@andyjansson Good work, some notes above

andyjansson:
I guess.

PR #577, opened 2018-08-06
fix: postcss-normalize-unicode add browserslist to dependencies

jbudz:
browserselist is imported [here](https://github.com/cssnano/cssnano/blob/master/packages/postcss-normalize-unicode/src/index.js#L1) but not in the package's dependency list

### DIFF
--- a/packages/postcss-normalize-unicode/package.json
+++ b/packages/postcss-normalize-unicode/package.json
@@ -28,6 +28,7 @@
   },
   "repository": "cssnano/cssnano",
   "dependencies": {
+    "browserslist": "^4.0.0",
     "postcss": "^6.0.0",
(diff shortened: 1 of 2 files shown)


andyjansson:
Can you bump it down to the same version we're using in the other packages (`4.0.0`)? We'd prefer not keeping multiple versions of the same package.

jbudz:
thanks for the review!  pushed in d86dcb4.  as a heads up I had to manually edit the yarn.lock to match the others but I think it's okay.

ianwalter:
@jbudz @evilebottnawi It seems to me that the same issue is in `postcss-minify-params` as well.

evilebottnawi:
@ianwalter please send a PR

PR #580, opened 2018-08-11
Make css-declaration-sorter safer

### LINKED ISSUE #535: cssDeclarationSorter is not safe when shorthand after longhand property
Input:

```css
.bounce {
	animation-name: bounce;
	animation: bounce2 1s ease;
	z-index: 1442;
}
```

Output:

```css
._bounce_{animation:_bounce2_ 1s ease;animation-name:_bounce_;z-index:1442}
```

Expected output:

```css
._bounce_{animation-name:_bounce_;animation:_bounce2_ 1s ease;z-index:1442}
```

/cc @evilebottnawi 

-----
Other bug (need add tests):

Input:

```css
.class {
  border-radius: 10px;
  border-bottom-left-radius: 0;
}
```

Output:

```css
.class{border-bottom-left-radius:0;border-radius:10px}
```

-----
Other bug (need also add tests):

Input:
```css
border-width: 2px;
border-style: solid;
border-top: 0;
border-left: 0;
```

Output:
```css
border-left: 0;
border-style: solid;
border-top: 0;
border-width: 2px;
```

Siilwyn:
Don't change the order of short- & longhand properties with the exception of `margin`,
`padding` & `border` since these are handled by the `mergeLonghand` optimisation.

Closes #535


zanettin:
any estimates, when this get merged? 👼

Siilwyn:
Still in progress, mostly doing local stuff. Will try to push more often. Though I'm also thinking of implementing it at the module instead of here.

zanettin:
thanks for your work on this. the "new" sort logic blocks us from upgrading to postcss v7 because the working version of `cssnano` (v3.10) relys on a fn which is no longer available on postcss v7.

Siilwyn:
@andyjansson ready for review! :)

packages/cssnano-preset-default/src/index.js
```diff
@@ -49,6 +51,9 @@ const defaultOpts = {
     normalizeCharset: {
         add: false,
     },
+    cssDeclarationSorter: {
+        customOrder: path.join(__dirname, 'safe-order.json'),
```
andyjansson: The declarations cannot be sorted based on a sort order defined in a `json` file. They need to respect the order of declaration in the CSS, else the precedence will be different and the CSS will render differently in the browser.

Siilwyn: Note that I'm leaving out the declarations that need to be kept in their original order, so this will work. Do you have an example that will fail?

packages/cssnano-preset-default/src/__tests__/css-declaration-sorter.js
```diff
@@ -0,0 +1,18 @@
...
+    passthroughCSS,
+    'h1{animation-name:a;animation:b}'
+);
+
+test(
+    'keep order of short- and longhand properties and sort remaining',
+    processCSS,
+    'h1{animation-name:a;z-index:0;animation:b;color:0;}',
+    'h1{animation-name:a;animation:b;color:0;z-index:0}'
+);
```
andyjansson: We need a lot more tests here.

Siilwyn: Could you provide a bit more descriptive feedback?

packages/cssnano-preset-default/src/__tests__/integrations/basscss5.css
```diff
@@ -18,13 +18,13 @@ th{font-weight:700;text-align:left}
 td,th{line-height:inherit;padding:.25rem 1rem}
 th{vertical-align:bottom}
 td{vertical-align:top}
-body{font-size:100%;line-height:1.5}
+body{line-height:1.5;font-size:100%}
```
andyjansson: Some strange reordering of declarations in all integration tests. Shouldn't these be grouped alphabetically?

Siilwyn: No, since `font-size` could conflict with `font` it is not being moved.

packages/cssnano-preset-default/src/__tests__/css-declaration-sorter.js
Siilwyn: Note that the module itself has tests too: https://github.com/Siilwyn/css-declaration-sorter/blob/master/tests/test.js

andyjansson: We need a range of tests to ensure that the module functions as expected, and to detect any future regressions. I expect tests covering cases where declarations should be rearranged, cases where they shouldn't, and cases with declarations of mixed precedence.

Siilwyn: The first test covers "cases where they shouldn't", the second has mixed precedence and declarations that should be rearranged right?

packages/cssnano-preset-default/src/__tests__/integrations/basscss5.css
andyjansson: What do you mean? There' s no `font` declaration here.

packages/cssnano-preset-default/src/__tests__/css-declaration-sorter.js
andyjansson: We're going to need tests in cssnano to ensure proper integration. I'm sorry if that wasn't clear.

zanettin:
thanks @Siilwyn for all your work on this ❤️ just one tiny input about sorting (not sure if you already covered this case), but in the current version of cssnano, all vendor prefixed properties are pushed to the beginning which breaks some code of us:

working code:
```
background: red;
 -webkit-background-clip: text;
```

broken after sorting:
```
 -webkit-background-clip: text;
background: red;
```

would be very nice, if you could make sure that this works as well ❤️ and again - thank you very much!!!

packages/cssnano-preset-default/src/__tests__/css-declaration-sorter.js
Siilwyn: Sorry but I still don't get what you're after. Both tests cover the cases you mentioned earlier.

packages/cssnano-preset-default/src/__tests__/integrations/basscss5.css
Siilwyn: It doesn't look at a selector by selector basis, the way this PR makes it safer is by not sorting properties that could conflict at all.

Siilwyn:
Thank you for your reply @zanettin your reply is much appreciated. :)

This is covered by this PR, sharp! I'll add a test to clarify this.

packages/cssnano-preset-default/src/__tests__/css-declaration-sorter.js
andyjansson: I'm after more tests so we can have confidence in the feature. We need tests to capture its function and any edge cases that might lurk about. Two test cases aren't nearly enough for that.

packages/cssnano-preset-advanced/src/index.js
```diff
@@ -9,6 +9,9 @@ const defaultOpts = {
     autoprefixer: {
         add: false,
     },
+    cssDeclarationSorter: {
+        customOrder: null,
```
andyjansson: Explain your reasoning behind having a different sorting order to the advanced preset and the consequences it'll have.

packages/cssnano-preset-default/src/__tests__/css-declaration-sorter.js
andyjansson: @Siilwyn when can you add more tests?

\cc @evilebottnawi

packages/cssnano/src/__tests__/fixtures.js
```diff
@@ -76,7 +76,7 @@ test(
     h2, h1 {
         font-weight: 400;
     }`,
-    `h1,h2{color:red;font-weight:400}`
+    `h1,h2{font-weight:400;color:red}`
```
evilebottnawi: Why it is happens?

packages/cssnano-preset-advanced/README.md
```diff
@@ -122,7 +122,13 @@ This plugin is loaded with the following configuration:
 
 > Sorts CSS declarations fast and automatically in a certain order.
 
-This plugin is loaded with its default configuration.
+This plugin is loaded with the following configuration:
+
+```js
+{
+	customOrder: null
```
evilebottnawi: Don't think what we need this option, no one will constantly keep their list, it's too time consuming and takes a lot of time and effort, let's remove it from here

packages/cssnano-preset-default/package.json
```diff
@@ -4,7 +4,7 @@
   "main": "dist/index.js",
   "description": "Safe defaults for cssnano which require minimal configuration.",
   "scripts": {
-    "prepublish": "cross-env BABEL_ENV=publish babel src --out-dir dist --ignore /__tests__/"
+    "prepublish": "cross-env BABEL_ENV=publish babel src --out-dir dist --ignore /__tests__/ --copy-files"
```
evilebottnawi: Why it is here?

packages/cssnano-preset-default/src/__tests__/css-declaration-sorter.js
evilebottnawi: Yep need more tests, example with border/padding/margin, also all tests from related issues

packages/cssnano-preset-advanced/README.md
Siilwyn: What do you mean? It's there so that the advanced preset does full sorting resulting in more optimization. Note that this is generated documentation too.

packages/cssnano-preset-default/package.json
Siilwyn: So that the custom order will get copied, babel normally ignores JSON files.

Siilwyn:
@evilebottnawi please read my last replies on the related issue, this PR uses simple ordering for now. As you see in some cases it can't order, this is intended.

packages/cssnano/src/__tests__/fixtures.js
Siilwyn: Because `font-weight` has a shorthand `font`.

packages/cssnano-preset-advanced/src/index.js
Siilwyn: The advanced order can make unsafe changes to save more bytes, so it orders everything.

packages/cssnano-preset-default/src/__tests__/css-declaration-sorter.js
Siilwyn: Once all other comments and questions are resolved I'll add more tests. 👍

packages/cssnano-preset-advanced/src/index.js
andyjansson: That's not really what the advanced preset is about, though. See the website on 'advanced transforms' for more info.

packages/cssnano-preset-default/src/__tests__/integrations/basscss5.css
andyjansson: I really don't know what you mean, and there's no tests to infer meaning from either. Can you expand on what you're saying?

packages/cssnano-preset-advanced/src/index.js
Siilwyn: From the advanced transform page: "Certain optimizations are not suitable for all use cases; unlike those that we bundle by default, advanced transforms all carry a certain amount of risk.". This exactly falls into this category. When using legacy code with double properties that override each other sorting everything comes with a certain amount of risk.

andyjansson: The risk with the advanced preset comes from mixing sources. When working within the same stylesheet, it should be safe. The `advanced` preset is not "bug mode".

Siilwyn: I disagree, to be more specific also quoting the same page: "This transform changes CSS semantically; it may remove styles based on certain characteristics of the CSS, or it might update values to make the CSS smaller overall. Depending on exactly what the transform does, this may be undesirable because it might change the appearance of your website in certain use cases.". Is what this optimization falls under, I think reducing the size here further is worth the tiny risk of breaking legacy code which other optimizations in this preset also do.

Thoughts @ai @evilebottnawi ?

andyjansson: >Certain optimizations are not suitable for all use cases [...] we've documented the assumptions that it must make about your CSS in order for the transform to be **safe**. [...] these transforms fall into either or both of categories.

>Assumes concatenation
> [...] assumes that the CSS passed through cssnano is all that is needed for a website to run [...]
>may be problematic if the styles you are writing in some way interact with third party styles,  or you are using multiple stylesheets instead of concatenating them 

>Changes semantics

>may remove styles based on certain characteristics of the CSS, or it might update values to make the CSS smaller overall. [...]  this *may* be undesirable [...]
>A good example is autoprefixer

What you are going is against the spirit of the preset. It has clear qualifiers on what it takes to use the advanced preset safely. It's not supposed to be a roll of the dice on whether it works or not. Please respect that.

Siilwyn: It's not a "roll of the dice" though, it is not safe when the CSS contains conflicting properties in one selector. I don't see how this is different from e.g. autoprefixer that removes outdated properties that is unsafe for CSS that contains outdated properties.

andyjansson: > it is not safe when the CSS contains conflicting properties in one selector.

Yeah, we can't have that. It's a super-common thing in CSS to have properties that affect the same space. Bringing this into the advanced preset would destroy any use case it has, rendering it useless.

Siilwyn: So being not super-common it would be useful... Is it super-common? For one, this issue took a while to surface which I think partly indicates this is uncommon. Secondly as far as I know CSS selectors normally do not contain properties that override each other, it's usually a mistake, e.g:

````
animation: bla;
animation-name: bla;
```

The `animation` property also contains the name so instead of putting `animation-name` after it normally I would just change the name in the `animation` shorthand value.

ai:
> The main idea of this plugin is to maximize the alphabetical order to better gzip, here i find many strange cases where we broke alphabetical order without reasons. Maybe we need use better logic here.

Unfortunately, it doesn’t work :(. We have an issue with this sorter right now in our current project (temporarily fixed with `!important`).

To be honest, this issue is really a high priority and I recommend to fix it ASAP.

ai:
Maybe we should disable the rule, release a fix and then continue to discuss?

CSS minifier must be safe. People should trust it. So it is very important to fix safety problems ASAP.

andyjansson:
@ai merging #599 would disable css-declaration-sorter.

@evilebottnawi @ai should we proceed with this course of action?

andyjansson:
@ai whether we should proceed with disabling css-declaration-sorter

ai:
I think we should disable this plugin at least for default preset

Siilwyn:
@ai do you still have this issue with the changes made in this PR?

Siilwyn:
CI tests are failing for an unrelated reason, running `util/buildFrameworks.js` locally for me produces different background image base64 output for some reason.

ai:
@Siilwyn yeap, accepting this PR is an option too

@andyjansson please, could we release it faster. It is a few months until the issue got reported. It is critical. We need to fix it ASAP. Disable plugin or accept this PR.

Siilwyn:
Hey it seems it got disabled looking at `master`, would like to know what needs to accept this PR. It's been open for so long it's weighing down on me. :/

andyjansson:
Read the changes we've requested. You've marked the threads as "fixed" although they haven't been.

Siilwyn:
@andyjansson do you mean the tests aren't sufficient?

packages/cssnano-preset-default/src/__tests__/integrations/basscss5.css
Siilwyn: Ah sorry I missed this reply.

I'll add an extra test, what I mean is that any property that has a conflicting property will not be sorted.

andyjansson: I don't understand what you mean by conflicting. Conflicting with *what*? The functionality, as it stands, is completely opaque and non-intuitive. The tests aren't communicating its behaviour, and that means we can't have any confidence in its result.

Siilwyn: :thinking: Ah I see, it completely makes sense to me but I know how the module works internally. Conflicting as in the value the property changes can also be changed by other properties, I've added tests to clarify the internal behavior. Could you check them to see if it clears the doubts up?

packages/cssnano/src/__tests__/fixtures.js
Siilwyn: Clarified this behaviour with a test: `move properties that are not in `safe-order.json` to the start and sort remaining` in bfe1257. Properties that are not 'safe', in other words properties that have a shorthand/longhand are moved to the start.

packages/cssnano-preset-default/src/__tests__/integrations/basscss5.css
Siilwyn: Clarified this behaviour with a test: `move properties that are not in safe-order.json to the start and sort remaining`. Properties that are not 'safe', in other words: properties that have a shorthand/longhand are moved to the start. What I meant with conflicting is for a value that can be changed by multiple properties such as the `font-size` in this example because it can also be changed with `font`. These properties 'conflict', maybe it's a bad word to describe this, I'm not a native speaker. :)

Siilwyn:
@andyjansson it is unclear for me when conversations are resolved, I have unresolved and replied to all of them. Could you go through them to resolve or give feedback? That would be great.

Siilwyn:
So @evilebottnawi if I understand it correctly this PR will not be merged because it is not optimal? Even though the current state of disabling the plugin is even worse in terms of optimizations.

Or does it depend on https://github.com/Siilwyn/css-declaration-sorter/issues/64? We can discuss it there too if you want.

Siilwyn:
:point_up: would love a reply.

Will close and delete this PR if there is no chance of being merged.

evilebottnawi:
@Siilwyn sorry, don't have time right now

PR #594, opened 2018-08-30
Moving browserslist to dependencies for postcss-minify-params

ianwalter:
Fixing the issue mentioned in #577


mrkishi:
@evilebottnawi Hi, I don't know cssnano's release process, but could this be released as a patch? It prevents installation through `pnpm` until adding `browserslist` as a direct project dependency.

PR #601, opened 2018-09-16
Meta improvements

Siilwyn:
While working on https://github.com/cssnano/cssnano/pull/580 I was struggling with the setup of this project. Made some small changes to hopefully make contributing a bit easier.

- running tests cause a lot of warnings in the console littering the test output, I first thought I did something wrong, setting `from` in the utils prevents this for most tests.
- got huge diffs from the frameworks, quickly found the `buildFrameworks` script but couldn't get it to run with `node`. After troubleshooting I found out `babel-node` should be used, putting it into a npm task/script clarifies this.
- I don't have `yarn` installed so wanting to install with `npm` I changed the client option in `lerna.json` however it still used `yarn`, removed this double option definition for easier switching.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
     "all-contributors": "all-contributors add",
     "lint": "eslint --ignore-path .gitignore packages/**/src util",
     "fixlint": "eslint --fix --ignore-path .gitignore packages/**/src util",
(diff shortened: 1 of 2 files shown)


evilebottnawi:
Thanks

PR #607, opened 2018-09-20
fix: handle uppercase property and values on `postcss-normalize-positons`

evilebottnawi:
Also fix README example


evilebottnawi:
Merge. Feel free to feedback. No big changes.

PR #608, opened 2018-09-20
fix: do not handle uppercase data URI in url function

evilebottnawi:


evilebottnawi:
Merge, feel free to feedback

PR #611, opened 2018-09-20
fix: handle uppercase `all` property in merge rules

evilebottnawi:


evilebottnawi:
Merge, feel free to feedback if something wrong

PR #612, opened 2018-09-20
fix: minify uppercase `weight` values in `font` property

evilebottnawi:


packages/postcss-minify-font-values/src/__tests__/index.js
```diff
@@ -119,7 +119,7 @@ test(
     'h1{font-family:"\\5FAE\\8F6F\\96C5\\9ED1"}'
 );
 
-test.only(
+test(
```
evilebottnawi: Big stupid error

evilebottnawi:
Merge, feel free to feedback

PR #617, opened 2018-09-25
Update css-declaration-sorter to version 4

radeno:
using PostCSS 7


radeno:
@evilebottnawi damn, my bad. sorry. just done it trough github. fix soon

radeno:
@evilebottnawi there are all 3 yarn.lock files outdated.
for cssnano and default and advanced profile. should i commit all of them?

evilebottnawi:
@radeno feel free to send 3 files

radeno:
@evilebottnawi it is necessary to increase minimal version of some dependency modules. for example `postcss-calc` supports postcss 7 from version 6.0.2 but in module is version ^6.0.0. let me check it and i will update PR

radeno:
@evilebottnawi `postcss-calc@6.0.2` is public now. i changed `cssnano-preset-default` to version `4.0.2` and need to be published.
then `preset-advanced` will be updated. it has default preset as dependency

packages/cssnano-preset-default/package.json
```diff
@@ -12,36 +12,36 @@
...
+    "postcss-normalize-string": "^4.0.1",
+    "postcss-normalize-timing-functions": "^4.0.1",
+    "postcss-normalize-unicode": "^4.0.1",
+    "postcss-normalize-url": "^4.0.1",
+    "postcss-normalize-whitespace": "^4.0.1",
+    "postcss-ordered-values": "^4.1.1",
+    "postcss-reduce-initial": "^4.0.2",
+    "postcss-reduce-transforms": "^4.0.1",
+    "postcss-svgo": "^4.0.1",
+    "postcss-unique-selectors": "^4.0.1"
```
evilebottnawi: No need update other plugin

radeno: version 4.0.0 has in `package.json` `postcss@6` but version 4.0.1 defines version 7.0.
it is reason why minimal version should be 4.0.1 or 4.0.2

evilebottnawi: :+1:

PR #621, opened 2018-10-02
fix(postcss-merge-longhand): Prefer concrete values over custom properties in shorthands

### LINKED ISSUE #619: border, and a border-color set to a css variable, is resulting in a unwanted border-color:currentcolor
Using both a `border` property and a `border-color` property, with the border-color property set to a css variable, is resulting in a unwanted `border-color:currentcolor` in the output.

input:
`:root{--color-var:red}div{border:1px solid;border-color:var(--color-var)}`

output:
`:root{--color-var:red}div{border-color:currentcolor;border:1px solid var(--color-var)}`

I'm able to verify this in the [cssnano playground](https://cssnano.co/playground/), which is showing:
> Latest release v4.1.4

andyjansson:
Fixes #619. 

This change tries to switch out a custom property in a border shorthand declaration for a concrete value, moving the custom property to its own longhand representation.

### DIFF
--- a/packages/postcss-merge-longhand/src/__tests__/borders.js
+++ b/packages/postcss-merge-longhand/src/__tests__/borders.js
@@ -497,3 +497,21 @@ test(
     passthroughCSS,
     'h1{border:6px solid red;border-top:6px solid var(--mycolor);}',
 );
+
+test(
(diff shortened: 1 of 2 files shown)


evilebottnawi:
@andyjansson Good work!

PR #632, opened 2018-10-11
fix: add cache for caniuse.isSupported() calls (#631)

### LINKED ISSUE #631: Speed improvement in postcss-merge-rules (124s to 16s)
We use cssnano and autoprefixer in webpack for building ~20 color schemes, each resulting in a standalone CSS file of ~300kB. The build takes really long - 124s on i5-6600 @ 3.3Ghz. CSS selectors heavily use pseudo-classes, such as :not(), etc.

Profiling shows that 112s (90%) of the time is spent on using `caniuse.isSupported()` API in **`postcss-merge-rules/dist/lib/ensureCompatibility.js`**. Replacing the `isSupported` call with cached variant dropped it to 4s so total build time dropped from 124s to 16s.

The cached variant replaced all `isSupported` calls by `isSupportedCached` in `ensureCompatibility.js`:
```js
let cache = new Map();
function isSupportedCached(feature, browsers) {
    let browsersSubCache = cache.get(feature);
    if (browsersSubCache === undefined) {
        browsersSubCache = new Map();
        cache.set(feature, browsersSubCache);
    }
    let result = browsersSubCache.get(browsers);
    if (result === undefined) {
        result = isSupported(feature, browsers);
        browsersSubCache.set(browsers, result);
    }
    return result;
}
```

The above is just an example of caching and might leak memory as it's not clearing the cache. There's `compatibilityCache` variable in the surrounding code which might be intended for this stuff but I didn't check in detail.

Right now I don't have public runnable example but I might prepare one if needed.

Profiles (can be opened in dev tools of Chrome using dev tools Settings => More Tools => Javascript Profiler => Load): [profiles.zip](https://github.com/cssnano/cssnano/files/2466795/profiles.zip)

Uncached:
image

Cached: 
image

webpack@4.20.2
postcss-loader@3.0.0
autoprefixer@9.1.5
cssnano@4.1.4 (default preset)

mareolan:
Fixes #631 

1. Split already present compatibilityCache into a cache of selectors and new cache of features. Used it in ensureCompatibility().
2. Made compatibilityCache be used in all code flows (previously it wasn't used in partialMerge - I found no reason why not to use it).

### DIFF
--- a/packages/postcss-merge-rules/src/index.js
+++ b/packages/postcss-merge-rules/src/index.js
@@ -28,13 +28,13 @@ function sameVendor (selectorsA, selectorsB) {
 
 const noVendor = selector => !filterPrefixes(selector).length;
 
-function canMerge (ruleA, ruleB, browsers, compatibilityCache) {
+function canMerge (ruleA, ruleB, browsers, compatibilityCache, caniuseCache) {
     const a = ruleA.selectors;
     const b = ruleB.selectors;
 
     const selectors = a.concat(b);
 
-    if (!ensureCompatibility(selectors, browsers, compatibilityCache)) {
+    if (!ensureCompatibility(selectors, browsers, compatibilityCache, caniuseCache)) {
         return false;
     }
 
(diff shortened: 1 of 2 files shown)


mareolan:
I changed it to standalone cache. It's still initialized at the same place as compatibilityCache because it needs to be browsers-specific.

evilebottnawi:
@mareolan i remember, i think how we can use this in all plugin to avoid duplicate code, in next release we implement this

mareolan:
Ok.

evilebottnawi:
Close in favor https://github.com/cssnano/cssnano/pull/681, anyway thanks for the PR and helping

PR #633, opened 2018-10-12
fix: do not merge properties with unset

jordrake:
I encountered this issue whilst using cssnano (great tool by the way, thank you very much!) and believe it falls into the same category as initial and inherit (where it impacts the entire property).

I wasn't sure if I need to create an issue first but if I do please let me know.

I realise that this is growing in complexity and could be done with a more generic solution (similar to the reserved keyword arrays I see in other packages within this repo) so I am happy to convert this file to that approach if you like. Similarly, I could also add 'none' to not be merged too.


evilebottnawi:
@jordrake Looks we need one-two additional tests to increase coverage

jordrake:
Just looking into the coverage right now, seems that some of those return falses are never getting hit which is curious!

jordrake:
I figured out what it is. The test case only reach canMerge if they contain all of the properties (TRBL) of the box.

PR #640, opened 2018-10-18
feat: remove extra space before media dn suppots at-rules

### LINKED ISSUE #324: media query extra space
```css
@media (min-width: 1px) {
  :root {
    background: lime;
  }
}
```

## correct minification

```css
@media(min-width:1px){:root{background:lime}}
```

- whitespace not required after `@media` unless followd by a word like `screen`, etc.

## CSSnano output

```css
@media (min-width:1px){:root{background:lime}}
```

- left optional whitespace after `@media`

alexander-akait:
fixes #324 
Also:
- More tests
- `yarn.lock` doesn't contain `cssnano-util-get-arguments`

### DIFF
--- a/packages/postcss-minify-params/src/__tests__/index.js
+++ b/packages/postcss-minify-params/src/__tests__/index.js
@@ -29,14 +29,91 @@ test(
     'should normalise @media queries (3)',
     processCSS,
     '@media (min-height: 680px),(min-height: 680px){h1{color:red}}',
-    '@media (min-height:680px){h1{color:red}}'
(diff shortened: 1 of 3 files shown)


andyjansson:
Tests look fine. Have we tested so that this change actually works as expected in different browsers? Perhaps we could throw something up on codepen and execute it with browserstack?

evilebottnawi:
@andyjansson maybe, will be great have simple tests for `browserstack`. Anyway other minifications for css remove this space and looks in their bug tracker no issue related to this.

PR #649, opened 2018-10-25
Preserve case of css custom variables. Fixes #648

### LINKED ISSUE #648: postcss-merge-longhand should preserve case of CSS custom properties
CSS custom properties are unlike other properties [case-sensitive](https://www.w3.org/TR/css-variables-1/#defining-variables)

postcss-merge-longhand however lowercases all properties which f.ex. turns

`.foo {
 border: 1px rgba(var(--camelCase),1);
}`

into

`.foo{border:1px rgba(var(--camelcase),1)}`

The culprit seems to be [parseWsc.js](https://github.com/cssnano/cssnano/blob/v4.1.7/packages/postcss-merge-longhand/src/lib/parseWsc.js#L27#L27)

Rather than bluntly lowercasing any match, it should try not to do that for custom properties.
I guess an easy workaround would be to just bailout if the string includes two consecutive dashes.

p-a:
Fix matches var(--somename) and preserve case of the variable name.

Fixes #648

### DIFF
--- a/packages/postcss-merge-longhand/src/__tests__/borders.js
+++ b/packages/postcss-merge-longhand/src/__tests__/borders.js
@@ -540,3 +540,16 @@ test(
     'h1{border:2px solid #fff;border-color:inherit}',
     'h1{border:2px solid;border-color:inherit}',
 );
+
+test(
+    'Should preserve case of css custom properties #648',
+    passthroughCSS,
+    'h1{border:1px solid rgba(var(--fooBar));}',
+);
+
+test(
+    'should overwrite some border-width props and save fallbacks and preserve case #648 2',
+    processCSS,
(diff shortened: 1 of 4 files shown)


andyjansson:
What happened to the diff?

p-a:
@evilebottnawi Thanks! Well, sure, but I'm not sure when I will be able to find time to do it! Will put it on my TODO list!

@andyjansson sorry, VS Code must have "helped out" with whitespaces.

p-a:
AFAICT there was only one place which lowercased strings, and that was taken care of by the initial commit. However, I added some more tests to ensure this.

andyjansson:
It's fine. In the long run, I'd probably prefer distinct formatting functions for widths, styles, and colors so we can handle any potential edge cases.

p-a:
When do you expect to make the merge? Will there also be a new minor release after that? (would be great)

evilebottnawi:
@p-a Today

PR #654, opened 2018-11-07
fix: load time

### LINKED ISSUE #651: Performance: Takes 4-6 seconds to load
I am using cssnano 4.1.4 with gulp (not the gulp plugin) and it's taking ~4seconds to load everytime:

```
console.time("Loading plugins"); // start measuring

const gulp = require('gulp');
const cssnano = require('cssnano');

console.timeEnd('Loading plugins');
```

And the result is usually 4-6seconds: `Loading plugins: 6321.839ms`

This is on MBP Retina with the extremely fast SSDs - it takes even longer on the standard SSDs. I read through the old issues but all I could find was that it'd be fixed in v4 which I'm already using.

evilebottnawi:
fixes #651

I find a lot of problems with perf, i will try to fix it tomorrow

### DIFF
--- a/packages/cssnano/src/index.js
+++ b/packages/cssnano/src/index.js
@@ -2,7 +2,6 @@ import path from 'path';
 import postcss from 'postcss';
 import cosmiconfig from 'cosmiconfig';
 import isResolvable from 'is-resolvable';
-import defaultPreset from 'lerna:cssnano-preset-default';
 
 const cssnano = 'cssnano';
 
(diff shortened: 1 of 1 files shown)


andyjansson:
Won't this change cause issues with the test suite by targeting the published version rather than the linked one?

evilebottnawi:
@andyjansson i will tested his manually, you it is decrease loading time, i will search how we can tests this un future

PR #655, opened 2018-11-08
fix: increase perf `postcss-svgo`

evilebottnawi:
If you `CSS` doesn't contain `url` with `svg` it is increase perf around 200-400ms

Why?
- Not all `CSS` contain `SVG` url
- import `svgo` cost `100 - 200` ms, creating `100-200` ms
- less memory usage because `svgo` was removed GC
- no need import/create svgo instance on top of plugin scope
- require is cacable (after first require) continue require cost `1ms`


evilebottnawi:
Close in favor https://github.com/cssnano/cssnano/pull/698

PR #672, opened 2019-01-04
fix path to postcss plugins

vitaliytv:


ludofischer:
I've opened a different PR to fix this as the URL has changed again in the meantime: https://github.com/cssnano/cssnano/pull/1009

PR #716, opened 2019-03-04
fix(postcss-colormin): Disable rgba(0,0,0,0) to transparent modification

### LINKED ISSUE #715: [postcss-colormin] rgba(0, 0, 0, 0) conversion to transparent not always valid
Originally posted on /ben-eb/colormin#43 (but I'm assuming that project is deprecated and issues related to colormin should be made here) and related to issue #630

> Even if browser supports the keyword transparent. It's not always valid for all rules. The color property for example does not have transparent as a value. Which mean IE10 does not allow transparency for color, but will allow rgba(0, 0, 0, 0). This seems less of and issue for other browsers.
> https://www.w3.org/TR/CSS21/colors.html#propdef-color

Another example in which transparent not being valid is Firefox appears to ignore `background-color: transparent` on inputs & textareas when using High Contrast Mode in Windows, but it respects `background-color: rgba(0,0,0,0)`

06b:
Not all browsers treat rgba(0,0,0,0) & transparent the same. Due to the fact I was unable to find a list of browsers, this change will disable the modification which should fix #715.


06b:
@evilebottnawi Good idea! I've been doing some testing on my side of things and you may have been right about this being a browser bug (or perhaps a windows high contrast mode issue with browsers in my case).

I'm going to create a bunch of different test cases, try to reach out to a few people and see what other information I can find - it might take me awhile but I'll update this issue once I find out a more.

To be safe, let's hold off on merging this for now... :)

evilebottnawi:
@06b Thanks, feel free to feedback

06b:
@evilebottnawi Just an update: So I feel that this pull request might be a bit too aggressive, especially after creating a bunch of test cases and testing them with windows high contrast mode. I'm going to try to update the original issue with my findings and create a separate pull request later this week and will close this pull request shortly after doing so.

evilebottnawi:
Feel free to feedback

PR #719, opened 2019-03-16
fix(merge-rules, merge-idents): add support for nested at-rules

### LINKED ISSUE #411: Turn off `mergeRules` for specific rules
Is it possible to switch `mergeRules` off for a specific item? At the moment I have a problem where this rule is merging `@supports` which then causes things to break, as the rule within `@supports` is actually supposed to override something but the ordering that it output is incorrect.

I know  that I can turn off `mergeRules` for everything, but I was wondering if I could do it for `@supports` rules only.

### LINKED ISSUE #720: postcss-merge-idents: nested at-rules are not supported
Filing this issue to be able to fix #411 

When using two-level at-rule nesting, sameParent only checks the first parent. This test would fail:

````
test(
    'should handle duplication within nested at-rules',
    passthroughCSS,
    [
        '@keyframes spin{from{transform:rotate(0deg)}to{transform:rotate(360deg)}}',
        '@media (max-width:400px){@keyframes spin{from{transform:rotate(0deg)}to{transform:rotate(360deg)}}}',
        '@supports (transform:rotate(0deg)){@keyframes spin{from{transform:rotate(0deg)}to{transform:rotate(360deg)}}}',
        '@media (max-width: 400px){@supports (transform:rotate(0deg)){@keyframes spin{from{transform:rotate(0deg)}to{transform:rotate(360deg)}}}}',
        '.spin{animation:1s spin infinite linear}',
    ].join('')
);
````

The `media... { supports... {` rule should passthrough like all the other rules do. It doesn't. 

The cached rules are removed from the document while processing. It happens here:

https://github.com/cssnano/cssnano/blob/d177936b2656ad286490bedc3c4ab9773a63e5bc/packages/postcss-merge-idents/src/index.js#L45

When it calls sameParent on the removed node in the next iteration, it will always return true. That's because removing it from the document gives it a `null` parent and `sameParent` will return true if any of two are null.

https://github.com/cssnano/cssnano/blob/d177936b2656ad286490bedc3c4ab9773a63e5bc/packages/postcss-merge-idents/src/index.js#L42

clshortfuse:
fixes #411 
fixes #720

### DIFF
--- a/packages/cssnano-util-same-parent/src/__tests__/index.js
+++ b/packages/cssnano-util-same-parent/src/__tests__/index.js
@@ -50,3 +50,95 @@ test('should calculate same parent (multiple at rules (uppercase))', t => {
         t.true(sameParent(h1, h2));
     });
 });
+
+test('should calculate same parent (nested at rules)', t => {
+    return postcss().process(`
+        @media screen {
+            @supports(pointer: course) {
+                h1 {}
+            }
+        }
+        @media screen {
+            @supports(pointer: course) {
+                h2 {}
+            }
+        }
+    `).then(result => {
+        const h1 = result.root.nodes[0].nodes[0].nodes[0];
+        const h2 = result.root.nodes[1].nodes[0].nodes[0];
+        t.true(sameParent(h1, h2));
+    });
+});
+
+
+test('should calculate not same parent (nested at rules)', t => {
+    return postcss().process(`
+        @media screen {
+            @supports(pointer: fine) {
+                h1 {}
+            }
+        }
+        @media screen {
+            @supports(pointer: course) {
+                h2 {}
+            }
+        }
+    `).then(result => {
+        const h1 = result.root.nodes[0].nodes[0].nodes[0];
+        const h2 = result.root.nodes[1].nodes[0].nodes[0];
+        t.false(sameParent(h1, h2));
+    });
+});
+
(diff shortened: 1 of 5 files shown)


clshortfuse:
I fixed it to pass the tests, but I'm noticing that the basic logic of depends on both having parents. If one of two doesn't have a parent, then it returns true. This is the line in the original code:

````
let hasParent = ruleA.parent && ruleB.parent;
````

That doesn't seem right to me. I would think if one of the two is `null` or `undefined` then it should return false. The tests for merge-idents and merge-rules don't like this though. I'm trying to figure out why.

clshortfuse:
There's a small flaw in `postcss-merge-idents`. Basically, the cached rules are removed from the document while processing. It happens here:

https://github.com/cssnano/cssnano/blob/d177936b2656ad286490bedc3c4ab9773a63e5bc/packages/postcss-merge-idents/src/index.js#L45

When it calls sameParent on the removed node in the next iteration, it will always return true. That's because removing it from the document gives it a `null` parent and `sameParent` will return true if any of two are null.

https://github.com/cssnano/cssnano/blob/d177936b2656ad286490bedc3c4ab9773a63e5bc/packages/postcss-merge-idents/src/index.js#L42

I have a fix for it in place. I just have to file the issue.

PR #723, opened 2019-03-20
fix(merge-rules): don't change specificity of prefixed properties

### LINKED ISSUE #717: mergeRules can change order of properties when it matters
In some cases, the order of properties is very importatnt, such as with vendor prefixes. See this input:
```css
/**
 * 1. Removes inner shadow in Safari (must not be textfield because of iOS 10)
 * 2. Removes the +- buttons in browsers (will be autoprefixed)
 */
[type="number"] {
	-webkit-appearance:none;   /* 1 */
	appearance:textfield;      /* 2 */
}

/**
 * 1. Removes the cross for emptying the field and removes border-radius in Safari (must not be textfield because of iOS 10)
 */
[type="search"] {
	-webkit-appearance:none; /* 1 */
}
```

It transforms into this (deminified):
```css
[type=number] {
	-moz-appearance:textfield;
	appearance:textfield
}
[type=number],
[type=search] {
	-webkit-appearance:none
}
```

The input first gets autoprefixed using autoprefixer. This step is fine - into the first rule, it adds `-moz-appearance:textfield` and leaves the `-webkit-appearance` untouched. That's all ok. But then cssnano [merges rules by partial declarations](https://github.com/cssnano/cssnano/tree/master/packages/postcss-merge-rules#by-partial-declarations) and that changes the order, so `-webkit` rule is last. That alone is a problem (vendor prefixes should never be last), more over in Firefox 62 it causes the number input to have a spinner buttons, since it kind of understands `-webkit` prefix from this version. Because of the need of different values of `-webkit-appearance` and `appearance`, it has to be in CSS before autoprefixer, otherwise I could just run cssnano before autoprefixer. In this case, it doesn't matter which one of autoprefixer or cssnano runs first.

clshortfuse:
fixes #717


clshortfuse:
This is a pretty significant change. It essentially a rewrite of the merging function. The logic is mostly the same save for some optimizations.

* Instead of converting the postCss.Rule nodes to string, the code now directly uses the `postCss.Declaration` class. It helps a bit with type-checking and ensure we're using the right objects. It also avoids a couple of string concatenation and manipulation functions.
* Instead of iterating through all of the declarations in the first rule in reverse, we iterate through the collected set of intersections. Each intersected declaration is then compared against the first rule's declarations and any rule with a later override/conflict (based on property name) is filtered out.
* After, instead of iterating through all declarations inside second rule, we iterate through the previously filtered intersections. For each intersected declaration, we filter, in order, those that do not appear in the same order as in the second rule's declarations, based on the declaration's base property name (without prefixes).

This allows use to safely merge new rulesets, such as:

````
.a{
  transform:scaleY(1);
  -webkit-transform:scaleX(-1)
}
.b{
  -webkit-transform:scaleX(-1);
  transform:scaleY(1)
}
````

Here, `-webkit-transform:scaleX(-1)` could safely be merged together, despite both having other `transform` rules.

Also, we can now merge this:

````
  .a{ display:block }
  .b{ display:block }
  .c{ display:block; display:inline-block}
````

Previously, the algorithm wouldn't break apart the `display` rules in `.c` because it wasn't analyzing the rule order. Now, the result would be: `.a,.b,.c{display:block} .c{display:inline-block}`;

PR #724, opened 2019-03-20
[postcss-colormin] Add Firefox Quantum to hasTransparentBug

### LINKED ISSUE #715: [postcss-colormin] rgba(0, 0, 0, 0) conversion to transparent not always valid

06b:
Fixes #715


packages/postcss-colormin/src/index.js
```diff
@@ -18,10 +18,27 @@ function walk (parent, callback) {
...
+    const firefoxQuantum = 57;
+    //const firefoxTransparentBugFixedVersion; //TODO: Waiting on this bug to be fixed https://bugzilla.mozilla.org/show_bug.cgi?id=1536616
+
+    if (browser.includes('firefox')){
+        if (browserVersionNumber >= firefoxQuantum) { // && browserVersionNumber < firefoxTransparentBugFixedVersion) {
+            browsersWithTransparentBug.push(browser);
+        }
+    }
+
+    return ~browsersWithTransparentBug.indexOf(browser);
```
evilebottnawi: We can simplify this using `return ~["ie 8", "ie 9", "firefox 57"].indexOf(browser);`

06b: Except the problem is this will happen with every version of firefox 57 & above until firefox fixes the issue.

evilebottnawi: I think we should solve this problem in other way, we should check if we in `-ms-high-contrast` and firefox in your browsers list, now we break optimization for many browsers who can't have problems

06b: I agree with you, I don't like the idea of breaking the optimization for the other browsers. 

Firefox >= 57 has a global coverage of 3.71% according to [browserlist](https://browserl.ist/?q=Firefox+%3E%3D+57) and I don't know the percent of users that use High Contrast Mode (that are also using Firefox >= 57), but this only is a problem if a developer is using `background-color: transparent !important;` on an input (as this doesn't appear to happen with `background-color: transparent`). _I've already changed my css so I'm not running into this issue anymore._

Considering a bug has already been sent to Firefox - **I think it's okay with leaving cssnano as it is**. 

If you agree @evilebottnawi, I can close this PR and I'll leave it up to you if you think #715 should be closed.

evilebottnawi: Yep, bug it very limited and should be fixed on firefox side, don't think good idea break optimization, let's keep open, i think about this in near future, maybe we should output warning and provide workaround for this to avoid bad DX

Issue #726, opened 2019-03-26
Used fonts are improperly discarded


GreenGremlin:
Given the following css:

```css
@font-face {
    font-family: 'FontAwesome';
    src: url('/fonts/fontawesome-webfont.woff2?v=4.7.0') format('woff2'),
         url('/fonts/fontawesome-webfont.woff?v=4.7.0') format('woff');
}

.icon {
    font: normal normal normal 14px/1 FontAwesome;
}
```

...the FontAwesome @font-face definition is improperly removed. This appears to be [the expected behavior](https://github.com/cssnano/cssnano/blob/master/packages/postcss-discard-unused/src/__tests__/index.js#L88), however this is valid css and should properly be detected as a use of the font. Would you be open to a PR that adds support for proper font use detection with mixed quote styles?

evilebottnawi:
@GreenGremlin i don't understand you, removing `@font-face` is unsafe operation so we don't do this. Please prove more information. What you actually have and what you expected.

GreenGremlin:
I apologize, I should have mentioned that I'm using the postcss-discard-unused plugin specifically to remove unused css from a concatenated css file. I believe this is also possible through cssnano using the `discardUnused` option. The problem is that, in this case, it is removing a font that is actually used.

evilebottnawi:
@GreenGremlin can you provide example or minimum reproducible test repo?

GreenGremlin:
I just created https://github.com/cssnano/cssnano/pull/727 to fix the issue. The code block above is a decent, minimal example. Simply processing that block of css through postcss with the postcss-discard-unused plugin enabled with the default settings will produce the bug.

```js
postcss([postcssDiscardUnused]).process(`
    @font-face {
        font-family: 'FontAwesome';
        src: url('/fonts/fontawesome-webfont.woff2?v=4.7.0') format('woff2'),
             url('/fonts/fontawesome-webfont.woff?v=4.7.0') format('woff');
    }

    .fa {
        font: normal normal normal 14px/1 FontAwesome;
    }
`);
```

Should not discard the `fontAwesome` @font-face, but it does.

smac89:
In case anyone stumbles upon this problem, may I suggest making sure that your fontawesome.css file does not have something like:

```css
i @font-face {
  font-family: 'FontAwesome';
  src: url('../../fonts/vendor/fontawesome-webfont.eot?v=4.1.0');
[...]

PR #727, opened 2019-03-27
fix: Used font-families discarded due to mixed quotation use

GreenGremlin:
Currently, postcss-discard-unused is improperly discarding @font-face definitions when their font-family's quotation style differs from its use in font and font-family styles.

**Example:**

```css
@font-face {
    font-family: 'FontAwesome';
    src: url('/fonts/fontawesome-webfont.woff2?v=4.7.0') format('woff2'),
         url('/fonts/fontawesome-webfont.woff?v=4.7.0') format('woff');
}

.icon {
    font: 14px/1 FontAwesome;
}
```

When executed on the above css, the postcss-discard-unused plugin will improperly detect the `FontAwesome` font as unused. This happens because its font-family definition in the @font-face rule is wrapped in quotes but the use of the font is not. The css specifications does not require that the font-family definition and rule match in their quotation style.

This change updates the postcss-discard-unused plugin to normalize font names before checking them against a "normalized" list of used fonts.

This required refactoring the way `fontCache.values` is built. Rather than just concatenating font and font-family values, it is now a bit smarter in that font values are first split by spaces, then commas, then flattened.

Lastly, this change simplifies the way `font-family` properties inside `@font-face` blocks are parsed. Since a valid `@font-face` should only ever contain a single `font-family` definition and that `font-family` should only have a single value, we can use `find` instead of `filter` and we also don't need to split the font-family value by commas. This makes the `filterFont` logic much more simple. This change is not necessary for the fix, and I would be ok reversing it if that would help get this MR merged.

Fixes https://github.com/cssnano/cssnano/issues/726


packages/postcss-discard-unused/src/__tests__/index.js
```diff
@@ -75,21 +75,27 @@ test(
...
+    '@font-face {font-family:"Does Exist";src:url("fonts/does-exist.ttf") format("truetype")}@font-face {font-family:\'DoesExist\';src:url("fonts/does-exist.ttf") format("truetype")}body{font-family: Does Exist; font: 10px DoesExist}'
+);
+
+test(
+    "should handle font properties that don't include a font-family",
     processCSS,
-    '@font-face {font-family:"Does Exist";src:url("fonts/does-exist.ttf") format("truetype")}body{font-family:Does Exist}',
-    'body{font-family:Does Exist}'
+    '@font-face {font-family:"Does Exist";src: url("fonts/does-exist.ttf") format("truetype")}body{font: italic bold 10px/1.5}',
+    'body{font: italic bold 10px/1.5}'
```
evilebottnawi: Don't remove existing tests, modify their output value and/or add new tests

packages/postcss-discard-unused/src/index.js
```diff
@@ -8,6 +8,11 @@ const atrule = 'atrule';
 const decl = 'decl';
 const rule = 'rule';
 
+function normalizeFontName (value) {
+    // Ignore casing and wrapping quotes when checking for font use.
+    return value.toLowerCase().replace(/^\s*(['"])(.*?)(\1)\s*$/, '$2');
+}
```
evilebottnawi: Bad idea, it is wrong, you can find how normalize font-name in other our plugin

GreenGremlin: What plugin is that? There are quite a few in this repo.

packages/postcss-discard-unused/src/__tests__/index.js
GreenGremlin: Which tests are you referring to? The “should not be responsible for normalising fonts” test was asserting the exact opposite of what this change is doing, it is replaced by the “should work with mixed quote styles” test, which has the same input with some added css and the new output. I added spaces to some tests, because none of the existing tests included spaces. I can reverse the space additions and add a new test to call out handling spaces, if that helps.

evilebottnawi: Yep, better add new tests :+1:

packages/postcss-discard-unused/src/index.js
evilebottnawi: Here right solution https://github.com/cssnano/cssnano/blob/a520e6906e7fa17951a64769f030ed7b6f44c38a/packages/postcss-minify-font-values/src/lib/minify-family.js, it is really not easy parse font-family, maybe we should move this utils to separate package

GreenGremlin: In what scenario do you see this failing? The only scenario I can think of would be if there were a font-face with a name that matched a non font-family value in a font property, something like `bold` or `10px`. That would be a very unusual edge case, and likely an invalid font name. Even then, the failure would be that the poorly named font-face would not be discarded.

evilebottnawi: It it nor poor, some font can have very weird characters in name, so we should parse this as w3 say

packages/postcss-discard-unused/src/__tests__/index.js
GreenGremlin: Ok, I've reversed the changes to the tests and committed new tests.

packages/postcss-discard-unused/src/index.js
GreenGremlin: According to [the w3's font shorthand documentation](https://www.w3.org/TR/CSS2/fonts.html#font-shorthand), the font shorthand must always include a font-size and font-family/line-height as the last two properties. Knowing this, it is actually pretty simple to extract the font-family value from a font shorthand. All we have to do is identify the last matching font-size/line-height property and strip it and everything before it out of the font value. I added a method `getFontFamilyFromShorthand` which does just that. I thin the current iteration of this change is pretty solid. Let me know what you think.

evilebottnawi: @GreenGremlin we should not use `regexp` for normalize, problem not in shorthard, problem what font name can contains unicode/"/' character and you regexp is broken in this case, we should use parser for this as i describe above

GreenGremlin: Maybe "normalize" isn't the best name for what this fix does. All this fix is trying to do is strip out wrapping quotes and escape characters from font names so that they can be compared in a base format. We really shouldn't need to fully "normalize" the fonts, like is done in the postcss-minify-font-values. I did rework my changes using postcss-value-parser, which makes the parsing a bit more clean. I also broke the font name parsing logic out to its own module, with its own tests. If you still think there are some unicode characters that will break this, please give me an example and I'll add a test for it.

GreenGremlin:
> Thanks, now it is looks good

Great!! Is there anything else I can do to help get this merged?

evilebottnawi:
@GreenGremlin thanks for PR, right now i rewrite `postcss-calc`, after this we do release

GreenGremlin:
I merged in the latest changes, resolved the conflicts, and ran prettier. It would be nice to get these changes merged before there's more conflicts.

ludofischer:
I have tested on the latest 5.0 RC and I think the issue has already been fixed.

PR #739, opened 2019-04-29
feat: better optimize gradients

alexander-akait:


ludofischer:
We can close this as https://github.com/cssnano/cssnano/pull/1941 solved it in a different way

PR #772, opened 2019-05-23
Chore migrate on latest babel and ava

evilebottnawi:


evilebottnawi:
lerna doesn't work fine with node@6 so i disable CI on node@6, but our babel setup for 6.9.0 so all work fine

evilebottnawi:
We need migrate on jest, ava is very slow on windows, let's merge this as is because test passed for linux and starts to migrate on jest

Issue #675, opened 2019-01-15
mergeLonghand is unsafe with border and CSS variables


manucorporat:
Using  4.1.8 (#671)

Input:
```css
.class {
  --border-width: 0 0 0 1px;
  --border-style: solid;
  --border-color: black;
  
  border-width: var(--border-width);
  border-style: var(--border-style);
  border-color: var(--border-color);
}
```

output:

```css
.class{
--border-width:0 0 0 1px;--border-style:solid;--border-color:#000;
border:var(--border-width) var(--border-style) var(--border-color)
}
```

Notice the generates CSS is not equivalent since border-width can take values like:

```css
border-width: 0 0 0 1px;  /* style bottom border */
```

leading a computed value of `border`:

```css
border: 0 0 0 1px solid black;
```
which is NOT valid.

**Expected behaviour:**

Do not merge border into a shortcut when using CSS variables.

Using the playground

<img width="1512" alt="screenshot 2019-01-15 at 17 58 33" src="https://user-images.githubusercontent.com/127379/51196460-74f56380-18ef-11e9-8706-55333c8c951b.png">

evilebottnawi:
@manucorporat Can't find any documentation what css variables supports multiple values

manucorporat:
yep, border seems too complicated to handle when CSS variables are used. 

We just reenabled this optimization, but seems like we have to disable it again :(
https://github.com/ionic-team/stencil/commit/cbd357926ec98d1b6a310352b4d2496c37d34adb

Happy to help, but I am not very familiarised with the css-nano codebase, where should I look into?

manucorporat:
I gave it a try, but I fall down the rabbit hole without results. I guess we'll just deactivate this optimization again

sdvg:
Similar issue with `font`:

```
--font-size: 1rem;

body {
  font: var(--font-size) sans-serif;
}
```

becomes

```
--font-size:1rem;body{font:var(--font-size),sans-serif}
```

(the comma should be a space)

Please tell if I should create a new issue, but I think this is the same problem.

evilebottnawi:
@sdvg yep, it is other issue, i think we need create new issue

kingback:
here's another case:

input

```css
div {
  height: constant(safe-area-inset-bottom);
  height: env(safe-area-inset-bottom);
  padding-bottom: constant(safe-area-inset-bottom);
  padding-bottom: env(safe-area-inset-bottom);
}
```

output

```css
div{
[...]

PR #775, opened 2019-05-30
fix(postcss-merge-longhand): treat env() as custom prop

isker:
- Treat `env()` (and the nonstandard equivalent `constant()`) as custom props.
- Don't deduplicate custom props.

This fixes the specific issue from https://github.com/cssnano/cssnano/issues/675#issuecomment-471985097 but does not fix the original issue in #675.


isker:
Hmm, `yarn test` did not run those tests on my machine.

I am not too sure what the right solution is now.  @evilebottnawi any idea?  My changes to the `decl` files caused those failures.

packages/postcss-merge-longhand/src/lib/isCustomProp.js
```diff
@@ -1 +1,3 @@
-export default (node) => ~node.value.search(/var\s*\(\s*--/i);
+export default (node) =>
+  ~node.value.search(/var\s*\(\s*--/i) ||
+  ~node.value.search(/(constant|env)\s*\(/i);
```
evilebottnawi: Let's rewrite this to `node.value.search(/(var|constant|env)\(/i)`, no need check on `--`

evilebottnawi:
Don't look on CI, we rewrite this right now

packages/postcss-merge-longhand/src/lib/isCustomProp.js
isker: Fixed.

isker:
Any update on this?

evilebottnawi:
Don't worry, will be merged and released for next release (this/next week)

ben-eb:
Closed in favour of https://github.com/cssnano/cssnano/pull/802.

PR #778, opened 2019-06-04
refactor(postcss-ordered-values): Consolidate utility functions.

ben-eb:


packages/postcss-ordered-values/src/lib/isMathFunctionNode.js
```diff
@@ -0,0 +1,13 @@
+import { both, compose, either, includes, prop, toLower } from 'ramda';
+import isFunctionNode from './isFunctionNode';
+import includedIn from './includedIn';
+
+const _isMathFunctionNode = compose(
+  either(includedIn(['min', 'max', 'clamp']), includes('calc')),
```
evilebottnawi: min/max/clamp like a calc function (new spec)

packages/postcss-ordered-values/src/lib/isVariableFunctionNode.js
```diff
@@ -0,0 +1,10 @@
+import { both } from 'ramda';
+import isFunctionNode from './isFunctionNode';
+import isNodeValueOneOf from './isNodeValueOneOf';
+
+const isVariableFunctionNode = both(
+  isFunctionNode,
+  isNodeValueOneOf(['var', 'env'])
```
evilebottnawi: `contant` (legacy `env`)

packages/postcss-ordered-values/src/lib/isMathFunctionNode.js
ben-eb: So we should handle `-webkit-clamp()` etc?

evilebottnawi: can't find docs about this, but i think will be great always drop prefix

ben-eb: OK, will update. 👍

PR #797, opened 2019-06-09
refactor(postcss-discard-duplicates): Consolidate utility functions.

ben-eb:


ben-eb:
Looks like we need a safe version of `eqProps` for testing against undefined nodes.

PR #799, opened 2019-06-10
refactor(postcss-convert-values): Consolidate utility functions.

ben-eb:


evilebottnawi:
Sorry, I am illing, I'll finish right after I get well, can write only with mobile now, sorry again, it is not professionally with my side, I am thinking i can continue work tomorrow or after tomorrow

ben-eb:
@evilebottnawi That's OK, take as long as you need. 👍

ludofischer:
Since this and the other 'consolidate utility functions' pull requests conflict with current master, and @ben-eb said they would not work on cssnano currently, I am going to close all of them to reduce clutter.

PR #809, opened 2019-08-07
discard-duplicates: option to keep first

christophehurpeau:
example:

```
.ResetButton { padding: 0 }
.SpecificButton { padding: 20px }
.ResetButton { padding: 0 }
```

without `keepFirst`:

```
.SpecificButton { padding: 20px }
.ResetButton { padding: 0 }
```

with `keepFirst: true`: 

```
.ResetButton { padding: 0 }
.SpecificButton { padding: 20px }
```

### DIFF
--- a/packages/postcss-discard-duplicates/src/index.js
+++ b/packages/postcss-discard-duplicates/src/index.js
@@ -88,13 +88,17 @@ function dedupeRule(last, nodes) {
   }
 }
 
-function dedupeNode(last, nodes) {
+function dedupeNode(last, nodes, keepFirst) {
   let index = ~nodes.indexOf(last) ? nodes.indexOf(last) - 1 : nodes.length - 1;
 
   while (index >= 0) {
     const node = nodes[index--];
     if (node && equals(node, last)) {
-      node.remove();
+      if (keepFirst) {
+        last.remove();
+      } else {
+        node.remove();
+      }
     }
   }
 }
@@ -106,8 +110,9 @@ const handlers = {
   comment: noop,
 };
 
-function dedupe(root) {
+function dedupe(root, opts) {
   const { nodes } = root;
+  const { keepFirst } = opts;
 
   if (!nodes) {
     return;
@@ -119,9 +124,9 @@ function dedupe(root) {
     if (!last || !last.parent) {
       continue;
     }
-    dedupe(last);
-    handlers[last.type](last, nodes);
+    dedupe(last, opts);
+    handlers[last.type](last, nodes, keepFirst);
   }
 }
 
-export default plugin('postcss-discard-duplicates', () => dedupe);
+export default plugin('postcss-discard-duplicates', (opts = {}) => (root) => dedupe(root, opts));


evilebottnawi:
Why we need this option?

ben-eb:
```html
<div class="SpecificButton ResetButton">
  Button
</div>
```

```css
div {
  background: red
}

.ResetButton { padding: 0 }
.SpecificButton { padding: 20px }
.ResetButton { padding: 0 }
```

In this case if the second rule is discarded the button has `20px` padding, whereas the default behaviour preserves the `0` padding.

evilebottnawi:
@ben-eb do you think we need this option?

ben-eb:
@evilebottnawi I'd argue no, unless there's a good reason that I've missed.

evilebottnawi:
@ben-eb thanks for feedback, let's wait answer

christophehurpeau:
With css modules, we have 2 files that imports the ResetButton with composes. I don't see another solution for this problem ?

evilebottnawi:
@christophehurpeau change order of classes?

christophehurpeau:
@evilebottnawi I don't understand your answer ?

anikethsaha:
I think this will create an inconsistency with the styling order.
-1 for this,

> @evilebottnawi I don't understand your answer?

@christophehurpeau you need to change the order of the class name where you are using these class in order to get the required styling

christophehurpeau:
@anikethsaha I don't understand, the order of classes is right (.ResetButton then .SpecificButton) the problem is that another css modules uses the reset css file, reimporting .ResetButton after .SpecificButton.
We cannot not import the reset file in the second file, because it needs it too.

The css is generated like this:
```
.ResetButton { padding: 0 }
.SpecificButton1 { padding: 20px }
.ResetButton { padding: 0 }
.SpecificButton2 { padding: 20px }
```

In my opinion, in that case, only the first `.ResetButton` should be kept, not the last.

anikethsaha:
```css
.ResetButton {
  padding: 0;
}
.SpecificButton {
  padding: 20px;
}
```

now if you have HTML element class using these

```html
<button class="ResetButton SpecificButton ">its 20px</button>
<button class="SpecificButton  ResetButton ">its 0px</button>
```

I am not getting your doubt. Is it like you are getting a process CSS and you cant change the order ?

christophehurpeau:
@anikethsaha in my understanding, changing the order in the dom doen't affect the applied css, only the css order matters

anikethsaha:
yeah my bad 😁

evilebottnawi:
> changing the order in the html doen't affect the applied css

I disagree, clarify what do you mean

anikethsaha:
```html
<style>
  .a {
    color: red;
  }
  .b {
    color: blue;
  }
</style>

<p class="b a">hello</p> <!-- blue -->
<p class="a b">hello</p> <!-- blue -->
```

cc @evilebottnawi

evilebottnawi:
@anikethsaha yep, you are right, but i can't understand why we need this option

anikethsaha:
> @anikethsaha yep, you are right, but I can't understand why we need this option

yeah same doubt for me too

> I think its more like a use-case dependent issue.

It should be changed with the input, not by cssnano !
order stuff is user-dependent, cause cssnano can't predict how it's being used in the HTML elements

@christophehurpeau do you have any strong reason to support this ?

christophehurpeau:
The reason is explained here: https://github.com/cssnano/cssnano/pull/809#issuecomment-553802579
I have two components that both uses the same css file with css modules `composes`
The result is something like:
```

anikethsaha:
I didnt use css module as of now of I guess thats what keeping me in doubt with your use-case.
I hope others might review this properly

Thanks 👍

evilebottnawi:
@chriseppstein please create reproducible test repo with css modules configuration

tallytarik:
How can I help keep this moving? Really need this feature at the moment!

anikethsaha:
As far as the above discussion is concerned, the use case is not clear. 

We would appreciate it if you can create an issue with repro may be using a repo or sandbox and also a use case might move the discussion forward.

Thanks

sigveio:
@tallytarik Due to no follow up in the past year and the amount of time that has passed since this PR was submitted, I'm closing it as stale. You are welcome to [open an issue](https://github.com/cssnano/cssnano/issues/new/choose) to discuss further if you believe there is a valid use case for the feature you suggested. Thanks =)

PR #810, opened 2019-08-29
Update svgo to v.1.3.0

urbanit:
Versions of js-yaml prior to 3.13.1 are vulnerable to Code Injection [https://www.npmjs.com/advisories/813]


anikethsaha:
I guess this needs the lock file as well.
also rebasing is required too 👍

evilebottnawi:
No need this update, because problem with svgo was fixed

PR #814, opened 2019-09-05
Make removing the last semicolon in postcss-normalize-whitespace configurable

silvenon:
I hope that it's ok to make this configurable, it would be really useful to me because then I could use [RTLCSS](https://rtlcss.com) after minification, which doesn't handle lack of semicolons well. I know that this isn't a great reason to add this option, but pretty please? 🙏

### DIFF
--- a/packages/postcss-normalize-whitespace/README.md
+++ b/packages/postcss-normalize-whitespace/README.md
@@ -15,18 +15,28 @@ npm install postcss-normalize-whitespace --save
 ### Input
 
 ```css
-h1{
-    width: calc(10px - ( 100px / var(--test) )) 
+h1 {
+    width: calc(10px - ( 100px / var(--test) ));
+    height: 20px;
 }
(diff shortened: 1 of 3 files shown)


silvenon:
Whatever is within the semicolon, RTLCSS treats as [value directives](https://rtlcss.com/learn/usage-guide/value-directives/), outside semicolons are [control directives](https://rtlcss.com/learn/usage-guide/control-directives/). When there is a value directive on the last declaration in the rule and the semicolon is removed, it's interpreted as a control directive. I don't see a way RTLCSS could solve this problem, because without that semicolon it can't know what the intention was.

evilebottnawi:
@silvenon it is violate CSS spec, not sure it is good idea, better open issue in RTLCSS, they should respect comment after last property `;`

silvenon:
Ok. I don't need this that badly anymore anyway.

Issue #822, opened 2019-10-01
Invalid minification of var() fallback


James0x57:
**Playground Input:**
```css
div {
  border-radius: 10px var(--foobar, );
}
```

**Actual output**
```css
div{border-radius:10px var(--foobar,)}
```
(the empty fallback in var() is invalid and results in the whole rule being ignored. The space is required)

**Expected output**
```css
div{border-radius:10px var(--foobar, )}
```
(must have the space in the fallback value to be valid)

Could be an issue somewhere upstream in postcss - happy to submit a PR either way if someone familiar with the code bases could point me roughly near the code that would be responsible.

evilebottnawi:
Yep, PR welcome

PR #835, opened 2019-11-07
fix: normalize-whitespace for var function and added tests

anikethsaha:
Fixed https://github.com/cssnano/cssnano/issues/822

And about the lockfile update, I guess some packages where outdated as I simply did `yarn install` and those changes came automatically

About the fix
---

Input

```css
div {
  border-radius: 10px var(--foobar, );
}
```

Output

```css
div{border-radius:10px var(--foobar, )}
```


anikethsaha:
> also lock files need return

didnt get this !!?

evilebottnawi:
@anikethsaha we should not delete lock files, just remove them from commit

anikethsaha:
@evilebottnawi its not deleted. I simply did `yarn install` and it automatically removed some packages, but lock files are there there.

packages/postcss-normalize-whitespace/src/__tests__/index.js
```diff
@@ -56,3 +56,11 @@ test(
...
   )
 );
+
+test(
+  'should not trim spaces inside of var function',
+  withDefaultPreset(
+    'div{border-radius:10px var(--foobar, )}',
+    'div{border-radius:10px var(--foobar, )}'
+  )
+);
```
evilebottnawi: Let's add tests:
```
div{var(--my-var, var(--my-background, pink, ))}
div{calc(var(--text-xxxl, ) * var(--text-scale-ratio-up, ))}
```
Also same tests and instead `var` use `env` (https://developer.mozilla.org/en-US/docs/Web/CSS/env)

evilebottnawi:
Maybe something wrong with lerna, update deps only in the package where you fix problem, or you can send other PR with updating all lock files across all packages

anikethsaha:
so should I un-commit the lock-files here ?

evilebottnawi:
@anikethsaha let's rebase this commit after merge PR with updating lock files

anikethsaha:
ok.

packages/postcss-normalize-whitespace/src/__tests__/index.js
evilebottnawi: Please add tests as described here

anikethsaha: sure 👍

packages/postcss-normalize-whitespace/src/index.js
```diff
@@ -19,8 +19,9 @@ function reduceWhitespaces(node) {
   } else if (node.type === 'div') {
     node.before = node.after = '';
   } else if (node.type === 'function') {
-    node.before = node.after = '';
-
+    if (node.value.toLowerCase() !== 'var') {
```
evilebottnawi: We should keep spaces in `var`and `env` functions

anikethsaha: should I add this for `env` also then ?

```js
&& node.value.toLowerCase() !== 'env'

```

evilebottnawi: Yes, better use `['var', 'env'].includes(node.value.toLowerCase())`

anikethsaha: yeah ok !

packages/postcss-normalize-whitespace/src/__tests__/index.js
anikethsaha: @evilebottnawi
removing whitespace inside of `calc` is not valid I guess.
cause of this

`div{border-radius:calc(10px var(--foobar, ))}`
⬇️
`div{border-radius:calc(10pxvar(--foobar, ))}`

And this minification is not valid. So I guess removing whitespace from `env, calc, var` is not safe.
even removing `before` and `after` of these are not safe either.

anikethsaha: @evilebottnawi This would be the fix if allowing removing of spaces inside `calc` but not inside `env and var`

```js
function reduceCalcWhitespaces(node) {
  if (node.type === "space") {
    node.value = " ";
  } else if (node.type === "function") {
    if (!["var", "env"].includes(node.value.toLowerCase())) {
      node.before = node.after = "";
    }
  }
}

function reduceWhitespaces(node) {
  if (node.type === "space") {
    node.value = "  ";
  } else if (node.type === "div") {
    node.before = node.after = "";
  } else if (node.type === "function") {
    if (!["var", "env"].includes(node.value.toLowerCase())) {
      node.before = node.after = "";
    }
    if (node.value.toLowerCase() === "calc") {
      valueParser.walk(node.nodes, reduceCalcWhitespaces);
      return false;
    }
  }
}
```

evilebottnawi:
```
div{border-radius:calc(10px var(--foobar, ))}
```

It is invalid CSS, so as i written early:

invalid CSS === invalid minification

anikethsaha: Yeah I doubted that too
Ok i will do trimming of spaces for all expected inside the var and env

anikethsaha:
image

I guess here the `Received` is the correct minification right ?

evilebottnawi:
Just add test which i written above

packages/postcss-normalize-whitespace/src/index.js
```diff
@@ -9,7 +9,9 @@ function reduceCalcWhitespaces(node) {
   if (node.type === 'space') {
     node.value = ' ';
   } else if (node.type === 'function') {
-    node.before = node.after = '';
+    if (!['var', 'env'].includes(node.value.toLowerCase())) {
```
evilebottnawi: Let's add `constant` (it is old version `env`) + simple test

anikethsaha: @evilebottnawi  I couldn't find example for `constant` to test upon. Can you share a link for that?
I can see postcss plugins for that or using other processors.

evilebottnawi: just change `env` to `constant` in safari and you can see (just add test where instead `env` you use `constant`)

anikethsaha: Ok

anikethsaha: @evilebottnawi done 👍

packages/postcss-normalize-whitespace/src/__tests__/index.js
```diff
@@ -56,3 +56,41 @@ test(
...
+
+test(
+  'should not trim spaces inside of env function',
+  processCSS(
+    'div{ border-radius:env(border-rad, ) }',
+    'div{border-radius:env(border-rad, )}'
+  )
+);
```
evilebottnawi: One more test:
```
test(
  'should not trim spaces inside of env function',
  processCSS(
    'div{ border-radius:var(border-rad, ) }',
    'div{border-radius:var(border-rad, )}'
  )
);
```

anikethsaha: @evilebottnawi done 👍

evilebottnawi:
Can you fix CI problem? We catch other problem, let's solve this in other PR

anikethsaha:
@evilebottnawi the CI is failing in this file [packages/cssnano-preset-default/src/**test**/small.test.js](https://github.com/cssnano/cssnano/blob/ed6a6fec2e850d2db6f258b47a200a47bf324535/packages/cssnano-preset-default/src/__tests__/small.test.js)

but in master, [there is no such file present](https://github.com/cssnano/cssnano/blob/master/packages/cssnano-preset-default/src/__tests__/small.test.js)

⬇️ From CI

image

🔽 in master . See no `small.test.js`

image


Where is this `small.test.js` is coming from. Is it some old cache? it was not there in previous PR also.

packages/cssnano-preset-default/src/__tests__/small.test.js
```diff
@@ -0,0 +1,12 @@
...
+const { processCSS: withDefaultPreset } = processCSSWithPresetFactory(
+  'default'
+);
+test(
+  'Check for the discarding of the duplicate',
+  withDefaultPreset(
+    '.cls{padding-top: 1rem;padding-top: calc(1rem + constant(safe-area-inset-top));padding-top: calc(1rem + env(safe-area-inset-top));}',
+    '.cls{padding-top: 1rem;padding-top: calc(1rem + constant(safe-area-inset-top));padding-top: calc(1rem + env(safe-area-inset-top));}'
+  )
+);
```
evilebottnawi: Remove this test from the PR

anikethsaha:
opps..... 
I guess that test came from one of my local branch might have forgot to commit the changes and check out to this branch.

PS :  I am rebasing to correct now. Made a little mistake while doing it. Please dont merge now. Will request your review when its ready

packages/cssnano-preset-default/src/__tests__/small.test.js
anikethsaha: done

anikethsaha:
@evilebottnawi ready for review

evilebottnawi:
Fix prettier problem https://dev.azure.com/cssnano/cssnano/_build/results?buildId=139&view=logs&jobId=3a559e2a-952e-58d2-b8db-2e604a9266d7

evilebottnawi:
Thanks!

PR #840, opened 2019-11-08
chore: updated dependencies using yarn upgrade

anikethsaha:
updated dependencies using `yarn upgrade`

some of the packages have `del-cli` in the build script and this package is not added in those package's dependencies. This won't affect the published package but may cause an error while building individual packages. Lerna bootstrap command won't affect with this.


evilebottnawi:
hm, all broken :smile: Need fix it

anikethsaha:
> hm, all broken 😄 Need fix it

Wierdly failing all linting! why ?

I don't think `yarn upgrade` will automatically update dep with breaking changes !!!!

anikethsaha:
cc @evilebottnawi , I couldn't locate the package which is having breaking changes cause of which the integration test is failing! any help !!

evilebottnawi:
Just update snapshots, maybe svgo improve algorithm compression and now we have other output

anikethsaha:
@evilebottnawi I guess its safe now, node 6 is outdated anyway.

evilebottnawi:
@anikethsaha we should keep compatibility in this PR and do breaking change in other PR, can you find what is package doesn't work with node@6?

anikethsaha:
Yup I did mean that to have a separate PR from this to resolve the node 6 issues.

evilebottnawi:
@anikethsaha let's do it, do node@6 and drop node@8

evilebottnawi:
@anikethsaha bow we can continue this PR

anikethsaha:
@evilebottnawi I rebased this one. Let's see if CI is clear or not 👍

anikethsaha:
@evilebottnawi we are good to go 😃

evilebottnawi:
@anikethsaha i think we can update some dependencies too, because we drop old node support

anikethsaha:
Yeah I guess we can, will try again. 
or else, dependent-bot could rebase to re-check if any dependencies are required or not !

Issue #829, opened 2019-10-28
The value of opacity in less was been modified by cssnano if I specificied the value in percent form


yanyu0517:
I had declared the opacity in percent form that maybe not a standard, and got the modification of value after processed by cssnano. Just like below picture

![image](https://user-images.githubusercontent.com/5326851/67652320-4c6ebf80-f97f-11e9-97a5-ea2397f79f3c.png)

98% changed to 1%

yanyu0517:
The reason for this issue has been located by my workmate. 
The 55th line of [cssnano/packages/postcss-convert-values/src/index.js
](https://github.com/cssnano/cssnano/blob/master/packages/postcss-convert-values/src/index.js#L55 ) is the point. If the opacity is declared in percent form like 98%, the content of  `pair`  object may like this: 
```
{
    number: 98,
    unit: '%',
}
```
`pair.number` is greater than 1 and the `node.value` would be assigned `1%`

anikethsaha:
seems like a bug. it should be

if a number is greater than 1
then it should be converted to a percentage of 1

like `(number/100) * 1`

opacities range from `0-1` so if the number is greater than one. convert it with respect to `1`.

or even if we can do with respect to the `unit`. if it's mentioned as % then simply pass it as it is. no need for any conversion the browser will take care of it.

evilebottnawi:
@anikethsaha feel free to send a PR with fix

evilebottnawi:
It is invalid syntax, please fix your code, invalid syntax is not allowed for minification

redeyes2015:
Just in case anyone like me bumping into this issue in the future: #881 have fixed this, but there is no release including it, yet, as of 2020/4/20. For now, if you need cssnano, replace the percentage syntax is the easiest option.

In [CSS color 3], percentage for opacity is allowed, but CSS color 4 will make this sytnax (`opacity: 98%`) valid:

* https://www.w3.org/TR/css-color-4/#transparency
* https://www.w3.org/TR/css-color-4/#typedef-alpha-value

[CSS color 3]: https://www.w3.org/TR/css-color-3/#opacity

And [Chrome] and [Firefox] seem to support it already; a short experiment also shows they do support percentage syntax of opacity.

[Chrome]: https://www.chromestatus.com/feature/5167940859068416
[Firefox]: https://bugzilla.mozilla.org/show_bug.cgi?id=1562086

parzibyte:
[...]

PR #842, opened 2019-11-13
fix: convert value for opacity consistency

anikethsaha:
fix https://github.com/cssnano/cssnano/issues/829

## Summary

- If the opacity value is greater than 1, then fallback to 1
- if the opacity value is greater than 1 and it is a percentage, then keep it as it is


evilebottnawi:
hm, it is invalid CSS syntax

evilebottnawi:
invalid syntax === invalid minification

anikethsaha:
you meant the `%` is invalid  ?

evilebottnawi:
Yes, try to use this in browser

anikethsaha:
ohh,....I guess it should be then converted into decimal 

`opacity : 40`

⬇️ 

`opacity : 0.4`

evilebottnawi:
@anikethsaha we are not linter or something like what handle incorrect syntax

anikethsaha:
> we are not linter or something like what handle incorrect syntax

😁 yea make sense

👍 closing

I guess good to close #829 too then !

interactiveRob:
Actually percentage is valid syntax and it does work in the browser. Not sure if this has changed somewhat recently. I am running into the same issue on my project using a third-party library that uses % values for opacity. 

https://developer.mozilla.org/en-US/docs/Web/CSS/opacity

Issue #812, opened 2019-09-04
background-image 


zhangchao828:
input：
background: url('https://ss0.bdstatic.com/70cFuHSh_Q1YnxGkpoWK1HF6hhy/it/u=508387608,2848974022&fm=26&gp=0.jpg?imageView2/1/w/750/h/1334');

output：
background: url('https://ss0.bdstatic.com/70cFuHSh_Q1YnxGkpoWK1HF6hhy/it/u=508387608,2848974022&fm=26&gp=0.jpg?imageView2%2F1%2Fw%2F750%2Fh%2F1334');

/->%2F

anikethsaha:
```css
background: url("https://ss0.bdstatic.com/70cFuHSh_Q1YnxGkpoWK1HF6hhy/it/u=508387608,2848974022&fm=26&gp=0.jpg?imageView2/1/w/750/h/1334");
```

⬇️

```css
background: url(
  https://ss0.bdstatic.com/70cFuHSh_Q1YnxGkpoWK1HF6hhy/it/u=508387608,
  2848974022&fm=26&gp=0.jpg?imageView2%2F1%2Fw%2F750%2Fh%2F1334=
);
```

This looks safe.
Whats the issue !?

@evilebottnawi I guess its good to close now

evilebottnawi:
/cc @anikethsaha 
It is bug, https://tools.ietf.org/html/rfc3986#section-3.4, `/` and `?` allowed in URL, we should fix it

anikethsaha:
yeah but whats wrong if we are transforming them to safe codes like `%` ?

evilebottnawi:
@anikethsaha 

> yeah but whats wrong if we are transforming them to safe codes like % ?

There may be very different servers settings so better keep them as is. Also it is less css

evilebottnawi:
Should be easy to fix, you can try to fix it

anikethsaha:
> Should be easy to fix, you can try to fix it

yeah we need to just leave it as it is right?...no need of any transform.

evilebottnawi:
> yeah we need to just leave it as it is right?...no need of any transform.

yes

zhangchao828:
> @anikethsaha
> 
> > yeah but whats wrong if we are transforming them to safe codes like % ?
> 
> There may be very different servers settings so better keep them as is. Also it is less css

yes

taoliujun:
[...]

PR #843, opened 2019-11-15
fix: removed norlization of url explicitly

anikethsaha:
fixes https://github.com/cssnano/cssnano/issues/812

### DIFF
--- a/packages/postcss-normalize-url/package.json
+++ b/packages/postcss-normalize-url/package.json
@@ -23,8 +23,6 @@
   ],
   "license": "MIT",
   "dependencies": {
-    "is-absolute-url": "^2.1.0",
-    "normalize-url": "^3.3.0",
     "postcss": "^7.0.16",
     "postcss-value-parser": "^3.3.1"
   },
--- a/packages/postcss-normalize-url/src/index.js
+++ b/packages/postcss-normalize-url/src/index.js
@@ -1,29 +1,10 @@
 import path from 'path';
 import postcss from 'postcss';
 import valueParser from 'postcss-value-parser';
-import normalize from 'normalize-url';
-import isAbsolute from 'is-absolute-url';
 
 const multiline = /\\[\r\n]/;
 // eslint-disable-next-line no-useless-escape
 const escapeChars = /([\s\(\)"'])/g;
-
-function convert(url, options) {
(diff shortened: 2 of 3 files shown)


anikethsaha:
cc @evilebottnawi

evilebottnawi:
@anikethsaha let's fix CI problems

anikethsaha:
> I will update the test cases as per the review if its goes in favor of removing the packages

@evilebottnawi  So are we going to remove that package ? will update the tests cases then !!

evilebottnawi:
@anikethsaha try to do everything in one go, tests will show us whether to remove this package or not, I can’t remember all the code, so I write tests to understand what has changed

anikethsaha:
> @anikethsaha try to do everything in one go, tests will show us whether to remove this package or not, I can’t remember all the code, so I write tests to understand what has changed

okay, will do it

anikethsaha:
@evilebottnawi should I update the `cssnano` package snapshots ?

packages/postcss-normalize-url/src/__tests__/index.js
```diff
@@ -271,7 +271,7 @@ test(
   'should handle protocol relative urls',
   processCSS(
     'h1{background:url(//website.com:80/image.png)}',
-    'h1{background:url(//website.com/image.png)}'
+    'h1{background:url(//website.com:80/image.png)}'
```
evilebottnawi: We should not remove `:80` because it is unnecessary

anikethsaha: > We should not remove `:80` because it is unnecessary

Currently it's not removing !!! you want it to be removed ?

evilebottnawi: Yes, it should be removed

anikethsaha: but will it be safe ?
like it will work fine for `80` but if its different than HTTP port like `3000`, then the port should be kept as it is right !

anikethsaha: @evilebottnawi

defcc:
Looking forward to it, Is there any good news now :)

zhangchao828:
+1

Wenjie-Shao:
is it done??

PR #851, opened 2019-12-05
fix: fixed the quotes conversion bug and added tests

### LINKED ISSUE #848: Apostrophes in selectors' attribute expressions ignored
cssnano-minified selectors wont match against the same html nodes as the original if it is looking for attributes with a `'`

E.g.

    [data-expression^="'Error"],
    [x="to'prove'a'point"],
    [x="a\000027workaround"],
    [x="should:be like!these"] {
      color : yellow;
    }

Gets transformed to `[data-expression^=Error],[x="should:be like!these"],[x=a\000027workaround],[x=toproveapoint]{color:#ff0}` in https://cssnano.co/playground/

Html like `<span data-expression="'Error: you have an error'"></span>` wont be marked yellow post minification, but should be.

I was surprised to see it wasn't just apostrophes at the ends (I thought it was overzealously truncating `"Error"`&`'Error'`), but also apostrophes in the middle.
It's just removing all of them.
I see that spaces, colons, exclamation marks, and other characters are treated specially and wrapping in quotes.
Interestingly I can work around the issue by using the codepoint.

anikethsaha:
fixes #848

I have added inline comments for details.


anikethsaha:
@evilebottnawi presets fixtures needs to be updated it seems!!

anikethsaha:
> Yes, update them

@evilebottnawi any idea how to update them. Those are not snapshots right !!!!

evilebottnawi:
https://github.com/cssnano/cssnano/blob/d4af9038034de4f776463015c312f7ad9ea8baf0/package.json#L12

anikethsaha:
@evilebottnawi I guess those changes in the fixtures seems safe.

evilebottnawi:
@anikethsaha we increase size of snapshots, it is bad, i think we should search other solution

anikethsaha:
@evilebottnawi I tried with a lot of possible regexes to match all the conditions in order to fix the issue. But as of now, I am not able to fix it.  the main issue I am facing is when the parser is building the final code as a string, for some reason it's adding `"` again to the value. 

I am closing this. If anyone else got a better fix, feel free to submit it 👍

PR #852, opened 2019-12-05
fix: min-width/height issue

### LINKED ISSUE #850: `min-width: initial` gets incorrectly modified to `min-width: 0`
This should be `min-width: auto`. This (sometimes) resolves to `min-width: 0`, but not at all times.

image

https://developer.mozilla.org/en-US/docs/Web/CSS/min-width

Caused by: https://github.com/cssnano/cssnano/tree/master/packages/postcss-reduce-initial

anikethsaha:
fixes #850 

Added fallback to `auto` for `min-width` and `min-height` !

@evilebottnawi I guess we need `auto` for `max-*` as well right ?


anikethsaha:
> Need add test to avoid future regression

@evilebottnawi done 👍

anikethsaha:
Okay sure

nex3:
Is there a release planned with this fix?

anikethsaha:
@nex3 we will do a major release soon. 

In  the meantime you can use the `nightly` tag. 

[`cssnano@nightly`](https://www.npmjs.com/package/cssnano/v/4.0.0-nightly.2020.7.24)

PR #855, opened 2019-12-12
Upgrade css declaration sorter

Siilwyn:
:wind_chime: A fresh take on PR #580 to resolve issue #535 making `css-declaration-sorter` safe by adding a new option `keepOverrides`.

For now I'm using the GitHub repository to get some feedback on this PR, once it's reviewed I'll publish the changes as a new major version 5.0.

I do need some help to get all tests working, I did update the snapshots but can't get the integration tests to pass.


packages/cssnano-preset-default/package.json
```diff
@@ -14,7 +14,7 @@
   ],
   "license": "MIT",
   "dependencies": {
-    "css-declaration-sorter": "^4.0.1",
+    "css-declaration-sorter": "git://github.com/Siilwyn/css-declaration-sorter#phoenix",
```
evilebottnawi: No need use git, some developers can't have git and it is failed

Siilwyn: Yes this is temporary so I can easily make changes to the module for now.

So I will publish version 5 before merging this PR, then I will change this!

Siilwyn: Hey @evilebottnawi I've published the new version!

Siilwyn:
@evilebottnawi tests are here in the package: https://github.com/Siilwyn/css-declaration-sorter/blob/master/tests/test.js
Do you want to add tests to cssnano too?

I've reset the yarn lock files.

Siilwyn:
@evilebottnawi since you did not reply I assumed you want tests in this project too, I have added test with commit d16ccfb.
Please review, thanks!

Siilwyn:
@ben-eb could you review this PR?

Only the framework integration tests are failing even after running `yarn build:integration`, any help appreciated.

ben-eb:
I'm pretty sure the integration tests work on diffing the whole CSS file, which unfortunately makes it difficult to know what changed. If you have changed the order then it's likely to have affected a lot of the integration tests; but there should be a command you can use to rebuild them and then verify the output in some way.

Ideas on changing the integration tests to make the output more human-readable of course welcomed. 🙂

Siilwyn:
@ben-eb thanks for your reply, what command would rebuild them? As I mentioned `yarn build:integration` gives the same failing fixture output.

ben-eb:
Ah, sorry I misread that part of your message. Seems strange that running those didn't fix it. 😞

Siilwyn:
@evilebottnawi @ben-eb could you try running the integration generation (`yarn build:integration`) on your machine? And if it results in changes push it to this branch or create a PR on this branch?
Maybe it just doesn't work on my Linux setup?

Siilwyn:
@evilebottnawi ready for review.

packages/cssnano-preset-advanced/README.md
```diff
@@ -106,15 +106,15 @@ Note that you may wish to publish your own preset to npm for reusability, should
 
 ## Plugins
 
-### [`autoprefixer`](https://github.com/postcss/autoprefixer) (external)
+### [`autoprefixer`](shortcut://github.com/postcss/autoprefixer) (external)
```
evilebottnawi: Please use `https`

packages/cssnano-preset-advanced/README.md
```diff
@@ -342,5 +342,5 @@ See [CONTRIBUTORS.md](https://github.com/cssnano/cssnano/blob/master/CONTRIBUTOR
 
 ## License
 
-MIT © [Ben Briggs](http://beneb.info)
+[MIT](https://opensource.org/licenses/MIT) © [Ben Briggs](http://beneb.info)
```
evilebottnawi: Do not change it

Siilwyn:
Alright I tried to reduce the diff amount @evilebottnawi :+1:

Siilwyn:
> **Can you describe briefly what has been fixed?**

Prevent breaking legacy CSS where shorthand declarations override longhand declarations. For example `animation-name: some; animation: greeting;` will be kept in this order.

> **What algorithm do you use ...?**

Using this function:
https://github.com/Siilwyn/css-declaration-sorter/blob/f8245d916d900f5c83d475b332ada4109a5fe4b8/src/index.js#L125-L134

Which checks if a declaration is shorthand property and if the declaration it's compared against is a longhand version of that property [using this list](https://github.com/Siilwyn/css-declaration-sorter/blob/master/src/shorthand-data.js). If the latter is true it will keep these two declarations in the same order.

> **... and show test examples**
- These are [the relevant tests](https://github.com/Siilwyn/css-declaration-sorter/blob/f8245d916d900f5c83d475b332ada4109a5fe4b8/tests/test.js#L147-L172) at the css-declaration-sorter repo.
- More tests are included in this PR at `packages/cssnano-preset-default/src/__tests__/css-declaration-sorter.js`

@evilebottnawi alright, hope this clarifies everything!

evilebottnawi:
I looked at the code and it seems to look good, but there is one point - unknown properties. It is theoretically possible that browsers will have more properties that can be written in short or long. For example:

```css
a {
  unknown-long-property: value;
  unknown-short-property: value;
}
```

I think we should keep them as is because we can potential break code.

Siilwyn:
@evilebottnawi I think you are right, didn't think of this niche case but it is certainly possible when older versions of cssnano & my module are used while CSS introduced new properties not accounted for.

I have published css-declaration-sorter 5.1.0 which looks up properties in [this JSON file](https://github.com/Siilwyn/css-declaration-sorter/blob/master/orders/alphabetical.json), if it doesn't find the property it will retain it's order compared with other unknown properties.
Also added extra tests for this.

evilebottnawi:
Good job, can you fix conflicts?

Siilwyn:
@evilebottnawi fixed the conflicts, I did fine the integration tests to timeout. Seems that the lookup for every property slows down the module a bit and all integration tests together push it to take more than the default Jest timeout of 5 seconds.

evilebottnawi:
Let's remove `passthrough` from file to fix lint problem and i think we can merge. I hope everything will be fine, because previously time we broke the code :smile:

evilebottnawi:
I think I found another problems:

1. Invalid order with `animation`:
```js
const postcss = require('postcss');
const cssDeclarationSorter = require('css-declaration-sorter');

const code = `
a {
    animation-name: Three; 
    animation: Two 2s;
}
`;

postcss([cssDeclarationSorter({ order: 'alphabetical' })])
  .process(code)
  .then((result) => console.log(result.css));

```

Other order animation order will be `Three`, but in my code it is `Two`.

evilebottnawi:
Other problem:
```
const postcss = require('postcss');
const cssDeclarationSorter = require('css-declaration-sorter');

const code = `
a {
    animation-name: Three; 
    animation: Two 2s;
    -webkit-animation: One 2s;
}
`;

postcss([cssDeclarationSorter({ order: 'alphabetical' })])
  .process(code)
  .then((result) => console.log(result.css));
```

Very rare usage, but in chrome I will have `One` animation, but after plugin it was `Three`

evilebottnawi:
Other problem, prefixes should keep place:
```js
const postcss = require('postcss');
const cssDeclarationSorter = require('css-declaration-sorter');

const code = `
a {
    animation-name: Three; 
    animation: Two 2s;
    -webkit-animation: One 2s;
    -webkit-name: Four;
}
`;

postcss([cssDeclarationSorter({ order: 'alphabetical' })])
  .process(code)
  .then((result) => console.log(result.css));
``

**Comment by Siilwyn** [comment]
@evilebottnawi ah yes didn't think of that. I have published a new version that uses postcss to strip the vendor prefix before comparing properties. Also updated the tests.

**Comment by Siilwyn** [comment]
@evilebottnawi the sorter only runs within at-rules so PostCSS will handle the AST, what do you want to test?

Things like, for example the following?
```js
test(
  'keep order of short- and longhand properties, within at-rule',
  passthroughCSS('@supports (display: grid) {div {animation: a; animation-name: b}}')
);
```

evilebottnawi:
@Siilwyn yes, also we should sort properties inside `at-rules`

evilebottnawi:
@Siilwyn please avoid using `require('fs').promises` for `node@10`, because it is create the warning and annoy developers, they will create a issue about it

evilebottnawi:
Also
```js
const postcss = require('postcss');
const cssDeclarationSorter = require('css-declaration-sorter');

const code = /* CSS */ `
a {
  animation-name: Three;
  animation: Two 2s;
  -webkit-animation: One 2s;
  -webkit-name: Four;
}
`;

postcss([cssDeclarationSorter({ order: 'alphabetical', keepOverrides: true })])
  .process(code)
  .then((result) => console.log(result.css));
```

Still broken, output:

```css
a {
  -webkit-name: Four;
  animation-name: Three;
  animation: Two 2s;
  -webkit-animation: One 2s;
}
```

`-webkit-name` should be after `-webkit-animation`

Siilwyn:
What is `-webkit-name`? I can't find this property anywhere on google or MDN docs.

Siilwyn:
The FS promises API is stable and won't produce a warning with `node@10`, see the docs: https://nodejs.org/dist/latest-v10.x/docs/api/fs.html#fs_fs_promises_api

evilebottnawi:
@Siilwyn oh, my mistake, sorry, all works fine, i mean `-webkit-animation-name` :smile:

evilebottnawi:
@Siilwyn it is stable for LTS release, but for non LTS releases it outputs the warning, for example use `v10.15.2` and you will see the warning

evilebottnawi:
Also please fix conflicts

evilebottnawi:
Also we have interesting case:
```js
const postcss = require('postcss');
const cssDeclarationSorter = require('css-declaration-sorter');

const code = /* CSS */ `
a {
  animation-name: Three;
  animation: Two 2s;
  animation-future-new-long-property: 20s;
}
`;

postcss([cssDeclarationSorter({ order: 'alphabetical', keepOverrides: true })])
  .process(code)
  .then((result) => console.log(result.css));

```

Output:

```css
a {
  animation-future-new-short-property: 20px;
  animation-name: Three;
  animation: Two 2s;
}
```

But will be great to keep that property in current place:

```css
a {
  animation-name: Three;
  animation: Two 2s;
  animation-future-new-short-property: 20px;
}


In theory, in future any short property can have a new long property, so we should keep them in right place, what do you think?

**Comment by Siilwyn** [comment]
@evilebottnawi I can keep unkown properties in place but it does 'block' the surrounding properties from being sorted.
E.g:
`a{z-index: 0; unkown-a: 0; border: 0;}` will stay like this.

Would you be okay with that? I think it's fine since most properties are known and this is a niche case.

**Comment by evilebottnawi** [comment]
@Siilwyn Yes, new unknown properties is very rare edge case, let's finish this PR:
- avoid using `fs.promise` to avoid `warning` on non LTS
- added a new test - `:root {--a-var: 'value'; --c-var: calc(10px + 20px); --b-var: 12px; }

**Comment by Siilwyn** [comment]
@evilebottnawi 
:heavy_check_mark: 
:heavy_check_mark: 
:heavy_check_mark: 

All done! :)

**Comment by evilebottnawi** [comment]
@Siilwyn Good job! 

To be honesty we can change order of variables too `:root {--a-var: 'value'; --c-var: calc(10px + 20px); --b-var: 12px; }`. 

For example:
```css
:root {
 --b: var(--a);
 --a: 10px;
}
```

It is work.

But it is feature. Let's open issue in `css-declaration-sorter` for that?

evilebottnawi:
Already exists https://github.com/Siilwyn/css-declaration-sorter/issues/54 :smile:

Siilwyn:
Haha yes, but should not be needed for this PR right?

evilebottnawi:
Yes, we can solve this late, not high priority

Siilwyn:
Yes! :tada:

PR #856, opened 2019-12-14
fix(postcss-colormin): fixed plugin running error

### LINKED ISSUE #818: postcss-colormin: code in "readme" contains an error
I added an example archive.

```
TypeError: colormin is not a function
```

And postcss-colormin cannot be used as simple postcss plugin, because is
```Error: [object Object] is not a PostCSS plugin```
I think these problems are related, so maybe it’s good to add the following archive, but i think it's excess.

[postcss-colormin.zip](https://github.com/cssnano/cssnano/files/3608734/postcss-colormin.zip)

### PR DESCRIPTION
**Pull request by anikethsaha**
fix #818 

I moved the generate script to an outside folder and changed its module to common js module and using the node runner to build the `keyword.json` instead of `babel-node`

### DIFF
--- a/packages/postcss-colormin/package.json
+++ b/packages/postcss-colormin/package.json
@@ -9,7 +9,7 @@
   ],
   "scripts": {
     "prebuild": "del-cli dist",
-    "build": "cross-env BABEL_ENV=publish babel-node --config-file ../../babel.config.js ./src/generate.js && babel src --config-file ../../babel.config.js --out-dir dist --ignore \"**/__tests__/,src/generate.js\"",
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by anikethsaha** [comment]
cc  @evilebottnawi

**Comment by anikethsaha** [comment]
With keeping that , there were some issues with running it with postcss.

I know the other plugins are using the same but I couldn't find why this was breaking here so I made this fix. Behavior is same 👍  but its fixing the issue

**Comment by anikethsaha** [comment]
You want the script to be run using babel-node ?

```shell
&& babel-node script/generate
```
?

evilebottnawi:
Yes

anikethsaha:
> Yes

ok cool. will change that 👍

PR #860, opened 2019-12-24
chore: added plugin options validation for postcss-covert-value

anikethsaha:
Added `schema-utils` as dep for `postcss-convert-values` to validate the options.

This PR is more like a PoC, so if its good to have, then I will go on with adding it to the rest of the packages.

> Added a custom jest matcher in order to match the error message in a formatted way.

### DIFF
--- a/packages/postcss-convert-values/src/options.json
+++ b/packages/postcss-convert-values/src/options.json
@@ -1,0 +1,22 @@
+{
+  "type": "object",
+  "properties": {
+    "length": {
+      "description": "Pass false to disable conversion from px to other absolute length units, such as pc & pt & vice versa.",
+      "type": "boolean"
+    },
+    "time": {
+      "description": "Pass false to disable conversion from ms to s & vice versa.",
+      "type": "boolean"
+    },
+    "angle": {
+      "description": "Pass false to disable conversion from deg to turn & vice versa.",
+      "type": "boolean"
+    },
+    "precision": {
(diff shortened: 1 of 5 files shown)


anikethsaha:
cc @evilebottnawi

packages/postcss-convert-values/src/options.json
```diff
@@ -0,0 +1,22 @@
...
+      "description": "Pass false to disable conversion from deg to turn & vice versa.",
+      "type": "boolean"
+    },
+    "precision": {
+      "description": "Specify any numeric value here to round px values to that many decimal places; for example, using {precision: 2} will round 6.66667px to 6.67px, and {precision: 0} will round it to 7px. Passing false (the default) will leave these values as is.",
+      "anyOf": [{ "type": "boolean" }, { "type": "number" }]
+    }
+  },
+  "additionalProperties": false
+}
```
evilebottnawi: Let's ignore this here, we need discussion about validate properties and do it in other PR

anikethsaha: I didnt get ? 
you want to remove the `false` ?

evilebottnawi: I want to avoid validation right now, it have some decrease performance and need discussion, do not start to implement some things before we had no discussion

anikethsaha: okay so should I close this PR or keep it open as a PoC?

evilebottnawi: Keep it open as a PoC

Issue #756, opened 2019-05-06
More properties for `postcss-ordered-values`


evilebottnawi:
Now we have only:
- animation: animation,
- border: border,
- border-top
- border-right
- border-bottom
- border-left
- outline: border,
- box-shadow
- flex-flow
- transition

ben-eb:
Have a look into these properties:

* https://developer.mozilla.org/en-US/docs/Web/CSS/border-block
* https://developer.mozilla.org/en-US/docs/Web/CSS/border-block-end
* https://developer.mozilla.org/en-US/docs/Web/CSS/border-block-start
* https://developer.mozilla.org/en-US/docs/Web/CSS/border-inline
* https://developer.mozilla.org/en-US/docs/Web/CSS/border-inline-end
* https://developer.mozilla.org/en-US/docs/Web/CSS/border-inline-start
* https://developer.mozilla.org/en-US/docs/Web/CSS/clip-path
* https://developer.mozilla.org/en-US/docs/Web/CSS/column-rule
* https://developer.mozilla.org/en-US/docs/Web/CSS/columns
* https://developer.mozilla.org/en-US/docs/Web/CSS/display
* https://developer.mozilla.org/en-US/docs/Web/CSS/font-synthesis
* https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant
* https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-alternates
* https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-east-asian
* https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-ligatures
* https://developer.mozilla.org/en-US/docs/Web/CSS/grid-auto-flow
* https://developer.mozilla.org/en-US/docs/Web/CSS/hanging-punctuation
* https://developer.mozilla.org/en-US/docs/Web/CSS/list-style
* https://developer.mozilla.org/en-US/docs/Web/CSS/shape-outside
* https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration
* https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration-line
* https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration-skip
* https://developer.mozilla.org/en-US/docs/Web/CSS/text-emphasis
* https://developer.mozilla.org/en-US/docs/Web/CSS/text-emphasis-style
* https://developer.mozilla.org/en-US/docs/Web/CSS/text-underline-position
[...]

PR #861, opened 2019-12-28
chore(ordered-values): added support for columns and column-rule

anikethsaha:
Ref #756

- `columns`
  In this, they takes two values, `width` and `count`, they both can be numbers and string(auto, inherit, unset etc. ) but `count` can't have `units`, and as per MDN's formal format, `width` should come first. So based on this, the function will change the order only if `unit` are mentioned.

- `column-rule`
  should be ordered like `border`

### DIFF
--- a/packages/postcss-ordered-values/src/__tests__/index.js
+++ b/packages/postcss-ordered-values/src/__tests__/index.js
@@ -585,8 +585,8 @@ test(
 test(
   'should order columns only if unit is mentioned',
   processCSS(
-    'h1 {columns: 2 auto;columns: auto 12em;columns: auto auto;columns: auto 12em auto 13em;}',
-    'h1 {columns: 2 auto;columns: 12em auto;columns: auto auto;columns: 12em 13em auto auto;}'
+    'h1 {columns: 2 auto;columns: auto 12em;columns: auto auto;}',
+    'h1 {columns: 2 auto;columns: 12em auto;columns: auto auto;}'
   )
 );
 
--- a/packages/postcss-ordered-values/src/__tests__/index.js
+++ b/packages/postcss-ordered-values/src/__tests__/index.js
@@ -573,3 +573,11 @@ test(
 );
 
 test('should use the postcss plugin api', usePostCSSPlugin(plugin()));
+
+test(
+  'should order border-block',
+  processCSS(
+    'border-inline: dashed blue 5px; border-inline-end: red solid .5em; border-block-end: dashed blue 5px; border-block: 5px solid red;border-block-start: rgba(0, 30, 105, 0.8) solid 1px;',
+    'border-inline: 5px dashed blue; border-inline-end: .5em solid red; border-block-end: 5px dashed blue; border-block: 5px solid red;border-block-start: 1px solid rgba(0, 30, 105, 0.8);'
+  )
+);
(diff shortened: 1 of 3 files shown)


packages/postcss-ordered-values/src/rules/columns.js
```diff
@@ -0,0 +1,27 @@
+import { unit } from 'postcss-value-parser';
+
+const strValues = ['auto', 'inherit', 'unset', 'initial'];
+
+export default (columns) => {
+  let newValue = { front: '', back: '' };
+  let shouldNormalize = false;
+  columns.walk((node) => {
+    const { type, value } = node;
+    if (type === 'word' && strValues.indexOf(value.toLowerCase())) {
```
jamesgeorge007:
```suggestion
    if (type === 'word' && strValues.includes(value.toLowerCase())) {
```

anikethsaha: I need to revert it, its failing the tests

jamesgeorge007: Oops! I did found couple of instances where `Array.includes()` was used for the purpose.

anikethsaha: 👍

jamesgeorge007: https://github.com/cssnano/cssnano/search?q=includes&unscoped_q=includes

anikethsaha: Not sure why then

packages/postcss-ordered-values/src/index.js
```diff
@@ -22,13 +23,19 @@ const borderRules = {
   'border-left': border,
 };
 
+const columnRules = {
+  'column-rule': border,
```
evilebottnawi: better create separate function, or rename `border` to `borderAndColumn`

packages/postcss-ordered-values/src/rules/columns.js
```diff
@@ -0,0 +1,27 @@
...
+    } else if (type === 'word') {
+      newValue.back = `${newValue.back} ${value}`;
+    }
+    // console.log({ type, value });
+  });
+  if (shouldNormalize) {
+    return `${newValue.front.trimStart()} ${newValue.back.trimStart()}`;
+  }
+  return columns;
+};
```
evilebottnawi: Remove `// console.log`

packages/postcss-ordered-values/src/index.js
anikethsaha: cool 👍

packages/postcss-ordered-values/src/rules/columns.js
anikethsaha: sure 👍

anikethsaha:
cc @evilebottnawi

packages/postcss-ordered-values/src/__tests__/index.js
```diff
@@ -581,3 +581,19 @@ test(
     'border-inline: 5px dashed blue; border-inline-end: .5em solid red; border-block-end: 5px dashed blue; border-block: 5px solid red;border-block-start: 1px solid rgba(0, 30, 105, 0.8);'
   )
 );
+
+test(
+  'should order columns only if unit is mentioned',
+  processCSS(
+    'h1 {columns: 2 auto;columns: auto 12em;columns: auto auto;columns: auto 12em auto 13em;}',
```
ben-eb: Is `auto 12em auto 13em` a valid value? I'd be wary of adding tests around invalid syntax.

anikethsaha: no it's not, sorry !!!
I will change it

anikethsaha:
cc @ben-eb @evilebottnawi

evilebottnawi:
Thanks!

PR #862, opened 2019-12-30
fix(reduce-idents): improve encode

### LINKED ISSUE #853: postcss-reduce-idents' encoder could be more efficient
Reported [from twitter](https://twitter.com/chriseppstein/status/1204125780478844928):

```
X
Y
Z
ab
bb
cb
db
```

We should start the second character from `a` not `b` and so the output should be:

```
X
Y
Z
aa
ba
ca
da
```

defcc:
fix #853


defcc:
Thanks, I totally agree with you. I could add more edge cases according to the encoder spec with hardcode later.

—-
There maybe another way to test it with a decode util，x=decode(encode(x))，but then we have to ensure both two are correct:)

PR #863, opened 2020-01-06
chore: added support for grid-*

anikethsaha:
Ref #756
Added normalization for grid-\*

- [X] grid-auto-flow `[ row | column ] || dense`
- [X] grid-column-gap `normal | <length-percentage>`
- [X] grid-row-gap `normal | <length-percentage>`
- [X] grid-column `<grid-line>+`
- [X] grid-row  `<grid-line>+`
- [X] grid-row-start  `<grid-line>`
- [X] grid-row-end  `<grid-line>`
- [X] grid-column-start  `<grid-line>`
- [X] grid-column-end  `<grid-line>`

### DIFF
--- a/packages/postcss-ordered-values/src/rules/grid.js
+++ b/packages/postcss-ordered-values/src/rules/grid.js
@@ -1,0 +1,83 @@
+export const normalizeGridAutoFlow = (gridAutoFlow) => {
+  let newValue = { front: '', back: '' };
+  let shouldNormalize = false;
+  gridAutoFlow.walk((node) => {
+    if (node.value === 'dense') {
+      shouldNormalize = true;
+      newValue.back = node.value;
+    } else if (['row', 'column'].includes(node.value.trim().toLowerCase())) {
+      shouldNormalize = true;
+      newValue.front = node.value;
+    } else {
+      shouldNormalize = false;
+    }
+  });
+  if (shouldNormalize) {
+    return `${newValue.front.trim()} ${newValue.back.trim()}`;
+  }
+  return gridAutoFlow;
+};
+
+export const normalizeGridColumnRowGap = (gridGap) => {
+  let newValue = { front: '', back: '' };
+  let shouldNormalize = false;
+  gridGap.walk((node) => {
+    // console.log(node);
+    if (node.value === 'normal') {
+      shouldNormalize = true;
+      newValue.front = node.value;
+    } else {
+      newValue.back = `${newValue.back} ${node.value}`;
+    }
+  });
+  if (shouldNormalize) {
+    return `${newValue.front.trim()} ${newValue.back.trim()}`;
+  }
+  return gridGap;
+};
+
+export const normalizeGridColumnRow = (grid) => {
+  // cant do normalization here using node, so copy it as a string
+  let gridValue = grid.toString(); // node -> string value
+  gridValue = gridValue.split('/'); // " 2 / 3 span " ->  [' 2','3 span ']
+  if (gridValue.length > 1) {
+    gridValue = gridValue.map((gridLine) => {
+      let normalizeValue = {
+        front: '',
+        back: '',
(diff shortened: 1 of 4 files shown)


packages/postcss-ordered-values/src/rules/grid.js
```diff
@@ -0,0 +1,83 @@
...
+      return `${normalizeValue.front.trim()} ${normalizeValue.back.trim()}`; // span 3
+    });
+    // returns "2 / span 3"
+    return gridValue
+      .join(' / ')
+      .trimStart()
+      .trimEnd();
+  }
+  // doing this separating if `/` is not present as while joining('/') , it will add `/` at the end
+  gridValue = gridValue.map((gridLine) => {
```
ben-eb: Prefer `return x.map( ... )` to `x = x.map( ... ); return x`

packages/postcss-ordered-values/src/rules/grid.js
```diff
@@ -0,0 +1,83 @@
...
+        if (node === 'span') {
+          normalizeValue.front = node; // span _
+        } else {
+          normalizeValue.back = `${normalizeValue.back} ${node}`; // _ 3
+        }
+      });
+      return `${normalizeValue.front.trim()} ${normalizeValue.back.trim()}`; // span 3
+    });
+    // returns "2 / span 3"
+    return gridValue
```
ben-eb: Might be nice to break this out into a utility so we can avoid the mutation on line 44 and do `return joinGridValue(gridValue.map( ... ))`

packages/postcss-ordered-values/src/rules/grid.js
```diff
@@ -0,0 +1,83 @@
...
+  });
+  if (shouldNormalize) {
+    return `${newValue.front.trim()} ${newValue.back.trim()}`;
+  }
+  return gridGap;
+};
+
+export const normalizeGridColumnRow = (grid) => {
+  // cant do normalization here using node, so copy it as a string
+  let gridValue = grid.toString(); // node -> string value
```
ben-eb: Might be nicer as:

```js
let gridValue = grid.toString().split('/');
```

Not sure the comment adds much here?

packages/postcss-ordered-values/src/rules/grid.js
```diff
@@ -0,0 +1,83 @@
...
+      .trimStart()
+      .trimEnd();
+  }
+  // doing this separating if `/` is not present as while joining('/') , it will add `/` at the end
+  gridValue = gridValue.map((gridLine) => {
+    let normalizeValue = {
+      front: '',
+      back: '',
+    };
+    gridLine = gridLine.trimStart().trimEnd();
```
ben-eb: This is what `[String.prototype.trim](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/Trim)` is for 🙂

ben-eb:
Nice work overall, have identified some minor points. 🙂

anikethsaha:
cc @evilebottnawi

packages/postcss-ordered-values/src/rules/grid.js
```diff
@@ -38,33 +40,30 @@ export const normalizeGridColumnRowGap = (gridGap) => {
...
-      .join(' / ')
-      .trimStart()
-      .trimEnd();
+    return joinGridValue(
+      gridValue.map((gridLine) => {
+        let normalizeValue = {
+          front: '',
+          back: '',
+        };
+        gridLine = gridLine.trimStart().trimEnd(); // '3 span ' -> '3 span'
```
ben-eb:
```js
gridLine = gridLine.trim();
```

Think I mentioned this before but must have gotten lost in the machine! 🙂

anikethsaha:
cc @ben-eb I made those changes you requested 👍

PR #867, opened 2020-01-16
Feat:  postcss lowercase text

anikethsaha:
Moved [`postcss-lowercase-text`](https://github.com/anikethsaha/postcss-lowercase-text) to `cssnano-preset-default`
Updated few integration tests.

PR #868, opened 2020-01-16
[Moving] :  cssnano-util-* packages  --->  cssnano-utils

anikethsaha:
## This PR introduces 🎉 

- moving all the methods, plugin(s) and their respective tests to a new utility package `cssnano-utils`
- Moved
  - `cssnano-util-raw-cache`
  - `cssnano-util-get-arguments`
  - `cssnano-util-get-match`
  - `cssnano-util-same-parent`

  to `cssnano-utils`
- added deprecation notice to the above packages
- refactor all cssnano plugins to use `cssnano-utils` instead of these four packages
- update READMEs for both presets
- checked all the cssnano internal and external plugins for the changes and all clean 😎

PR #871, opened 2020-01-28
[Breaking]: dropping node 6.9 to >=10.13.0

anikethsaha:
https://github.com/cssnano/cssnano/pull/867#discussion_r371733114


evilebottnawi:
Let's update all deps in next PR

anikethsaha:
@evilebottnawi i think merging #868 and then updating dep will be easy and those util packages will be depreciated so less work

PR #875, opened 2020-02-13
chore: added support for list-style ordering

anikethsaha:
Added list-style ordering support

Ref : https://github.com/cssnano/cssnano/issues/756

Link to refer :
https://developer.mozilla.org/en-US/docs/Web/CSS/list-style
https://developer.mozilla.org/en-US/docs/Web/CSS/list-style-image
https://developer.mozilla.org/en-US/docs/Web/CSS/list-style-position
https://developer.mozilla.org/en-US/docs/Web/CSS/list-style-type

Input

```css
h1 {
  list-style: none inside;
  list-style: unset;
  list-style: none;
  list-style: square inside;
  list-style: inside square;
  list-style: square unset;
  list-style: inside none;
  list-style: unset inside;
  list-style: circle unset none;
  list-style: circle none none;
  list-style: unset none circle;
  list-style: none outside none;
  list-style: circle url("https://mdn.m");
  list-style: circle url("https://mdn.m") none;
}
```

Output

```css
h1 {
  list-style: none inside;
  list-style: unset;
  list-style: none;
  list-style: square inside;
  list-style: square inside;
  list-style: square unset;
  list-style: none inside;
  list-style: unset inside;
  list-style: circle unset none;
  list-style: circle none none;
  list-style: unset none circle;
  list-style: none outside none;
  list-style: circle url("https://mdn.m");
  list-style: circle none url("https://mdn.m");
}
```

### DIFF
--- a/packages/postcss-ordered-values/src/__tests__/index.js
+++ b/packages/postcss-ordered-values/src/__tests__/index.js
@@ -669,3 +669,86 @@ test(
     'grid-column-end: 2 /  4; grid-column-end: span 2 /  7; grid-column-end: auto;grid-column-end: 3;grid-column-end: custom-indent-name /  3;'
   )
 );
+
+test(
+  'should order list-style 1',
+  processCSS('ul{list-style: unset}', 'ul{list-style: unset}')
+);
+
+test(
+  'should order list-style 2',
+  processCSS('ul{list-style: none}', 'ul{list-style: none}')
+);
+
+test(
+  'should order list-style 3',
+  processCSS('ul{list-style: square inside}', 'ul{list-style: square inside}')
+);
+
+test(
+  'should order list-style 4',
+  processCSS('ul{list-style: inside square}', 'ul{list-style: square inside}')
+);
+
+test(
+  'should order list-style 5',
+  processCSS('ul{list-style: square unset}', 'ul{list-style: square unset}')
+);
+
+test(
+  'should order list-style 6',
+  processCSS('ul{list-style: inside none}', 'ul{list-style: none inside}')
+  /**
+   *  This change is CORRECT as
+   *  list-style: inside none;
+          list-style-position: inside;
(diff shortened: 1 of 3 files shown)


packages/postcss-ordered-values/src/rules/listStyle.js
```diff
@@ -0,0 +1,45 @@
...
+
+const definedTypes = [
+  'disc',
+  'circle',
+  'square',
+  'decimal',
+  'georgian',
+  'trad-chinese-informal',
+  'kannada',
+];
```
evilebottnawi, line 11: I think in this case we better not store all the values, in the future there may be more of them, is it possible to rewrite it?

```if
if (definedPosition.includes(decl.value)) {
} else if (decl.value === 'none')
else if (/* Something */) {
} else {
  // its a type field
  order.type = `${order.type} ${decl.value}`;
}
```

anikethsaha: may be need to fetch them from here. https://github.com/mdn/browser-compat-data/blob/master/css/properties/list-style-type.json

anikethsaha: done

packages/postcss-ordered-values/src/__tests__/index.js
```diff
@@ -669,3 +669,86 @@ test(
...
+  )
+);
+
+test(
+  'should order list-style 12',
+  processCSS(
+    'ul{list-style: circle url("https://mdn.mozillademos.org/files/11981/starsolid.gif") none}',
+    'ul{list-style: circle none  url("https://mdn.mozillademos.org/files/11981/starsolid.gif")}'
+  )
+);
```
evilebottnawi, line 754: Can we add test for:
```
list-style: unknown unset none
```

PR #731, opened 2019-04-18
fix(postcss-fix-rules): prevent merging of :host(tag) and tag

MapTo0:
- Combining `:host(tag) { color: red }` and `tag { color: red }` to
`:host(tag), tag { color: red }` is not a valid syntax.

Fixes: https://github.com/cssnano/cssnano/issues/730

PR #876, opened 2020-02-13
fix: prevent merging of :host(tag) and tag

### LINKED ISSUE #730: mergeRules merges :host(tagname) with tagname
If we have the following CSS in a custom element's shadow root:

````css
:host(custom-element) {
  width: 200px;
  height: 40px;
  background: red;
}

custom-element {
  width: 200px;
  height: 40px;
  background: red;
}
````

becomes

```css
:host(custom-element),
custom-element {
  width: 200px;
  height: 40px;
  background: red;
}
```

both selectors are getting merged which is not a valid syntax.

anikethsaha:
fixes #730
closes #731


anikethsaha:
increasing jest timeout to 30s to fix this and for future changes as well

anikethsaha:
@evilebottnawi 

I think this fix is not required as we are already checking the compatibility. 
I could write a test for `IE 11` env in `browserlist`, but I think `caniuse-lite`  doesnt have the `:host` feature in their [feature data](https://github.com/ben-eb/caniuse-lite/tree/master/data/features) and also tried running querying with `:host`, `css-host` , `host-selector` and `host selector` but it will through `ReferenceError: Please provide a proper feature name. Cannot find css-host`.

So I don't think there is a way as of now to test it with the compatibility. 

We can keep an eye on [this list](https://github.com/ben-eb/caniuse-lite/tree/master/data/features) and when they add `css-host` we can write a test for this issue. 

What do you think

anikethsaha:
Okay I guess `host` is in [level 1 CSS Selector](https://drafts.csswg.org/css-scoping/#host-selector) ?

And I think the `caniuse-lite` query for that is [css-selection](https://github.com/ben-eb/caniuse-lite/blob/master/data/features/css-selection.js)

evilebottnawi:
@anikethsaha :+1:

anikethsaha:
Closing this as this is already done. Just need to add test, I will add that later once got to know the query for level 1 selectors in `caniuse-lite`

PR #881, opened 2020-02-21
Resolve Opacity Bug

genemecija:
Resolving bug in which opacity values defined as percentages are being converted to 1%.

### DIFF
--- a/packages/postcss-convert-values/src/__tests__/index.js
+++ b/packages/postcss-convert-values/src/__tests__/index.js
@@ -67,6 +67,11 @@ test(
   processCSS('h1{width:100.00%}', 'h1{width:100%}')
 );
 
+test(
+  'should preserve opacities defined as percentages',
+  passthroughCSS('h1{opacity:100%}')
+);
+
 test('should not mangle flex basis', passthroughCSS('h1{flex-basis:0%}'));
 
 test('should not mangle flex basis (2)', passthroughCSS('h1{FLEX-BASIC:0%}'));
--- a/packages/postcss-convert-values/src/index.js
+++ b/packages/postcss-convert-values/src/index.js
@@ -53,7 +53,7 @@ function clampOpacity(node) {
   }
   let num = Number(pair.number);
   if (num > 1) {
-    node.value = 1 + pair.unit;
+    node.value = pair.unit === '%' ? num + pair.unit : 1 + pair.unit;
   } else if (num < 0) {
     node.value = 0 + pair.unit;
   }


packages/postcss-convert-values/src/__tests__/index.js
```diff
@@ -67,6 +67,11 @@ test(
   processCSS('h1{width:100.00%}', 'h1{width:100%}')
 );
 
+test(
+  'should preserve opacities defined as percentages',
+  passthroughCSS('h1{opacity:100%}')
+);
+
```
anikethsaha: Can you add some more.

may be like these
⬇️ 

```js
test(
  'should convert the opacity to 1px',
  processCSS('h1{opacity : 80px}', 'h1{opacity: 1px}')
);
test(
  'should convert the opacity from 80% -> 80%',
  processCSS('h1{opacity :80%}', 'h1{opacity :80%}')
);
```

Also check this one 
⬇️ 
```js
test(
  'should convert the opacity without any unit',
  passthroughCSS('h1{opacity :80}')    // expected is 80
);

```

genemecija: Regarding your tests:
```
test(
  'should convert the opacity to 1px',
  processCSS('h1{opacity : 80px}', 'h1{opacity: 1px}')
);
```
'px' is not a valid unit, and I opted not to check for invalid units as that's not the tool's purpose (minification vs linting)

---
```
test(
  'should convert the opacity from 80% -> 80%',
  processCSS('h1{opacity :80%}', 'h1{opacity :80%}')
);
```
If I'm not mistaken, `processCSS('h1{opacity :80%}', 'h1{opacity :80%}')` would be the same test as `passthroughCSS('h1{opacity:80%}')` or the test I created, `passthroughCSS('h1{opacity:100%}')`. Please correct me if I'm wrong about that.

---
```
test(
  'should convert the opacity without any unit',
  passthroughCSS('h1{opacity :80}')    // expected is 80
);
```
Any number above '1.0' is invalid and would be converted to '1'. There's a check for this at line 281:
```
test(
  'should clamp opacity to 1 maximum',
  processCSS(
    'h1{opacity:150;opacity:15;opacity:1.5}',
    'h1{opacity:1;opacity:1;opacity:1}'
  )
);
```
---
The only other tests I can think that aren't already tested for are for validating unit type, but that would be a linter's job.

anikethsaha: > Regarding your tests:
> ```js
> test(
>  'should convert the opacity to 1px',
>  processCSS('h1{opacity : 80px}', 'h1{opacity: 1px}')
> );
> ```
> 
> 'px' is not a valid unit, and I opted not to check for invalid units as that's not the tool's purpose (minification vs linting)

Invalid CSS should lean to either no minification or invalid minification. This is an example of invalid minification. 
You can keep it or omit it. either looks good. 
_but I would check the behavior by keeping it for future use_

> If I'm not mistaken, processCSS('h1{opacity :80%}', 'h1{opacity :80%}') would be the same test as passthroughCSS('h1{opacity:80%}') or the test I created, passthroughCSS('h1{opacity:100%}'). Please correct me if I'm wrong about that.

Yea, but I prefer using `processCSS` as its more visible. again your choice to use either of them. but they both are fine 👍 

> Any number above '1.0' is invalid and would be converted to '1'. There's a check for this at line 281:

if its done,  leave it then 👍 

> The only other tests I can think that aren't already tested for are for validating unit type, but that would be a linter's job.

Feel free to choose whether to test it or not. @evilebottnawi what you think?

anikethsaha:
> What is an original problem, I lose context, can you provide example? Or a link on issue?

This is the current behavior 

`opacity : 40%` -> `opacity : 1%`

It is wrong minification

evilebottnawi:
@genemecija `opacity : 40%` is invalid css syntax, why you need that line?

anikethsaha:
> @genemecija `opacity : 40%` is invalid css syntax, why you need that line?

@evilebottnawi I think this is an correct syntax. I tried on developer tools. its working

evilebottnawi:
hm, yes, you are right https://developer.mozilla.org/en-US/docs/Web/CSS/opacity, never use `%` :smile:

packages/postcss-convert-values/src/__tests__/index.js
evilebottnawi: I prefer to use `passthroughCSS` when values was not changed, I think in future we refactor tests on snapshots, because it is more easy to read

herecydev:
@evilebottnawi was there a release for this patch?

evilebottnawi:
@herecydev yes, in future, still a lot of work

SylvainGarrigues:
Any possibility for a hot bug fix release instead of waiting for a new formal release with other features? This is breaking UI for sites in production which rely on cssnano. 

X% (unminified) -> 1% (minified) means elements become invisible in production builds and not in development builds. Can drive people crazy.

anikethsaha:
it will be done soon. 
in the meantime use [`postcss-ignore-plugin`](https://www.npmjs.com/package/postcss-ignore-plugin) or I would suggest disabling this plugin

svendlarsen:
@anikethsaha @evilebottnawi Please reconsider making a separate patch release for this bugfix! Three months is a long time between merge and release for a fix of this nature.

anikethsaha:
it will be released soon, https://github.com/cssnano/cssnano/pull/881#issuecomment-613595925

klesun:
um, bump? Guys, your lib is used by a decent amount of people, would you consider making the hotfix release, please?

anikethsaha:
Please use [`cssnano@nightly`](https://www.npmjs.com/package/cssnano/v/4.0.0-nightly.2020.8.10) in the meantime

anikethsaha:
nightly is up-to-date with the `master`

Issue #642, opened 2018-10-18
mergeRules should not merge unknown pseudo selectors


fracmak:
It appears that if a browser (tested in Chrome 69, Firefox 60, Safari 12, Edge 16) does not recognize a pseudo selector, it drops the entire selector, including any combined selectors. 

Example:

https://codepen.io/anon/pen/ZqxNEq

![image](https://user-images.githubusercontent.com/378178/47170937-6e962200-d2d5-11e8-9664-35889a16eb5f.png)

saulhardman:
I've recently come across this issue myself and I'm wondering if it's currently possible to disable optimisations on a line-by-line basis? E.g. like ESLint's `// eslint-disable-line` comment format.

evilebottnawi:
@saulhardman no, it is hard to implement using comments, but PR welcome

saulhardman:
@evilebottnawi I figured as much, thanks for your response 👍

ben-eb:
@evilebottnawi We have a system for handling mergeable pseudo elements here:

https://github.com/cssnano/cssnano/blob/master/packages/postcss-merge-rules/src/lib/ensureCompatibility.js#L13

Perhaps this should be changed; currently this list prevents a merge for known pseudo selectors, but allows unknown ones through. If we only allow a merge on known pseudo selectors, we will have less bug reports on broken behaviour and hopefully more reports of acceptable pseudo selector merging.

In any case this is a matter of adding `':focus-visible': 'css-focus-visible'` to that list.

evilebottnawi:
Good idea, can you take care about this?

ben-eb:
@evilebottnawi Probably easier if we fix the immediate bug and then can look into changing behaviour later?

evilebottnawi:
Sounds good, need open issue about this or update currently

felixfbecker:
`:focus-visible` is now on the list (so rules using it will be merged), but not supported by any browsers other than Firefox, meaning CSSNano currently breaks Chrome when e.g. using a `focus-visible` polyfill that creates duplicate rules for `:focus-visible` and a polyfilled `.focus-visible` class. Is there a way to instruct CSSNano to not consider `:focus-visible` "supported"?

anikethsaha:
are you using the nightly ?

cause this has been fixed in https://github.com/cssnano/cssnano/pull/882 but not released yet.

anikethsaha:
> :focus-visible is now on the list (so rules using it will be merged)

[...]

PR #882, opened 2020-02-23
chore: fixed focus-visible issue and added tests

anikethsaha:
closes https://github.com/cssnano/cssnano/pull/781
fixes https://github.com/cssnano/cssnano/issues/642

### DIFF
--- a/packages/postcss-merge-rules/src/__tests__/index.js
+++ b/packages/postcss-merge-rules/src/__tests__/index.js
@@ -786,3 +786,11 @@ test(
     '.a{color:red;display:flex;font-size:10px;}.a,.c{unicode-bidi:normal;}.c{all:unset;color:red;display:flex;font-size:10px;}'
   )
 );
+
+test(
+  'should not merge :focus-visible',
+  processCSS(
+    'a{color : green;} a:focus-visible{ color : green;} a:focus-visible{ background : red}',
+    'a{color : green;} a:focus-visible{ color : green;} a:focus-visible{ background : red}'
+  )
+);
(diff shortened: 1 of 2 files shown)


packages/postcss-merge-rules/src/__tests__/index.js
```diff
@@ -786,3 +786,11 @@ test(
...
   )
 );
+
+test(
+  'should not merge :focus-visible',
+  processCSS(
+    'a{color : green;} a:focus-visible{ color : green;} a:focus-visible{ background : red}',
+    'a{color : green;} a:focus-visible{ color : green;} a:focus-visible{ background : red}'
+  )
+);
```
evilebottnawi: Can we add test where we merge them? Here example of the test https://github.com/cssnano/cssnano/blob/5db8a8e5aa9a6a744b3f6c34dc86d992d73bf1c7/packages/postcss-merge-rules/src/__tests__/index.js#L517

anikethsaha: sure , 
I will add them .....

anikethsaha: this is not  supported fully by any browser as of now

image

evilebottnawi: Yes, just test

anikethsaha: which browser version to use in order to merge it ? none of them are supporting it. So it can be merged

evilebottnawi: Oh, yes, I see

evilebottnawi:
Thanks!

PR #883, opened 2020-02-23
fix: fixed merging of unsupported pseudos by caniuse and browserlist

anikethsaha:
This change won't merge rules (having pseudo-classes) which are not either supported by caniuse in browserlist 's browsers. 

The bug was even they were not supported, the compatibility was not changed which leads to false merging.


alexander-akait (commented):
Looks good, what is changed in snapshots?

anikethsaha:
Integration fixtures. there where some of the places in the framework fixtures where it was merging unsupported rules.

Output of `build:integration`

evilebottnawi:
Can you look on snapshot changes, and write them here (no need all, just couple)

anikethsaha:
for `cancise.css`
---

It didnt merge the `::-webkit-slider-thumb` selector (not supported from `caniuse` in `browserlist` 's default env)

image

image



`gumpy.css`

`:visited` not supported by us. not in the list [**`here`**](https://github.com/cssnano/cssnano/blob/master/packages/postcss-merge-rules/src/lib/ensureCompatibility.js#L13-L57)

image

evilebottnawi:
Can we fix `visited` here?

anikethsaha:
> Can we fix `visited` here?

I will surely fix it , but I dont know the name for `caniuse` to search it.  

I tried 
`:visited`
`visited`
`css-visited`

but for all of these, `caniuse` api is return error with `cannot find name ...`

image

image

image


image


Do you know the feature name for this. ?

evilebottnawi:
https://caniuse.com/#feat=mdn-css_selectors_visited, think we should try `css_selectors_visited`

evilebottnawi:
https://caniuse.com/#feat=mdn-css_selectors_visited, think we should try `css_selectors_visited`

anikethsaha:
no luck 

image

packages/postcss-merge-rules/src/lib/ensureCompatibility.js
```diff
@@ -54,6 +54,7 @@ export const pseudoElements = {
   '::marker': 'css-marker-pseudo',
   '::placeholder': 'css-placeholder',
   '::selection': 'css-selection',
+  '::-moz-selection': 'css-selection',
```
evilebottnawi: Let's put all elements and classes from https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-elements and https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-classes

Also please use http://api.postcss.org/vendor.html#.unprefixed for removing vendor prefixes

evilebottnawi:
I will look on visited in near future

packages/postcss-merge-rules/src/lib/ensureCompatibility.js
anikethsaha: > Let's put all elements and classes from https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-elements and https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-classes

I will try to do this for elements/classes for which the `caniuse` api is responding,  Cause for most of them, `caniuse` might have different format name, which I dont know where to look for other than requesting manually from their api.

evilebottnawi: Set them unsupported by default, then I will look on list and give advices

anikethsaha: > Set them unsupported by default, then I will look on list and give advices

if they are not on the list, this PR changes are considering them unsupported.

we can even add those elements incrementally in future. Cause by default they are unsupported

anikethsaha: Instead of increasing the content of this PR diff, 

Can I add these elements in future PR after merging this? I will add the  `verdor.unprefixed` in this PR and rest in the next PR?

evilebottnawi: No, it is a breaking change, we should solve it here

anikethsaha: ok coool

anikethsaha:
cc @evilebottnawi

packages/postcss-merge-rules/src/lib/ensureCompatibility.js
```diff
@@ -55,6 +56,47 @@ export const pseudoElements = {
...
+  // pseudo-elements
+  '::after (:after)': '::after',
+
+  '::before (:before)': '::before',
+  '::first-letter (:first-letter)': 'first-letter',
+  '::first-line (:first-line)': 'first-line',
+  '::grammar-error': 'grammar-error',
+
+  '::part()': '::part',
+  '::slotted()': '::slotted',
```
evilebottnawi, line 98: Why some have `()` parenthesis, some doesn't have? It is bug?

anikethsaha: let me check

anikethsaha:
I think,  `visited` is at [selector level 3](https://www.w3.org/TR/selectors-3/#the-link-pseudo-classes-link-and-visited) ?

packages/postcss-merge-rules/src/lib/ensureCompatibility.js
anikethsaha: Ok these should not have `()` but these are `level 4 CSS Selector` except for the `visited`.

And I think level 4 is not supported by `caniuse-lite` , so this all will be removed

evilebottnawi:
@anikethsaha Yes :+1:

anikethsaha:
there were level 4 CSS selectors in early commits of this PR, I had to remove them as `caniuse-lite` doesnt support them i guess

alexander-akait (dismissed):

evilebottnawi:
@anikethsaha So we do not merge selector for 4 level?

anikethsaha:
currently no!

Cause we dont know their browser support so that may lead to wrong minification.
Once `caniuse-lite` supports that, I will add them.

evilebottnawi:
We have some of them https://caniuse.com/#search=selector%204

anikethsaha:
I guess `caniuse-lite` API doesn't support that. 

I couldn't locate the [selectors here](https://github.com/ben-eb/caniuse-lite/tree/master/data/features). 

_(may be I missed something, let me know)_

evilebottnawi:
https://github.com/ben-eb/caniuse-lite/blob/master/data/features/css-not-sel-list.js
https://github.com/ben-eb/caniuse-lite/blob/master/data/features/css-any-link.js

anikethsaha:
> https://github.com/ben-eb/caniuse-lite/blob/master/data/features/css-not-sel-list.js
> https://github.com/ben-eb/caniuse-lite/blob/master/data/features/css-any-link.js

ohh great, I thought they might have packed [like level 2,3 are](https://github.com/ben-eb/caniuse-lite/blob/master/data/features/css-sel3.js)

I will add those which are supported.

packages/postcss-merge-rules/src/lib/ensureCompatibility.js
```diff
@@ -55,6 +56,10 @@ export const pseudoElements = {
   '::marker': 'css-marker-pseudo',
   '::placeholder': 'css-placeholder',
   '::selection': 'css-selection',
+  ':visited': cssSel3,
+  ':any-link': 'css-any-link',
+  ':read-only': 'css-read-only-write',
+  ':read-write': 'css-read-only-write',
```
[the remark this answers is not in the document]
evilebottnawi: Can we create script to generate this stuff from `caniuse-api`? It is allow to validate we nothing is missed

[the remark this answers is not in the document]
anikethsaha: i will check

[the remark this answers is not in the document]
evilebottnawi: Thanks, I am afraid we can missing something, so we need script to generate and validate it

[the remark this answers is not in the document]
anikethsaha: what I am thinking is to simply query using the code name like for `:react-write` , query using `react-write`. but two issue can come
- there may be more that one result, and we dont know which is correct
- there may be no result as the query can be different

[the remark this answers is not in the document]
evilebottnawi: Maybe we can parse all existing pseudo from `css-whatwg`?

ludofischer:
I think continuing this work could also fix https://github.com/cssnano/cssnano/issues/999

PR #891, opened 2020-03-18
chore: updated dep and auditing fixes

anikethsaha:
closes #889 #888 

Updated dep and auditing

### DIFF
--- a/packages/css-size/yarn.lock
+++ b/packages/css-size/yarn.lock
@@ -2,6 +2,16 @@
 # yarn lockfile v1
 
 
+"@types/q@^1.5.1":
+  version "1.5.2"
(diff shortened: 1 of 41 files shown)


anikethsaha:
as `cssnano-utils` is not published as of now, I am getting package not found for cssnano plugins.
Should I link all using `yarn link` and then proceed with upgrading each ?

evilebottnawi:
Yes

anikethsaha:
@evilebottnawi I couldnt able to upgrade the dep of packages which had `cssnano-utils` as dep even after running `lerna link` !

any suggestions ?

evilebottnawi:
> got the same error as package not found

We need publish package, feel free to do it and add me as admin to npm

anikethsaha:
> > got the same error as package not found
> 
> We need publish package, feel free to do it and add me as admin to npm

ok I will publish that and then will update this PR

PR #894, opened 2020-03-31
fix: units being removed in math functions

wongjn:
When using `min()`, `max()` or `clamp()`, `0` values get their units stripped, making the value invalid. This pull request preserves any units within these CSS functions.

### DIFF
--- a/packages/postcss-convert-values/src/index.js
+++ b/packages/postcss-convert-values/src/index.js
@@ -106,7 +106,12 @@ function transform(opts, decl) {
           });
           return false;
         }
-        if (lowerCasedValue === 'url') {
+        if (
+          lowerCasedValue === 'url' ||
+          lowerCasedValue === 'min' ||
+          lowerCasedValue === 'max' ||
+          lowerCasedValue === 'clamp'
+        ) {
           return false;
         }
       }
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -591,6 +591,15 @@
       "contributions": [
         "platform"
       ]
+    },
+    {
+      "login": "wongjn",
+      "name": "wongjn",
+      "avatar_url": "https://avatars2.githubusercontent.com/u/11310624?v=4",
+      "profile": "https://github.com/wongjn",
+      "contributions": [
+        "code"
+      ]
     }
   ],
   "repoType": "github"
--- a/jest.config.js
+++ b/jest.config.js
@@ -10,4 +10,5 @@ module.exports = {
     '_processCss.js',
     '_webpack.config.js',
   ],
+  testTimeout: 30000,
 };
--- a/packages/postcss-convert-values/src/__tests__/index.js
+++ b/packages/postcss-convert-values/src/__tests__/index.js
@@ -351,6 +351,12 @@ test(
   passthroughCSS('h1{line-height:0rem}')
 );
 
+test('should keep unit in max()', passthroughCSS('h1{margin:max(0px)}'));
+
+test('should keep unit in min()', passthroughCSS('h1{margin:min(0px)}'));
+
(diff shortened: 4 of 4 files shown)


anikethsaha:
Please fix the CI, else LGTM

packages/postcss-convert-values/src/index.js
```diff
@@ -106,7 +106,12 @@ function transform(opts, decl) {
           });
           return false;
         }
-        if (lowerCasedValue === 'url') {
+        if (
+          lowerCasedValue === 'url' ||
+          lowerCasedValue === 'min' ||
+          lowerCasedValue === 'max' ||
+          lowerCasedValue === 'clamp'
```
evilebottnawi: We need to look on CSS spec for all possible values

wongjn:
Most of the failing CI seems to be disconnected from the changes I made, so not sure if it is in the scope of this pull request to fix them?

packages/postcss-convert-values/src/index.js
wongjn: `min()`, `max()` and `clamp()` were introduced in [CSS Values Module Level 4](https://www.w3.org/TR/css-values-4/#calc-notation), although I don't know if there are other CSS functions in other modules that don't accept unitless `0`.

anikethsaha:
jest timeout has to be increased .. I guess 30s might be enough.

wongjn:
Would it be appropriate to do that in this PR?

evilebottnawi:
Yes

evilebottnawi:
Thanks!

packages/postcss-convert-values/src/index.js
evilebottnawi: Hm, what about calc?

wongjn: I've had a look and units on 0 values are preserved already in `calc()` — maybe these should be moved to the same place?

evilebottnawi: It is strange, I think logic for calc should be here too

wongjn: I think that would introduce a lot of regressions in the current transformations in `calc`.

How about something like this to bring the functions together:

```diff
         if (
           lowerCasedValue === 'calc' ||
+          lowerCasedValue === 'min' ||
+          lowerCasedValue === 'max' ||
+          lowerCasedValue === 'clamp' ||
           lowerCasedValue === 'hsl' ||
           lowerCasedValue === 'hsla'
         ) {
           walk(node.nodes, (n) => {
             if (n.type === 'word') {
               parseWord(n, opts, true);
             }
           });
           return false;
         }
-        if (
-          lowerCasedValue === 'url' ||
-          lowerCasedValue === 'min' ||
-          lowerCasedValue === 'max' ||
-          lowerCasedValue === 'clamp'
-        ) {
+        if (lowerCasedValue === 'url') {
           return false;
         }
       }
```

evilebottnawi: Yes, let's do it

wongjn: New pull request?

evilebottnawi: Yep

helins-io:
I believe this is an important fix. I've just spend 3 hours trying to figure this one out in a project. It is not a well-known fact fact a unitless 0's mess with function such as `max` or `clamp` and you wouldn't expect cssnano to misbehave with something like that.

Meanwhile, a hack would be to use a very low value (eg. 0.01px) or just "0px" via a custom properties. But those are just hacks.

srsly4:
Would it be possible to backport this fix into the 4.x.x branch?

At the moment there is no possibility to fix this while working on a react-script project without upgrading the whole webpack.

PR #895, opened 2020-04-01
fix: Keep math functions together

wongjn:
Follow up to #894


anikethsaha:
can you add some tests if possible .

anikethsaha:
thanks 👍

PR #899, opened 2020-04-09
chore: update contributors list

jamesgeorge007:
Added myself to the list.


sigveio:
Hey @jamesgeorge007 - apologies for this PR being left without a reply. I'm guessing it was perhaps deemed a bit unnecessary to create a separate PR for it - but that's not your fault; our contributing documentation/guidelines should mention it in that case. So I just wanted to follow up and thank you for your contributions - they are appreciated. :-)

sigveio:
/cc @ludofischer / @alexander-akait 

I will probably submit a PR in the not so distant future with some proposed changes to contributing guidelines and issue/PR templates (#1157) - so in relation to that, I'm wondering; how do we want to handle adding of contributors in the future? Should contributors add themselves alongside a contribution as a rule, or is a separate PR like this also fine?

ludofischer:
>  Should contributors add themselves alongside a contribution as a rule, or is a separate PR like this also fine?

There aren't that many contributors appearing each month, so separate PR does not add much work. But I find contrinutors-cli cumbersome: if someone does not add themselves, should we add them anyway? Is there a 1% chance that someone might not wish to appear there? Maybe we could replace it by only printing the Github usernames extracted from the commits.

sigveio:
> There aren't that many contributors appearing each month, so separate PR does not add much work. But I find contrinutors-cli cumbersome: if someone does not add themselves, should we add them anyway? Is there a 1% chance that someone might not wish to appear there? Maybe we could replace it by only printing the Github usernames extracted from the commits.

Automating it was the thought that crossed my mind first, but I think "self-adding" is better in order to give people a choice as you suggest. I see they also have a [bot](https://allcontributors.org/docs/en/bot/usage) that could reduce the need to interact with the cli tool, which creates a separate PR for the actual change. So I suppose we could add people when we see they are a first time contributor; but wait with merging the PR until they approve?

PR #905, opened 2020-05-07
fix:  writing-mode: initial != horizontal-tb 

ICELI:
https://stackoverflow.com/questions/59560622/writing-mode-initial-horizontal-tb-at-chrome

[demo](https://jsrun.pro/AjWKp/edit)

### DIFF
--- a/packages/postcss-reduce-initial/data/toInitial.json
+++ b/packages/postcss-reduce-initial/data/toInitial.json
@@ -29,6 +29,5 @@
   "text-emphasis-position": "over right",
   "transform-box": "border-box",
   "transform-origin": "50% 50% 0",
-  "vertical-align": "baseline",
-  "writing-mode": "horizontal-tb"
+  "vertical-align": "baseline"
 }


anikethsaha:
initial for `writing-mode` is `horizontal-tb` [ref this](https://drafts.csswg.org/css-writing-modes-4/#block-flow)

ICELI:
> initial for `writing-mode` is `horizontal-tb` [ref this](https://drafts.csswg.org/css-writing-modes-4/#block-flow)

First of all, thanks for such a quick reply.
Yes, I know that. But in that case, they did behave inconsistently in the Chrome browser. How do I solve it, give up chrome or cssnano, is there an easy way to solve this problem.

ICELI:
In version 81 of the Chrome browser, this can be solved as follows:
`` `
-webkit-writing-mode: horizontal-tb;
     writing-mode: initial;
`` `
But in the Chrome 83 (beta) version, it does not work.

anikethsaha:
Thanks for the details 👍 , 
We are using `caniuse-api` to check the support, it is working as expected. 
perhaps, you can change the version in `browserlist` config for chrome and try,
also,
we don't want to move away from official specs. So personally I would wait until there is any changes in the spec. 

in the meantime, you can 
- use [postcss-ignore-plugin](https://github.com/anikethsaha/postcss-ignore-plugin)
- or , [disable](https://github.com/cssnano/cssnano/blob/master/site/content/guides/presets.md#excluding-transforms) the reduce initial plugin in cssnano option

PR #908, opened 2020-05-18
Add failing test on animation with keyframe value that looks like hex

SkReD:


ludofischer:
Already merged in https://github.com/cssnano/cssnano/commit/e64272fa229dca709c8400f760a850cbf7739e24

Issue #428, opened 2017-12-01
Low hanging fruits


xem:
Hello,

I'm here thanks to @tomhodgins.

First, I'd like to salute your great work on cssnano. Your tool is probably the best one out there, and (imo) only lacks a live CSS minifier. Sometimes, it's great to have that instead of command-line only, but no big deal... :)

So I'm making a [mini CSS minifier](https://github.com/xem/miniMinifier) just for fun, and in my recent tests, I compared the output of my app to cssnano's output, and remarked a lot of easily-removable chars (mostly spaces) that cssnano doesn't remove in CSS code, as well as a few other "safe" transformations that aren't performed.

Are you aware of these cases?

Would you be interested to have a test file showing all the low-hanging fruits that cssnano doesn't seize?
(my final test file will be ready in a few days, and I have no problem sharing it with you, if you want).

Cheers!

michael-ciniawsky:
Sure things like that are always appreciated I guess 🙃. The rumors say `cssnano` is also actively looking for maintainers atm

xem:
I'm afraid I'm not good enough to maintain other peoples' code, that's why I spend all my time reinventing wheels on my Github :) 

Anyway, as promised, here are the easy wins, that my tiny minifier does and cssnano doesn't (afaik):

Please tell me if some shortcuts I suggest here are not valid!


````
(cssnano output => expected output)
@media (max-width:500px) => @media(max-width:500px)
:nth-of-type(- 2n+1) => :nth-of-type(-2n+1)
box-shadow:inset 0 2px #000 => box-shadow:0 2px#000 inset
calc ((2vh - 2px + 2cm)/2 * 4) => calc((2vh - 2px + 2cm)/2*4)
border-color:#111 #222 #333 #444 => border-color:#111#222#333#444
@media screen and (min-resolution:0dpi){ => @media screen and(min-resolution:0){
0deg => 0
0rad => 0
0grad => 0
0turn => 0
background:RGBA(33,22,11,1) => background:#21160b
hsla(11,22,33,1) => hsl(11,22,33)
rgba (0,0,0,0)} => rgba(0,0,0,0)
rgba(0,0,0,0) => transparent
:root => html
margin:100mm -100mm => margin:10cm -10cm
url(https://test.com/test.png) => url(//test.com/test.png)
@media (width >= 600px) => @media(width>=600px)
````

infinnie:
`:root` may be applied to the `svg` element or other XML elements that are root nodes.

PR #921, opened 2020-06-28
[normalize-whitespace]: removal of whitespace in at-rule and params

anikethsaha:
ref #428 

Some of the transforms

```css
@media and (...) => @media and(...)
@media screen and () => @media screen and()
@media ( max-width:   50px ) => @media (max-width:50px)
```


packages/postcss-normalize-whitespace/src/index.js
```diff
@@ -42,6 +42,55 @@ export default plugin('postcss-normalize-whitespace', () => {
...
+         */
+        paramClone = paramClone.replace(/\(.*?\)/gi, (i) =>
+          i.replace(/ +/g, '')
+        );
+
+        // convert all multiple spaces to one in params
+        paramClone = paramClone.replace(/ +/g, ' ');
+        paramClone = paramClone.trim();
+        node.params = paramClone;
+      }
```
evilebottnawi, line 85: Using `regex` is unsafe here, we need parser for this stuff, there are a lot of edge cases when we broken syntax

evilebottnawi:
@media () => @media() -> need test on IE
@media and () => @media and() -> invalid
@media ( max-width:   50px ) => @media (max-width:50px) -> good idea, but we should uses parser or very very very strict regex + many tests
@support () , @support () => @support(),@support() -> i think it is invalid syntax, no examples in spec, but maybe I wrong

packages/postcss-normalize-whitespace/src/index.js
anikethsaha: I couldn't find any parser that parser at rules params.

anikethsaha:
> @media () => @media() -> need test on IE

ok

> @media and () => @media and() -> invalid

changed the  PR description, 

> @support () , @support () => @support(),@support()

yes invalid, I removed it from the PR description, I couldn't find any example in the spec, but postcss should throw parsing error right if this is an invalid syntax ?

> @media ( max-width: 50px ) => @media (max-width:50px) -> good idea, but we should uses parser or very very very strict regex + many tests

I think the current regex will only convert multiple spaces to a single one. I think it is safe, it won't do any changes to the newlines, neither to tabs

evilebottnawi:
> yes invalid, I removed it from the PR description, I couldn't find any example in the spec, but postcss should throw parsing error right if this is an invalid syntax ?

If it is invalid syntax - yes, but I am not usre we need read spec

evilebottnawi:
> I think the current regex will only convert multiple spaces to a single one. I think it is safe, it won't do any changes to the newlines, neither to tabs

I am afraid about `var()` function, it can be part of `@supports`

evilebottnawi:
We can use `postcss-value-parser` for this, because syntax is same

anikethsaha:
> If it is invalid syntax - yes, but I am not usre we need read spec

the spec [says](https://drafts.csswg.org/css-conditional-3/#at-supports), 

```
@supports <supports-condition> {
  <stylesheet>
}
```

so I suppose there can be nested at-rules, but no atules in `<supports-condition>`

> I am afraid about `var()` function, it can be part of `@supports`

`var` does supports spaces inside of the `(..)` so even removing them should not be unsafe. I will add some test to show that. 

> We can use `postcss-value-parser` for this, because syntax is same

does it supports parsing at rules ?

anikethsaha:
@evilebottnawi I think there is some issue if we go with AST.
In AST, it parse the spaces as a node, so for example

```css
@media screen and ( max-width: 500px) {
}
```

this is the tree

```
    ValueParser {
      nodes: [
        { type: 'word', sourceIndex: 0, value: 'screen' },
        { type: 'space', sourceIndex: 6, value: ' ' },
        { type: 'word', sourceIndex: 7, value: 'and' },
        { type: 'space', sourceIndex: 10, value: ' ' },
        {
          type: 'function',
          sourceIndex: 11,
          value: '',
          before: ' ',
          after: '',
          nodes: [
            { type: 'word', sourceIndex: 13, value: 'max-width' },
            { type: 'div', sourceIndex: 22, value: ':', before: '', after: '' },
            { type: 'word', sourceIndex: 23, value: '500px' }
         ]
        }
      ]
    }
```

it would be unsafe to remove all the space. 

we can remove the whitespace inside of the function i.e inside of the parens `(...)` but not it would be kind of hacky for `and (`
we need to check spaces before `(` so while traversing the AST, if we encounter `(` we can't go back to the previous node.

alexander-akait (commented):
Can you test this compression on IE8/IE10/IE11, need to investigate

evilebottnawi:
I am afraid `@media () => @media()` is not valid

anikethsaha:
> Can you test this compression on IE8/IE10/IE11, need to investigate

I have added test for ie11. should we still support ie10, 8?

anikethsaha:
> I am afraid `@media () => @media()` is not valid

I changed it 👍

evilebottnawi:
I mean we should test it manually https://github.com/cssnano/cssnano/pull/921#issuecomment-654775921 :smile:

anikethsaha:
I dont have ie in my machine. 


I think we can add a check using browserlist + caniuse-lite.  It would be a breaking change then.

evilebottnawi:
@anikethsaha I think it is time to learn virtual box :smile: I'm not sure if this does not work, maybe everything is fine, just need test

anikethsaha:
I have some bad experience with Virtual box earlier, I do have it installed and my config for that is messed up. 

I will try to look for some online VM to test them.

Scutua (approved):

trinitronx:
> @support () , @support () => @support(),@support() -> I think it is invalid syntax, no examples in spec, but maybe I wron

At least for `@media` queries, the spaces appear to be necessary for Chrome and Chromium-based browsers (e.g. Edge, etc...) to parse and apply them correctly.

For example, the current `cssnano` (`7.1.0`) transforms this:

```css
@media screen and (prefers-contrast:more) and (-ms-high-contrast:active) and (prefers-color-scheme: light),
       screen and (prefers-contrast:more) and (-ms-high-contrast:black-on-white) { 
/* some rules */
}
```

To this without spaces between `)` + `and`, and between `,` + `screen`:

```css
@media screen and (prefers-contrast:more)and (-ms-high-contrast:active)and (prefers-color-scheme: light),screen and (prefers-contrast:more)and (-ms-high-contrast:black-on-white), { 
/* some rules */
}
```

When using Chrome's "<kbd>Rendering</kbd>" -> "<kbd>Emulate CSS media feature `prefers-contrast`</kbd>", the media queries don't match after `cssnano` processing.

trinitronx:
> @trinitronx Sounds like a bug in chrome...

🤔 Hmm... It's not clear whether it is a bug or not...

In the W3C recommendation spec for "`Media Queries Level 3`", there is "`Example 21`" under "**_Malformed media query_**" section:

> EXAMPLE 21
> 
> The following is an malformed media query because having no space between ‘`and`’ and the expression is not allowed. (That is reserved for the functional notation syntax.)
> 
>     @media all and(color) { … }
> 
> Reference: [W3C Recommendation: Media Queries Level 3 - § 3.1. Error Handling
][1]

It makes no mention of the preceding space before an "`and`" whether it's required or not.  So that's inconclusive.

Then, in the W3C spec for "`Media Queries Level 4`" they have a more formal Backus-Naur [definition of the media query syntax][2], which does mention how a comma-separated `<media-query-list>` is parsed, as well as defining the `<media-and> = and <media-in-parens>` operator.  There is also a similar section about the space after the `and` and before an opening paren `(`:

> [!NOTE]
> Whitespace is required between a [`not`](https://www.w3.org/TR/mediaqueries-4/#valdef-media-not), `and`, or `or` keyword and the following `(` character, because without it that would instead parse as a [<function-token>](https://www.w3.org/TR/css-syntax-3/#typedef-function-token). This is not made explicitly invalid because it’s already covered by the above grammar. It’s fine to have whitespace between a `)` and a following keyword, however.
> 
> Reference: [W3C Recommendation: Media Queries Level 4 - § 3. Syntax][2]

That still makes no mention of omitting the whitespace between a closing paren `)` and a boolean operator `and`, `or`, or `not`.  Yet, since the BNF definition for `<media-and>` has spaces on both sides of the `and` ("`<media-and> = and <media-in-parens>`")... perhaps the Chromium developers interpreted this as being required, or added it into the parsing in such a way as to effectively make it required?

Just speculation:  It might be parsing it as a `not all` or `false`, or `unknown`... It's unclear which, because there is no useful output in the developer console saying what is going on.  It could also be discarding the media query list portion up to the next comma `,` given the following:

> A media query that does not match the grammar in the previous section must be replaced by `not all` during parsing.
> 

> [!NOTE]
> > Note: Note that a grammar mismatch does not wipe out an entire [media query list](https://www.w3.org/TR/mediaqueries-4/#media-query-list), just the problematic [media query](https://www.w3.org/TR/mediaqueries-4/#media-query). The parsing behavior defined above automatically recovers at the next top-level comma.
> 
> Reference: [W3C Recommendation: Media Queries Level 4 - § 3.2 Error Handling][3]

Since the broken example has `),screen and ` ... it could be that the `,screen` portion is breaking it because a typical `@media screen` requires the space in front of `screen` `<media-type>` (as in: `<media-type> = `[`<ident>`](https://www.w3.org/TR/css-values-3/#typedef-ident))

If it's a Chromium bug, then it's broken out in the wild and places limitations on what minification transforms are "_**safe**_" effectively due to the bug.  If it's not a bug, then it's not "_**safe**_" to do the transform in any case.

Whatever the case, current Chrome versions (tested on `137.0.7151.119` and above) seem to choke on the minified `@media` queries produced by `cssnano`'s current '`default`' preset.  So, in the practical sense, those particular transforms shouldn't be considered "_**safe**_".

[1]: https://www.w3.org/TR/mediaqueries-3/#c2
[2]: https://www.w3.org/TR/mediaqueries-4/#mq-syntax
[3]: https://www.w3.org/TR/mediaqueries-4/#error-handling

trinitronx:
Ok, so I have found something that works while digging more into the `,screen` spacing issue for Chrome.

It turns out that it works if I prefix the nested source CSS `@media` query rules with `only ` (e.g. `only screen and (...)` ).

For example, these nested `@media` queries:

```css
@media (prefers-contrast: more) {
	@media (forced-colors: none) and (prefers-color-scheme: dark),
		   only screen and (-ms-high-contrast: active) and (prefers-color-scheme: dark),
		   only screen and (-ms-high-contrast: white-on-black)  {
/* ... some rules ... */
	}
}
```

Transforms into:

```css
/* 
   Note that the PostCSS plugin prefers-color-scheme-query, is run here before CSSNano.
   The corresponding JS polyfill has been added to the page also.
   The previous example had it turned off and it still did not parse correctly in Google Chrome.
*/

@media (prefers-contrast:more) and (forced-colors:none) and (color: 48842621),only screen and (prefers-contrast:more) and (-ms-high-contrast:active) and (color:48842621),only screen and (prefers-contrast:more) and (-ms-high-contrast:white-on-black) {
/* ... some rules ... */
}

@media (prefers-contrast:more) and (forced-colors:none) and (prefers-color-scheme: dark),only screen and (prefers-contrast:more) and (-ms-high-contrast:active) and (prefers-color-scheme:dark),only screen and (prefers-contrast:more) and (-ms-high-contrast:white-on-black) {
/* ... same rules as above block duplicated ... */
/* ... for browsers that support prefers-color-scheme ... */
}
```

Note that now we have `(...),only screen and (...)` instead of just `(...),screen and (...)`.  That seems to parse OK in Chrome, and the styles are applied when `prefers-contrast` emulation is set to "`more`" and `prefers-color-scheme` is set to "`dark`".

Therefore, it seems that before we had a syntax construct `,screen and (-ms-high-contrast:active)` where the Microsoft Edge-specific vendor media query would probably evaluate to an `unknown` value, or perhaps as `,screen and not all,`.  Yet when prefixed with `,screen and ` it seemed to break the entire query, even for previous `,`-separated queries that should have been processed as if a logical OR were used.

Yet, now with the prefixed `,only screen and `, it probably evaluates to a `,true and (unknown)`.  The W3C "`Media Queries Level 4`" spec says this:

> If the result of any of the above productions is used in any context that expects a two-valued boolean, “`unknown`” must be converted to “`false`”.
> 

> [!NOTE]
> This means that, for example, when a [media query](https://www.w3.org/TR/mediaqueries-4/#media-query) is used in a [@media](https://www.w3.org/TR/css3-conditional/#at-ruledef-media) rule, if it resolves to “`unknown`” it’s treated as “`false`” and fails to match.
> 
> Media Queries use a three-value logic where terms can be “`true`”, “`false`”, or “`unknown`”. Specifically, it uses the [Kleene 3-valued logic](https://en.wikipedia.org/wiki/Three-valued_logic#Kleene_and_Priest_logics). In this logic, “`unknown`” means “_either true or false, but we’re not sure which yet_”.
> In general, an `unknown` value showing up in a formula will cause the formula to be `unknown` as well, as substituting “`true`” for the `unknown` will give the formula a different result than substituting “`false`”. The only way to eliminate an `unknown` value is to use it in a formula that will give the same result whether the `unknown` is replaced with a `true` or `false` value. This occurs when you have “`false AND unknown`” (evaluates to `false` regardless) and “`true OR unknown`” (evaluates to `true` regardless).
> 
> This logic was adopted because [<general-enclosed>](https://www.w3.org/TR/mediaqueries-4/#typedef-general-enclosed) needs to be assigned a truth value. In standard boolean logic, the only reasonable value is “`false`”, but this means that `not unknown(function)` is `true`, which can be confusing and unwanted. Kleene’s 3-valued logic ensures that unknown things will prevent a [media query](https://www.w3.org/TR/mediaqueries-4/#media-query) from matching, unless their value is irrelevant to the final result.

So, if we were dealing with an `unknown` here... then the transform of `,screen and (unknown)` may have been parsed incorrectly or differently than `,only screen and (unknown)`.

It would seem as if the `only` prefix is not optional when omitting the space.

PR #924, opened 2020-07-20
Upgrade cosmiconfig

TrySound:
Ref https://github.com/davidtheclark/cosmiconfig/blob/master/CHANGELOG.md#600

The new version slightly changed api, dropped old node support and
tweaked dependencies.

### DIFF
--- a/packages/cssnano/package.json
+++ b/packages/cssnano/package.json
@@ -25,7 +25,7 @@
   ],
   "license": "MIT",
   "dependencies": {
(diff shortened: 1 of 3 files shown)


TrySound:
Btw why yarn workspaces are not used here? A lot of lockfiles are outdated.

anikethsaha:
> Btw why yarn workspaces are not used here? A lot of lockfiles are outdated.

I have not used yarn workspace as of now. I will try to have a look into it and to see how well it works with lerna

PR #925, opened 2020-07-24
Add test case that demonstrates a crash with ::slotted + attribute selector

papandreou:
I'm playing around with generating random CSS and stumbled upon a selector that crashes `postcss-minify-selectors`.

PR #926, opened 2020-07-25
fix(postcss-minify-selectors): added check for node.operator for namespace

anikethsaha:
fixes https://github.com/cssnano/cssnano/pull/925

Explanation details : https://github.com/cssnano/cssnano/pull/925#issuecomment-663862095
Ref: [this](https://drafts.csswg.org/selectors-4/?expand=1#attrnmsp)


evilebottnawi:
Thanks!

PR #929, opened 2020-07-30
feat(postcss-reduce-initial): added ignore options

anikethsaha:
fixes #905 

Added new option `ignore` to skip the property passed in it.

### DIFF
--- a/packages/postcss-reduce-initial/src/index.js
+++ b/packages/postcss-reduce-initial/src/index.js
@@ -6,6 +6,8 @@ import fromInitial from '../data/fromInitial.json';
 import toInitial from '../data/toInitial.json';
 
 const initial = 'initial';
+
+// In most of the browser including chrome the initial for `writing-mode` is not `horizontal-tb`. Ref https://github.com/cssnano/cssnano/pull/905
 const defaultIgnoreProps = ['writing-mode'];
 
 export default plugin('postcss-reduce-initial', () => {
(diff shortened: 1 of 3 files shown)


packages/postcss-reduce-initial/src/index.js
```diff
@@ -20,6 +20,9 @@ export default plugin('postcss-reduce-initial', () => {
 
     css.walkDecls((decl) => {
       const lowerCasedProp = decl.prop.toLowerCase();
+      if (resultOpts.ignore && !!~resultOpts.ignore.indexOf(lowerCasedProp)) {
```
evilebottnawi, line 23: Let's put `writing-mode` in default config because it is broken in the most popular browser - chrome

anikethsaha: yeah, but what if they want to transform that ?

evilebottnawi: Ref: https://bugs.chromium.org/p/chromium/issues/detail?id=507397

I think we should always ignore this now, ideally we should use `browserlist` for this, but most of chromium based browsers are affected, we can't handle all of them using `browserlist` :disappointed:

packages/postcss-reduce-initial/src/index.js
```diff
@@ -6,6 +6,7 @@ import fromInitial from '../data/fromInitial.json';
 import toInitial from '../data/toInitial.json';
 
 const initial = 'initial';
+const defaultIgnoreProps = ['writing-mode'];
```
evilebottnawi, line 9: Let's add comment to the issue

PR #940, opened 2020-08-15
new: added support for running other plugins

anikethsaha:
added support for `plugins` in cssnano's config file

Checklist

- [x] tests
- [x] docs


packages/cssnano/package.json
```diff
@@ -1,53 +1,56 @@
...
+    "bugs": {
+        "url": "https://github.com/cssnano/cssnano/issues"
+    },
+    "engines": {
+        "node": ">=10.13.0"
+    },
+    "devDependencies": {
+        "autoprefixer": "^9.8.6"
+    }
+}
```
evilebottnawi, line 56: Why it was changed?

Issue #909, opened 2020-05-18
[Bug] : keyframe names with e at beginning are trying to convert


SkReD:
**Describe the bug**
Keyframes names like those `h1{ -webkit-animation: e836684w2 } h2{ animation: e836684w2 }` with `e` at beginnig are trying to convert by `postcss-convert-values`

**To Reproduce**
Here is PR with failing test on this case
https://github.com/cssnano/cssnano/pull/908

**Expected behavior**
Expected that no conversion performed

**Screenshots**

**Desktop (please complete the following information):**
- CSSNANO Version - `4.1.10`
- preset - `default` 
- Add the output of `npx envinfo && npm ls cssnano` here
```
  System:
    OS: Windows 10 10.0.18363
    CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
    Memory: 4.04 GB / 15.86 GB
  Binaries:
    Node: 12.14.1 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.19.1 - D:\Soft\Yarn\bin\yarn.CMD
    npm: 6.10.2 - C:\Program Files\nodejs\npm.CMD
  Managers:
    Cargo: 1.31.0 - ~\.cargo\bin\cargo.EXE
    pip2: 19.2.3 - d:\soft\python27\Scripts\pip2.EXE
  Utilities:
    Git: 2.24.0.
  Virtualization:
    Docker: 19.03.8 - C:\Program Files\Docker\Docker\resources\bin\docker.EXE
  Languages:
    Bash: 4.4.19 - C:\Windows\system32\bash.EXE
    Go: 1.11.2 - D:\Soft\Go\bin\go.EXE
    Python: 2.7.17
  Databases:
    MongoDB: 4.0.1 - D:\Soft\MongoDB\Server\4.0\bin\mongo.EXE
  Browsers:
    Edge: 44.18362.449.0
    Internet Explorer: 11.0.18362.1
  Monorepos:
    Yarn Workspaces: 1.19.1

fb-mobile@1.281.0 C:\CSSSR\mvno-forms
`-- @tinkoff-boxy/mmf@1.51.1
  `-- @tinkoff-boxy/mmb@3.37.0
    `-- optimize-css-assets-webpack-plugin@5.0.3
      `-- cssnano@4.1.10
```

anikethsaha:
[...]

PR #941, opened 2020-08-24
Bump postcss-values-replace

SkReD:
Closes https://github.com/cssnano/cssnano/issues/909


evilebottnawi (approved):
Thanks

/cc @anikethsaha

evilebottnawi (approved):

packages/postcss-convert-values/src/index.js
```diff
@@ -24,7 +24,12 @@ function parseWord(node, opts, keepZeroUnit) {
   const pair = unit(node.value);
   if (pair) {
     const num = Number(pair.number);
-    const u = pair.unit;
+    let u = pair.unit;
+
+    if (u[0] === '.') {
+      u = u.slice(1);
+    }
+
```
anikethsaha, line 32: Why removing the leading `.` ?

anikethsaha (changes requested):
One question. 
And also fix the linting issue

packages/postcss-convert-values/src/index.js
SkReD: my thought  is that is not the part of `unit` but the `value`

anikethsaha: for input like https://github.com/TrySound/postcss-value-parser/blob/master/test/unit.js#L10, there would be still number (length) in the unit. 

I wonder whether this should be fixed in `postcss-value-parser` as `.3rem` or `3rem` is even a value unit or not. 

cc @evilebottnawi

anikethsaha (changes requested):

packages/postcss-convert-values/src/index.js
anikethsaha: @evilebottnawi

anikethsaha (changes requested):
Could you please add a test case mentioned here https://github.com/cssnano/cssnano/issues/958

packages/postcss-convert-values/src/index.js
evilebottnawi: Yes, it is not good idea

anikethsaha: @SkReD can you undo this change and we can submit a issue in the parser repo as this should be handled by the parser not by the minifier

ludofischer:
`css-value-parser` has been updated on master already but I think the test cases in this issue are interesting. Should check what happens with the current rc.

ludofischer:
Fixed in 5.0.0. I've added the commit with the test case to https://github.com/cssnano/cssnano/pull/1059

PR #956, opened 2020-09-24
fix(postcss-merge-longhand): allow mixing custom properties with regular values when merging

Maistho:
This resolves an issue where merging properties would break the css

Example of the issue
```css
/* This declaration should not be changed */
h1{border:var(--v1) solid var(--v2, #abc123);border-right-color:blue}

/* Before this commit it would be, into this */
h1{border-right-color:var(--v2,#abc123);border-right:var(--v1) solid #00f;border-style:solid;border-width:var(--v1)}
```

### DIFF
--- a/packages/postcss-merge-longhand/src/__tests__/borders.js
+++ b/packages/postcss-merge-longhand/src/__tests__/borders.js
@@ -1229,3 +1229,11 @@ test(
 );
 
 test('should handle empty border', processCSS('h1{border:;}', 'h1{border:;}'));
+
+test(
+  'should handle mixed custom property and regular declarations',
+  processCSS(
+    'h1{border:var(--v1) solid var(--v2, #abc123);border-right-color:blue}',
+    'h1{border:var(--v1) solid var(--v2, #abc123);border-right-color:blue}'
+  )
+);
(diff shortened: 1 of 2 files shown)


evilebottnawi (approved):
/cc @anikethsaha

anikethsaha:
how about keeping the merge check and remove the whole else block alone [here](https://github.com/cssnano/cssnano/pull/956/files#diff-4c8907312ced1c26f0e265ce04a00818L274) ?

cause this would merge non custom props val with props val

Maistho:
> cause this would merge non custom props val with custom props val

When would this cause an issue? I would like to add a test case for this if it does cause any issues, since the current tests all run OK with my changes.

I admit to being a bit sceptical when removing the check myself, but since it didn't seem to cause any issues and it completely solved my issue it seemed fine.

anikethsaha:
Maybe, try using CSS where there is a customs value and non-custom value

It should not merge them.

ludofischer:
https://github.com/cssnano/cssnano/pull/1057 solved this. I've added the test case to https://github.com/cssnano/cssnano/pull/1059

PR #959, opened 2020-10-19
Eslint 7

ludofischer:
`yarn lint` and `yarn fixlint`run without errors on the current code. When I break the indentation in `packages/postcss-colormin/src/colours.js`, and `yarn lint` reports an error and `yarn fixlint` manages to fix it automatically.
 
I did not notice any breaking changes that concern the current `.eslintrc.js` ESLint config  in the [ESLint v7 migration guide](https://eslint.org/docs/user-guide/migrating-to-7.0.0) and [ESLint v6 migration guide](https://eslint.org/docs/user-guide/migrating-to-6.0.0).


evilebottnawi (approved):

evilebottnawi:
Thanks

PR #954, opened 2020-09-20
feat: postcss 8 and new modern postcss API

anikethsaha:
Adding support for new postcss 8 and its new API

#### Status

- [x] dep updates
- [x] test helpers
- [x] site playground
- [x] cssnano core
- [x] postcss-colormin
- [x] postcss-convert-values
- [x] postcss-discard-comments
- [x] postcss-discard-empty
- [x] postcss-discard-duplicates
- [x] postcss-discard-overridden (the test runner needs to be fixed)
- [x] postcss-discard-unused
- [x] postcss-merge-idents
- [x] postcss-merge-longhand
- [x] postcss-merge-rules
  - [x] plugin refactor
  - [x] tests case for `should not crash when node.raws.value is null` not working properly as postcss runner is not able to run the plugin - maybe the postcss is `< 8` there, not sure. error `'true' is not a postcss plugin` when using the `postcss-simple-vars` plugin
- [x] postcss-minify-font-values
- [x] postcss-minify-gradients
- [x] postcss-minify-params
- [x] postcss-minify-selectors
  - [x] the plugin code
  - [x] tests: the `toModules` plugin's visitor keys are not being visited. But the plugin is being called. Need to fix that.
  - [ ] (non-blocker): https://github.com/jonathantneal/postcss-font-magician/issues/91
- [ ] postcss-normalize-charset
- [ ] postcss-normalize-display-values
- [ ] postcss-normalize-positions
- [ ] postcss-normalize-repeat-style
- [ ] postcss-normalize-string
- [ ] postcss-normalize-timing-functions
- [ ] postcss-normalize-unicode
- [ ] postcss-normalize-url
- [ ] postcss-normalize-whitespace
- [ ] postcss-ordered-values
- [ ] postcss-reduced-idents
- [ ] postcss-reduced-initial
- [ ] postcss-reduced-transform
- [ ] postcss-svgo
- [ ] postcss-unique-selectors
- [ ] postcss-zindex
- [ ] stylehacks
- [ ] postcss-calc (external)
  - [ ] it is supported in upstream
  - [ ] need to update dependency in our end
- [ ] css-declaration-sorter (external)
  - [x] it is supported in upstream https://github.com/Siilwyn/css-declaration-sorter/issues/106
  - [ ] need to update dependency in our end
- [ ] autoprefixer (external) 
  - [x] it is supported in upstream
  - [ ] need to update dependency in our end

PR #961, opened 2020-10-27
Update more plugins to PostCSS 8

ludofischer:
- [x] postcss-normalize-charset
- [x] postcss-normalize-display-values
- [x] postcss-normalize-positions
- [x] postcss-normalize-repeat-style
- [x] postcss-normalize-string
- [x] postcss-normalize-timing-functions
- [x] postcss-normalize-unicode
- [x] postcss-normalize-url
- [x] postcss-normalize-whitespace
- [x] postcss-ordered-values
- [x] postcss-reduced-idents
- [x] postcss-reduced-initial
- [x] postcss-reduced-transform
- [x] postcss-svgo
- [x] postcss-unique-selectors
- [x] postcss-zindex
- [x] stylehacks
- [x] find out why colormin fails


cossssmin:
Can you please also include `postcss-merge-longhand`? 🙏

anikethsaha:
@cossssmin done here https://github.com/cssnano/cssnano/pull/954

ludofischer:
I think I did what we talked about in #954, so I only touched the plugins themselves and not the fixtures and the runners.
### Known issues
1.  when running cssnano, itself some  plugins seem to have no effect or the plugin order seems to count now when running cssnano. You can see this in the `cssnano-preset-lite` tests, `normalize-whitespace` seems to work only if you place it first in the plugin array (at least on my machine). Might be an issue with promises resolving too soon, so cssnano finishes before all plugins had a chance to run?

2.  tests fail for two plugins:
 * colormin: these tests run the full cssnano with the `standard` preset. The assertions rely on normalize-whitespace to strip empty space around numbers. Unfortunately this does not work any more, probably because of (1) above
 * merge-rules: tests use postcss 7, which is dragged in `node_modules` by its `cssnano-utils` dependency (according to `yarn why`). Although I thought lerna was supposed to link it to the repo version of `cssnano-utils`, so maybe I am missing something?

Don't hesitate to ask any questions, this has turned into a large PR and some superfluous changes might have slipped through.

evilebottnawi:
> when running cssnano, itself some plugins seem to have no effect or the plugin order seems to count now when running cssnano. You can see this in the cssnano-preset-lite tests, normalize-whitespace seems to work only if you place it first in the plugin array (at least on my machine). Might be an issue with promises resolving too soon, so cssnano finishes before all plugins had a chance to run?

Can you provide simple example? Sorry for stupid questions, I'm a little focused on other projects, so it's hard for me to quickly get into the problem and advise solutions

ludofischer:
For example, in [`packages/cssnano-preset-lite/src/__tests__/index.js`](https://github.com/cssnano/cssnano/blob/91604d3ccde260047187dd9cecaaf17ad08f6208/packages/cssnano-preset-lite/src/__tests__/index.js), `Simple test with comment #2`
passes only if I move ` [postcssNormalizeWhitespace, options.normalizeWhitespace]` to the top in the `plugins` array in `packages/cssnano-preset-lite/src/index.js`, but then `Simple test` and `Simple test with comment #3` fail as the comments and empty elements are not removed.

evilebottnawi:
This due new visitor API? Some plugins can depend on each other, I think in this case we should use `Root` and `RootExit` with old API, otherwise it will now work

evilebottnawi:
> merge-rules: tests use postcss 7, which is dragged in node_modules by its cssnano-utils dependency (according to yarn why). Although I thought lerna was supposed to link it to the repo version of cssnano-utils, so maybe I am missing something?

Can you look at dependencies, maybe we forget something update?

ludofischer:
> This due new visitor API? Some plugins can depend on each other, I think in this case we should use `Root` and `RootExit` with old API, otherwise it will now work

I suspect the way `cssnano` currently invokes the plugins somehow makes the situation worse.  There are a lot less errors  if I run the plugins through postcss directly. In [packages/cssnano-preset-lite/src/__tests__/index.js](https://github.com/cssnano/cssnano/blob/91604d3ccde260047187dd9cecaaf17ad08f6208/packages/cssnano-preset-lite/src/__tests__/index.js), if I replace

```javascript
const { processCSS } = processCSSWithPresetFactory(preset);
```
with

```javascript
const plugins = [
  postcssDiscardComments,
  postcssNormalizeWhitespace,
  postcssDiscardEmpty,
  rawCache,
];

const { processCSS } = processCSSFactory(plugins);
```
I only have one test failure in `Simple test with comment #3`, because the initial empty space does not get removed, which I found is an actual issue with `postcss-normalize-whitespace` (it happens also when testing that plugin in isolation).

> Can you look at dependencies, maybe we forget something update?

That one should was a false alarm, something must have broken in my project directory. The `merge-rules` tests pass on CI, and deleting every `node_modules` directory and resinstalling made it pass on my machine as well :)

evilebottnawi:
> I only have one test failure in Simple test with comment #3, because the initial empty space does not get removed, which I found is an actual issue with postcss-normalize-whitespace (it happens also when testing that plugin in isolation).

Weird, do you have ideas what is wrong? I still think we use invalid event here

ludofischer:
> Weird, do you have ideas what is wrong? I still think we use invalid event here

This [commit](https://github.com/cssnano/cssnano/pull/961/commits/bcd3e795007b68151a318067df6c6cca3e63e4e2) fixes the failure with cssnano-preset-lite.
I changed cssnano to invoke postcss with all the plugins in an array at once, instead of calling postcss once per plugin. The problem is that the commit to invoke postcss once per had been just made on the `postcss-8` branch, so I wonder if I am missing something here.
Maybe @anikethsaha can explain if it was necessary to call postcss with a single plugin at the time?

anikethsaha:
> I changed cssnano to invoke postcss with all the plugins in an array at once, instead of calling postcss once per plugin. The problem is that the commit to invoke postcss once per plugin had been just made on the `postcss-8` branch, so I wonder if I am missing something here.

is it in the cssnano core ? can you show the line which is calling postcss per plugin  (forgot the implementation, sorry ). I dont remember explicitly doing this if it was not mentioned in the migration guide. (maybe I am missing something here.)

ludofischer:
> is it in the cssnano core ? can you show the line which is calling postcss per plugin (forgot the implementation, sorry ). I dont remember explicitly doing this if it was not mentioned in the migration guide. (maybe I am missing something here.)

Yer it's cssnano core:  https://github.com/cssnano/cssnano/blob/5f9749449da13b6d9e55f91301248a66cf607cca/packages/cssnano/src/index.js#L121
`postcss()` gets called inside the `forEach()` loop, so every time it gets called with a single plugin.

evilebottnawi:
@ludofischer can we add https://github.com/postcss/postcss/blob/master/package.json#L6 tro our packages in this PR?

package.json
```diff
@@ -18,7 +18,7 @@
     "publish": "lerna publish"
   },
   "engines": {
-    "node": ">=6.9.0"
+    "node": "^10 || ^12 || >=14"
```
evilebottnawi, line 21: Let's do it for each plugin

evilebottnawi (commented):
Note

ludofischer:
I fear switching the visitor API exposes too many infinite loops, as plugins undo and redo each other work. It might be better  to create a PostCSS 8  compatibility release based on current `master` where we switch everything to `Once()` so people can get rid of multiple PostCSS versions in their build without the risk of their build freezing because of an infinite loop.

evilebottnawi:
> I fear switching the visitor API exposes too many infinite loops, as plugins undo and redo each other work. It might be better to create a PostCSS 8 compatibility release based on current master where we switch everything to Once() so people can get rid of multiple PostCSS versions in their build without the risk of their build freezing because of an infinite loop.

We do the same for CSS modules, because new API is incompatibility when you use many plugins

/cc @ai Don't have ideas how to better solve it, but I you think you need to put attention on this

ludofischer:
> We do the same for CSS modules, because new API is incompatibility when you use many plugins

For cssnano, I believe you would need to refactor the transformation logic in some plugins (e.g. merge-longhand and colormin) to first detect whether the node requires transforming. The hardest plugins to port use an (implicit) model of the structure of CSS _property values_, so to solve it at the PostCSS level you probably need a more detailed AST. I think this has already been discussed (as far back when there were discussions of merging PostCSS and CSSTree).

ai:
> We do the same for CSS modules, because new API is incompatibility when you use many plugins

There is always “give up” option to use `Once`. I did it for Autoprefixer.

ludofischer:
> There is always “give up” option to use `Once`. I did it for Autoprefixer.

Case in point, I've managed to trigger another inifinite loop https://github.com/postcss/postcss-calc/pull/125 I think when a plugin uses `postcss-value-parser`, it breaks the single-tree traversal model, as the 'value' AST does not persist across passes, so it's hard to do an efficient change comparison. But making a good CSS value AST looks daunting, some plugins contain [hundreds of lines of code](https://github.com/postcss/postcss-calc/blob/master/src/lib/reducer.js) that analyze a value.

evilebottnawi:
@ludofischer it's sad that in so many years, PostCSS still not have ast for values/at-rules...

What is our future plan?

evilebottnawi:
@ai This is very sad, and not my first project where I update PostCSS and the new visitor API just impossible to use, I think you should think about it, otherwise it loses its meaning and essence, I can spend time and think about possible solutions, but unfortunately I have very little free time

ai:
@evilebottnawi do you have only problem with infinite loops?

evilebottnawi:
@ai I would not say that the problem with infinite loops. They're pretty logical. For example - In both plugins we listen changes and insert new nodes on this changes, this can be solved using `Symbol('walked')`, bit it is the extremely bad solution, we can’t tell exactly when we should not touch node again, plugins are independent from each other. The only really correct solution - value and at-rule built-in parser. But I understand perfectly well that this cannot be done quickly, but without that new visitor API is not suitable for most of cases (to be honestly, only in `css-loader` I was able to take advantage of it, and that's because I always know `css-loader` plugins is the latest).

ai:
> without that new visitor API is not suitable for most of cases

Yeap, visitor API was created for value parser in-mind. For instance, CSS function processing will be much better.

anikethsaha:
I think we should continue the migration as of now and after release we can update when postcss is ready with the changes. 
What do you guys think?

alexander-akait:
Yep, let's use `Root` right now, and in future we will improve some plugins there it is possible

ai:
I recommend `OnceExit`. It is better for minifier to run at the end.

anikethsaha:
So we will be moving the logic in single event i.e `OnceExit`? or with `*Exit` of each event type ?

anikethsaha:
> > is it in the cssnano core ? can you show the line which is calling postcss per plugin (forgot the implementation, sorry ). I dont remember explicitly doing this if it was not mentioned in the migration guide. (maybe I am missing something here.)
> 
> Yer it's cssnano core:
> 
> https://github.com/cssnano/cssnano/blob/5f9749449da13b6d9e55f91301248a66cf607cca/packages/cssnano/src/index.js#L121
> 
> 
> `postcss()` gets called inside the `forEach()` loop, so every time it gets called with a single plugin.

@ludofischer Thanks a lot for pointing this.

Yea I think we can reduce the postcss calls here to one with all those resolved plugin to an array. 
I think we have more or less the same for current version.

Feel free to make that change (if not done already) and check the result if its fixing the issue. Other it would take some time for me to change and check this.

Let me know.

ludofischer:
> Feel free to make that change (if not done already) and check the result if its fixing the issue. Other it would take some time for me to change and check this.

I have already changed that, but if we use `OnceExit` everywhere, I think it's simpler to start over, as the transformation logic is all going back into a `walk()` method like before.
Also maybe calling `postcss().process()` inside a plugin won't work well with PostCSS 8, as sometimes other plugins don't have any effect if you do that. You can see the issue affecting postcss-modules (see https://github.com/css-modules/postcss-modules/pull/114 ).  I suspect `process() creates a new AST traversal, and other plugins that register as visitors on the top AST traversal don't get called properly (see simpler reproduction in this repo https://github.com/ludofischer/postcss-strange-cases).
 In [this branch](https://github.com/ludofischer/cssnano/blob/postcss-8-once/packages/cssnano/src/index.js) I have tried to expose cssnano as a processor (see the PostCSS API https://postcss.org/api/#acceptedplugin), maybe it's more compatible, assuming we can preserve the plugin loading behaviour.

ludofischer:
Happy holidays! I have created a pull request #975 where I switched plugins to `OnceExit()` (I have used a codemod in part), as discussed. I hope it will be possible to merge it, I know, not a very exciting holiday present, but at least it paves the way for future PostCSS upgrades.

PR #985, opened 2021-01-15
Bundle cssnano using Parcel

aminya:
Cssnano takes about [3.5 seconds to load](https://github.com/parcel-bundler/parcel/issues/5072#issuecomment-761182466). This PR tries to diagnose the issue

This PR reduces the build time of Parcel by 2 seconds!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ _webpackOutput
 # Cache
 
 .eslintcache
+.parcel-cache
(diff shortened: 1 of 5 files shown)


alexander-akait (commented):
No thanks, look at bundle, a lot of bugs

aminya:
For those who come across this PR. The performance issues caused by CSSNano are now resolved in parcel-css.
https://github.com/parcel-bundler/parcel-css

PR #1008, opened 2021-03-04
5.0.0 rc1

ludofischer:
I have struggled a bit with the changelog. Unfortunately lerna wants to bump every package, I think it is related to this issue: https://github.com/lerna/lerna/issues/2437

Closes: https://github.com/cssnano/cssnano/pull/1007


anikethsaha (approved):

anikethsaha:
> I have struggled a bit with the changelog. Unfortunately lerna wants to bump every package, I think it is related to this issue: lerna/lerna#2437

I guess its fine as it is a rc only.

PR #1009, opened 2021-03-05
docs: fix link to PostCSS docs

ludofischer:
Closes https://github.com/cssnano/cssnano/pull/672


ludofischer:
Fixed in https://github.com/cssnano/cssnano/pull/1022

PR #1011, opened 2021-03-07
chore: transpile only for supported Node.js versions

ludofischer:
Moving to PostCSS 8 drops Node.js < 10.13 support, so let's avoid bloating to the generated code by transpiling for older versions.
I've also updated Babel to use the new top-level targets option: https://babeljs.io/blog/2021/02/22/7.13.0

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -143,7 +143,7 @@ module.exports = function (api) {
         '@babel/preset-env',
         {
           targets: {
-            node: '6.9.0',
+            node: '10.13.0',
           },
         },
       ],
--- a/babel.config.js
+++ b/babel.config.js
@@ -137,17 +137,8 @@ module.exports = function (api) {
   }
 
   return {
-    presets: [
-      [
-        // ES features necessary for user's Node version
-        '@babel/preset-env',
-        {
-          targets: {
-            node: '10.13.0',
-          },
-        },
-      ],
-    ],
+    targets: { node: '10.13.0' },
+    presets: ['@babel/preset-env'],
     plugins,
   };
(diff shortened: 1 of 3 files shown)


ludofischer:
I discoveres this by trying to understand why there's a problem requiring the presets in the website playground, see the comments here: https://github.com/cssnano/cssnano/blob/80dd2e3f2bdc0e6c900f18b9714ef79bdcfba470/site/src/components/editor/postcss_runner.js#L7

alexander-akait (approved):

anikethsaha (approved):

anikethsaha:
> I discoveres this by trying to understand why there's a problem requiring the presets in the website playground, see the comments here:
> 
> https://github.com/cssnano/cssnano/blob/80dd2e3f2bdc0e6c900f18b9714ef79bdcfba470/site/src/components/editor/postcss_runner.js#L7

yea it was intentional, I was trying to get the lazy loading thing working but there were some issues I couldnt debug.

ludofischer:
> yea it was intentional, I was trying to get the lazy loading thing working but there were some issues I couldnt debug.

Yes, I remember somehow the Docusaurus webpack config had issues both `require()`-ing and `importing` the published packages. It wonder if it is a problem with their webpack config or whether cssnano itself is hard to bundle. Importing from the source directories  would require replacing the `'lerna` prefix on imports.

ludofischer:
I think this pull request introduced unnecessary duplicate Babel dependencies. See https://github.com/cssnano/cssnano/pull/1014 for a fix.

PR #1016, opened 2021-03-14
fix: update SVGO

ludofischer:
* Update SVGO dependency
* Refactor plugin to use new SVGO API
* Refactor test to use new SVGO API
* Update test since removing xml:space="preserve" was an
  SVGO bug
* Update documentation since SVGO is now synchronous


packages/postcss-svgo/src/__tests__/index.js
```diff
@@ -23,39 +24,39 @@ test(
...
+    'h1{background:url(\'data:image/svg+xml;charset=utf-8,<svg xmlns="http://www.w3.org/2000/svg" xml:space="preserve"><circle cx="50" cy="50" r="40" fill="%23ff0"/></svg>\')}'
   )
 );
 
 test(
   'should optimise inline svg (uppercase property)',
   processCSS(
     'h1{background:URL(\'data:image/svg+xml;utf-8,<?xml version="1.0" encoding="utf-8"?><!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd"><svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" xml:space="preserve"><circle cx="50" cy="50" r="40" fill="yellow" /><!--test comment--></svg>\')}',
-    'h1{background:URL(\'data:image/svg+xml;charset=utf-8,<svg xmlns="http://www.w3.org/2000/svg"><circle cx="50" cy="50" r="40" fill="%23ff0"/></svg>\')}'
+    'h1{background:URL(\'data:image/svg+xml;charset=utf-8,<svg xmlns="http://www.w3.org/2000/svg" xml:space="preserve"><circle cx="50" cy="50" r="40" fill="%23ff0"/></svg>\')}'
```
ludofischer, line 35: According to the SVGO commit log, removing `xml:space="preserve"` was a bug https://github.com/svg/svgo/commit/776ec1e71bcd2aa768b2d3ac987177f706834547

packages/postcss-svgo/src/__tests__/index.js
```diff
@@ -1,6 +1,7 @@
 import { readFileSync as file } from 'fs';
 import postcss from 'postcss';
 import filters from 'pleeease-filters';
+import svgo from 'svgo';
```
TrySound, line 4: Could you use named exports please?

packages/postcss-svgo/src/index.js
```diff
@@ -53,71 +52,50 @@ function minifyPromise(decl, getSvgo, opts) {
...
-            after: '',
-          });
-        })
-        .catch((error) => {
-          throw new Error(`${PLUGIN}: ${error}`);
-        })
-    );
+    const result = optimize(svg, opts);
+    if (result.error) {
+      throw new Error(`${PLUGIN}: ${result.error}`);
```
TrySound, line 57: There are possible exception from optimize call which probably need to be wrapped too. This is something I missed why developing v2. v3 will probably only throw.

ludofischer: Do you mean it's best to call `optimize()` inside a `try/catch` ?

TrySound: I mean both handle result.error and with try/catch or like this
```
let reuslt
try {
  result = optmize(svg, opts)
  if (result.error) {
    throw Error(result.error)
  }
} catch (error) {
  throw Error(`${PLUGIN}: ${error.message}`)
}
```

alexander-akait (approved):
Good job, thanks, let's do new rc release after this

Issue #910, opened 2020-05-27
[Bug] : Incorrect calc(...) value crashes cssnano


strarsis:
**Describe the bug**
Using some (specific) incorrect calc(...) values crashes cssnano.

**To Reproduce**
Use as input:
```css
.test {
  min-height: calc(100vh20 / 80 * 20rem - 1rem);
}
````
You can try it in the [cssnano playground](https://cssnano.co/playground/).

On the cssnano playground:
````
Uncaught (in promise) 
JisonLexerError
hash: {errStr: "Lexical error on line 1: Unrecognized text.↵↵  Erroneous area:↵1: 100vh20 / 80 * 20rem - 1rem↵^..^", recoverable: false, text: "", token: null, line: 0, …}
postcssNode: Declaration {raws: {…}, type: "decl", parent: Rule, source: {…}, prop: "min-height", …}
message: "Lexical error on line 1: Unrecognized text.↵↵  Erroneous area:↵1: 100vh20 / 80 * 20rem - 1rem↵^..^"
name: "JisonLexerError"
stack: "JisonLexerError: Lexical error on line 1: Unrecognized text.↵↵  Erroneous area:↵1: 100vh20 / 80 * 20rem - 1rem↵^..^↵    at <input css 86>:2:3↵    at Object.parseError (https://cssnano.co/phenomic/phenomic.main.22573732.js:131424:15)↵    at Object.lexer_parseError [as parseError] (https://cssnano.co/phenomic/phenomic.main.22573732.js:132557:44)↵    at Object.lexer_next [as next] (https://cssnano.co/phenomic/phenomic.main.22573732.js:133552:22)↵    at Object.lexer_fastLex [as fastLex] (https://cssnano.co/phenomic/phenomic.main.22573732.js:133627:18)↵    at fastLex (https://cssnano.co/phenomic/phenomic.main.22573732.js:131827:27)↵    at Parser.parse (https://cssnano.co/phenomic/phenomic.main.22573732.js:131901:30)↵    at https://cssnano.co/phenomic/phenomic.main.22573732.js:127716:30↵    at walk (https://cssnano.co/phenomic/phenomic.main.22573732.js:76903:7)↵    at ValueParser.walk (https://cssnano.co/phenomic/phenomic.main.22573732.js:417:3)↵    at transformValue (https://cssnano.co/phenomic/phenomic.main.22573732.js:127710:50)"
__proto__: Error
````

On a node build environment:
````
[...]

PR #1034, opened 2021-03-31
refactor: remove is-svg

ludofischer:
Hopefully this will stop further alarmed comments.

### DIFF
--- a/packages/postcss-svgo/src/__tests__/index.js
+++ b/packages/postcss-svgo/src/__tests__/index.js
@@ -1,5 +1,4 @@
 import { readFileSync as file } from 'fs';
-import postcss from 'postcss';
 import filters from 'pleeease-filters';
 import { extendDefaultPlugins } from 'svgo';
 import plugin from '../';
@@ -191,6 +190,13 @@ test(
 );
 
 test(
+  'should skip svgs containing unclosed tags',
+  passthroughCSS(
+    'h1{background:url(data:image/svg+xml;charset=utf-8,<svg>style type="text/css"><![CDATA[ svg { fill: red; } ]]></style></svg>)}'
+  )
+);
+
+test(
   'should pass through links to svg files',
   passthroughCSS('h1{background:url(unicorn.svg)}')
 );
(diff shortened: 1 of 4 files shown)


alexander-akait (dismissed):
Need fix CI

ludofischer:
Those look like legit test failures. I'l need to investigate what is happening.

ludofischer:
Cannot update past `is-svg` `4.2.2`, which luckily contains the CVE fix.  

Interestingly there's an actual regression in `is-svg` `4.3.x`, it fails to reject strings that contain random text after the svg.

See the `is-svg` test: https://github.com/sindresorhus/is-svg/blob/034967d4053dd6ca907a6bfeb047cd61b4c12f0a/test.js#L39
and the corresponding issue in the library that `is-svg` uses: https://github.com/NaturalIntelligence/fast-xml-parser/issues/327

alexander-akait:
We should wait fix in this case

ludofischer:
We could maybe skip the `isSvg` step completely.  Since `is-svg` was never 100% accurate, some non-SVGs things are caught at the `isSvg` step and some in the `svgo.optimize()` step. I wonder if we could just try `svgo.optimize()` and catch the error. That way we would also stop depending on two different XML parsers (the one that comes from `is-svg` and the one from `svgo`). The problem is that we would feed a lot of the non-svg things inside `url()` to SVGO, but in fact both `is-svg` and SVGO start by trying to parse the input as XML, they just offer a different interface for reporting errors.

alexander-akait:
Sounds good

packages/postcss-svgo/src/__tests__/index.js
```diff
@@ -203,19 +209,6 @@ test(
   )
 );
```
ludofischer, line 211: I have changed these tests because I 've realized there is no meaningful difference between these cases. They're all invalid SVGs. They were handled separately because some failed at the `isSvg` stage and some at the `svgo` stage.

packages/postcss-svgo/src/index.js
```diff
@@ -48,14 +47,13 @@ function minify(decl, opts) {
...
 
-    if (!isSvg(svg)) {
-      return;
-    }
-
     let result;
     try {
       result = optimize(svg, opts);
       if (result.error) {
+        if (result.error.startsWith('Error in parsing SVG')) {
```
ludofischer, line 54: I could not find a better way to detect the cause of the error. Maybe @TrySound has got a better suggestion? Anyway I wonder whether it makes sense to throw an error just because we cannot process sommething, instead of just leaving it alone. See this comment https://github.com/cssnano/cssnano/issues/910#issuecomment-811431880

TrySound:
I don't get why do we even need to check something? `data:image/svg+xml` implies we deal with svg. So we run svgo with it and if this svg is invalid throw error. Am I missing something?

packages/postcss-svgo/src/index.js
alexander-akait: I think we don't need throw the error in this cases (and all cases, for example `postcss-calc`, if we can't parse it just log warning, so developers can report about it and we still continue work), let's use `console.warn` and continue work

TrySound: Doesn't postcss have `warn` method?

alexander-akait: oh, yep, forgot about it

ludofischer:
> I don't get why do we even need to check something? `data:image/svg+xml` implies we deal with svg. So we run svgo with it and if this svg is invalid throw error. Am I missing something?

I 99% agree with you. Except in the context of cssnano I would not throw on invalid SVG , as it might be preferable to skip minification of that particular section rather than crashing  completely.  
After all, cssnano is not a linter.

alexander-akait:
So let's avoid using `throw new Error` and just use postcss `warn` method, in future we should avoid using `throw new Error` everything

alexander-akait (approved):

PR #1015, opened 2021-03-11
test: migrate ci task from azure-pipelines to github actions

armano2:
Migrate Ci runs from azure-devops to github actions

https://github.com/armano2/cssnano/actions/runs/644343038

ref #833

PR #1060, opened 2021-04-30
chore(css-size): update dependencies

ludofischer:
After updating the C++ compiler, I could not build the `brotli-size` package required by css-simple any more because. I updated `brotli-size` and a bunch of other dependencies that were stuck on very old versions, which gets rid of a boatload of transitive dependencies. The only change of note is that some dependencies now require Node.js 10,x, but it does not change anything for us since we required Node.js 10 already.

Here's a summary of the major dependency changes:
* `brotli-size`: require Node.js 10.x [changelog](https://github.com/erwinmombay/brotli-size/releases)
* `gzip-size`: require Node.js 10.x [changelog](https://github.com/sindresorhus/gzip-size/releases)
* `round-precision`: removed and inlined the code instead. This package has two dependencies of its own, but all it does  is throw if the precision is not an integer (no need, since our code only uses the literal values `2` and `4`) and or if the number to round is infinite (no need) (https://github.com/bendrucker/round-precision/blob/master/index.js)

### DIFF
--- a/packages/css-size/package.json
+++ b/packages/css-size/package.json
@@ -34,14 +34,13 @@
   },
   "repository": "cssnano/cssnano",
   "dependencies": {
-    "brotli-size": "^0.1.0",
+    "brotli-size": "^4.0.0",
     "cli-table3": "^0.5.1",
     "cssnano": "^5.0.2",
-    "gzip-size": "^5.1.1",
+    "gzip-size": "^6.0.0",
     "minimist": "^1.2.0",
-    "pretty-bytes": "^5.2.0",
-    "read-file-stdin": "^0.2.1",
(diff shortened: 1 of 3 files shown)


alexander-akait (approved):

alexander-akait:
@ludofischer maybe we can migrate on github actions?

ludofischer:
> @ludofischer maybe we can migrate on github actions?

Someone already made a pull request for that but it uses outdated `actions/cache` and `actions/setup-node`: https://github.com/cssnano/cssnano/pull/1015/files. Can you take a look at it and say if you see anything else wrong with it? Then we could build on top of that.

PR #1061, opened 2021-04-30
chore: remove babel-plugin-module-resolver

ludofischer:
I am pretty sure the `lerna:` prefix in the imports is not needed, at least with recent Lerna versions (it should not be needed even if we switch to npm 7 workspaces). Removing lets us get rid of one extra Babel plugin and simplify the Babel config.

I've also converted the config to JSON as according to the Babel docs that makes it more cacheable. I could not detect a significant speedup though.


alexander-akait (approved):

alexander-akait:
Something strange with coverage :confused:

PR #1066, opened 2021-05-06
Fix destructuring when undefined

marcusds:
Keep getting the following error, not sure why, can't find anything in my CSS that causes it. Seems build properly with this change below.

```
Module build failed (from ./node_modules/postcss-loader/dist/cjs.js):
TypeError: Cannot destructure property 'name' of 'ruleA.parent' as it is undefined.
    at canMerge (.\node_modules\postcss-merge-rules\dist\index.js:119:5)
    at .\node_modules\postcss-merge-rules\dist\index.js:375:20
    at .\node_modules\postcss\lib\container.js:115:18
    at .\node_modules\postcss\lib\container.js:74:18
    at Root.each (.\node_modules\postcss\lib\container.js:60:16)
    at Root.walk (.\node_modules\postcss\lib\container.js:71:17)
    at Root.walkRules (.\node_modules\postcss\lib\container.js:113:19)
    at OnceExit (.\node_modules\postcss-merge-rules\dist\index.js:433:15)
    at LazyResult.runAsync (.\node_modules\postcss\lib\lazy-result.js:388:19)
    at async Object.loader (.\node_modules\postcss-loader\dist\index.js:87:14)
```

### DIFF
--- a/packages/postcss-merge-rules/src/index.js
+++ b/packages/postcss-merge-rules/src/index.js
@@ -100,7 +100,7 @@ function canMerge(ruleA, ruleB, browsers, compatibilityCache) {
   }
 
   const parent = sameParent(ruleA, ruleB);
-  const { name } = ruleA.parent;
+  const name = ruleA.parent && ruleA.parent.name ? ruleA.parent.name : null;
   if (parent && name && ~name.indexOf('keyframes')) {
     return false;
   }


ludofischer:
Can you supply a test that fails without the fix and passes with it? I'm afraid this fix masks a bug somewhere else, as it looke like the AST shape is not what the code expects it to be at this point,  maybe this whole block of code should not be called in this case.  What PostCSS and cssnano versions are you using?

sigveio:
@marcusds are you able to provide us with more information on this please? A test case would be useful, like @ludofischer suggests. Or at least steps to reproduce it, and the output of running `npx envinfo && npm ls cssnano`. Thanks!

PR #1067, opened 2021-05-06
postcss-minify-selectors: update selector parser

ludofischer:
Replace the old selector parser release used by `postcss-minify-selectors` so all packages use the new `postcss-selector-parser` 6.0.0 release.

I have tried to follow the postcss-selector parse [API docs](https://github.com/postcss/postcss-selector-parser/blob/master/API.md) but I was forced to operate on raw attribute values since the automatic quotes removal introduced in postcss-selector-parser 4.0 does not seem to work well in some cases (see https://github.com/postcss/postcss-selector-parser/pull/245).

### DIFF
--- a/packages/cssnano/src/__tests__/postcss-minify-selectors.js
+++ b/packages/cssnano/src/__tests__/postcss-minify-selectors.js
@@ -79,7 +79,7 @@ test(
   'should handle deep combinators',
   processCss(
     'body /deep/ .theme-element{color:#00f}',
-    'body /deep/ .theme-element{color:#00f}'
+    'body/deep/.theme-element{color:#00f}'
   )
 );
 
--- a/packages/postcss-minify-selectors/src/__tests__/index.js
+++ b/packages/postcss-minify-selectors/src/__tests__/index.js
@@ -87,7 +87,10 @@ test(
 
 test(
   'should handle deep combinators',
-  passthroughCSS('body /deep/ .theme-element{color:blue}')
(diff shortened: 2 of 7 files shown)


packages/cssnano/src/__tests__/postcss-minify-selectors.js
```diff
@@ -79,7 +79,7 @@ test(
   'should handle deep combinators',
   processCss(
     'body /deep/ .theme-element{color:#00f}',
-    'body /deep/ .theme-element{color:#00f}'
+    'body/deep/.theme-element{color:#00f}'
```
ludofischer, line 82: I think this is correct because the parser now recognizes [named combinators](https://github.com/postcss/postcss-selector-parser/blob/master/CHANGELOG.md#support-for-named-combinators) so is able to properly remove spaces around them.

packages/postcss-minify-selectors/src/__tests__/index.js
```diff
@@ -87,7 +87,10 @@ test(
 
 test(
   'should handle deep combinators',
-  passthroughCSS('body /deep/ .theme-element{color:blue}')
+  processCSS(
+    'body /deep/ .theme-element{color:blue}',
+    'body/deep/.theme-element{color:blue}'
```
ludofischer, line 92: Same as other comment.

alexander-akait (approved):
Good job

PR #1072, opened 2021-05-11
fix(postcss-merge-rules): prevent breaking rule merges

### LINKED ISSUE #999: [Bug] : postcss-merge-rules should only merge well-known and well-supported selectors
If a style rule has a single selector that the browser doesn't understand, the entire style rule is ignored (in my opinion, this is the worst behavior in CSS, but that's neither here nor there). This means that any time `postcss-merge-rules` combines two rules, there's a risk that one of those rules isn't understood and this causes *both* rules to be ignored, changing the behavior of the CSS.

For example:

```css
p {color: blue}
:nonsense {color: blue}
```

Cssnano currently merges these rules, converting this to:

```css
p,:nonsense {color: blue}
```

Which [doesn't color the text](https://codepen.io/nex3/pen/vYyJyjr).

Note that this isn't just an issue for *invalid CSS*—even valid selectors that aren't yet supported by the browser rendering the CSS will break. As such, I think selectors should *only* be merged if they contain a specific set of known, well-supported selectors (possibly enhanced with browserslist if that's feasible).

ludofischer:
This PR fixes two related issues.

First commit, fixes: https://github.com/cssnano/cssnano/issues/730 by _not_ merging the `:host` selector with others.
I have rebased https://github.com/cssnano/cssnano/pull/731/ as it seems to me it fixes a legitimate problem (https://github.com/angular/angular-cli/issues/18672).
Although this comment https://github.com/cssnano/cssnano/pull/876#issuecomment-641435927 says the PR is already done, the test fails on master.
There's a comment https://github.com/cssnano/cssnano/issues/730#issuecomment-484067090 mentioning the fix is not safe if the browser does not support `:host` but it seems the current situation is what would caus the bug. 

As https://github.com/cssnano/cssnano/issues/999 states, merging unknown and known selectors causes the browser to ignore the whole rule, so _not_ merging should be the safe path.
Second commit fixes #999. If the pseudo-selector is unknown and does not start with a vendor prefix, do not merge.


alexander-akait (approved):

ludofischer:
I think the pseudo-selector list https://github.com/cssnano/cssnano/blob/master/packages/postcss-merge-rules/src/lib/ensureCompatibility.js is incomplete, which causes deoptimizations after this pull request. For example, the list is missing `visited`, so we get:

```
blockquote cite a, blockquote cite a:visited {color: #555} 
```

becomes

```
blockquote cite a { color: #255 }blockquote cite a:visited {color: #555}
```

That's because before it sees `:visited` as an unknown pseudo-selector, so it does not merge it anymore.

PR #1096, opened 2021-05-12
chore(postcss-zindex): refactor to drop dependencies

ludofischer:
Replace uniqs and has with a map.
### Updated algorithm
1. for each z-index, create a map from the z-index to itself
2. for every map key, update the value to the optimized z-index
3. replace every z-index by retrieving the value corresponding to that z-index key

Compared to the previous version, it has a few advantages
1. no external dependencies
2. fewer steps (no need to deduplicate an array)
3. I find it clearer uses the same data structure from beginning to end, while previously it swapped an array with an object midway,


packages/postcss-zindex/src/index.js
```diff
@@ -29,7 +29,7 @@ function pluginCreator(opts = {}) {
             return;
           }
 
-          cache.optimizeValues();
+          cache.optimizeValues(opts.startIndex || 1);
```
ludofischer, line 32: The start index is only used inside `optimizeValues()` so I find passing it as an argument reduces indirection (instead of first storing it as an instance variable)

packages/postcss-zindex/src/lib/layerCache.js
```diff
@@ -1,33 +1,16 @@
...
-  list[value] = index + this._startIndex;
-
-  return list;
-}
-
-LayerCache.prototype._findValue = function (value) {
-  if (has(this._values, value)) {
-    return this._values[value];
+LayerCache.prototype.optimizeValues = function (startIndex) {
+  const sortedValues = Array.from(this._values.keys()).sort(ascending);
```
ludofischer, line 10: Assign to each z-index its rank in the sorted array.

packages/postcss-zindex/src/lib/layerCache.js
```diff
@@ -1,33 +1,16 @@
...
+    this._values.set(sortedValues[i], i + startIndex);
   }
-
-  return false;
-};
-
-LayerCache.prototype.optimizeValues = function () {
-  this._values = uniq(this._values)
-    .sort(ascending)
-    .reduce(reduceValues.bind(this), {});
```
ludofischer, line 30: This is where the old code does its 'magic'. `this._values` is initially an array, which then gets assigned to an object here.

packages/postcss-zindex/src/lib/layerCache.js
```diff
@@ -1,33 +1,16 @@
...
   return a - b;
 }
 
-function reduceValues(list, value, index) {
-  list[value] = index + this._startIndex;
-
-  return list;
-}
-
-LayerCache.prototype._findValue = function (value) {
```
ludofischer, line 19: This whole method is not needed anymore, since we can access the map by key directly.

alexander-akait (dismissed):

alexander-akait (approved):

alexander-akait:
Less deps === less problems

PR #1107, opened 2021-05-18
feat(postcss-colormin): switch to colord and solve multiple issues

### LINKED ISSUE #819: postcss-colormin: incorrect "colormin"'s color conversion
I added an example archive.

```rgba(128, 128, 128, .5)``` converted to float percent:
```hsla(0,0%,50.2%,.5)```, but must be integer percent, like this:
```hsla(0,0%,50%,.5)```

[postcss-colormin.zip](https://github.com/cssnano/cssnano/files/3609108/postcss-colormin.zip)

### LINKED ISSUE #1042: [Bug] : postcss-colormin produces odd results for #rrggbbaa input
Using `postcss-colormin@4.0.3`:

Original:
```css
:root {
 --color-light:#00000004;
 --color-light-border:#0000001d;
 --color-hover:#0000000f;
 --color-active:#00000014;
}
```

After colormin:
```css
:root {
  --color-light:rgba(0,0,0,0.01568627450980392);
  --color-light-border:rgba(0,0,0,0.11372549019607843);
  --color-hover:rgba(0,0,0,0.058823529411764705);
  --color-active:rgba(0,0,0,0.0784313725490196);
}
```

I think in this case, `colormin` should not touch the original color because it is already pretty minimal.

ludofischer:
* ensure that the replacement color values are shorter than the original. Previously, the plugin returned the converted value without checking whether it was shorter than the original, we now ensure we always return the shortest color value
* rework the colomin cache so use only one cache as in https://github.com/cssnano/cssnano/pull/771 (without using Ramda as in that PR). The cache key is now just the declaration value. Before there were two caches, one with a CSS `property|value` key and one that cached the single colors.
* change the underlying library to [colord](https://github.com/omgovich/colord).
Using `colord` comes with some trade-offs. On the plus side, it automatically rounds up up, which fixes an [open issue](#819) without additional work, reduces transitive dependencies from 5 to 0 (although I think the previous 4 are all maintained by the same people), is smaller and claims to be faster (haven't checked but I suspect css-value-parser is more the bottle neck than color conversion here).
On the minus side, the `colord` parser is very lenient so it accepts invalid CSS colors (rgb with mixed percentages and numbers) which cssnano did not convert before, so I've added some checks to skip these values. Overall there are less lines of code.
* get rid of the JSON file with precomputed CSS color names. I guess the idea was that the JSON file only contained color names that are shorter than the hex, but the color conversion libraries include all color names anyway to be able to parse, so in fact we ship the color names twice. I don't think the extra code for generating the file, plus loading the JSON is worth it to avoid a an extra length check.

fix(postcss-colormin): do not replace original value if shorter than conversion

fix #1042

feat(postcss-colormin): change to colord

* round color values fix #819
* reduce dependencies

perf(postcss-colormin): improve caching

Fix #771

refactor(postcss-colormin): drop the generate script


alexander-akait (approved):

alexander-akait:
Great job!

ludofischer:
Performance might have improved as well. I've run a quick test with [benchmark.js](https://benchmarkjs.com/) and I get

Branch|Mean|Standard deviation
---------|---------|---------------
master|64ms|27ms
colord|39ms|18ms

omgovich:
Hey guys! Thanks for choosing colord as a color manipulation library.
I'll check the code later and think about how to improve your DX.

If you'll face any issues, contact me here or via issues in the colord's repo.
I'm always glad to help.

omgovich:
BTW guys, are 4 and 8 digit color formats are supported by the cssnano?
Performing conversion to the shorthand alpha-hex string might make the color output way lighter.

alexander-akait:
> BTW guys, are 4 and 8 digit color formats are supported by the cssnano?

https://github.com/cssnano/cssnano/pull/1107/files#diff-7f46afa35b95730b615fe24cb6667201f6471b587abd70c48841fe78c8c047fdR136

ludofischer:
> BTW guys, are 4 and 8 digit color formats are supported by the cssnano?
> Performing conversion to the shorthand alpha-hex string might make the color output way lighter.

It tries to convert it to a shorthand hex only if alpha is 1 https://github.com/cssnano/cssnano/blob/28c247175032fa03f04911cde56ad82d74d211cc/packages/postcss-colormin/src/colours.js#L26 Is that enough?

omgovich:
By using 4 digit hexadecimal `"#RGBA"` format we could minify some rgba values.
Here is an example:
`rgba(0, 0, 0, 0.4)` -> `#0006` (same as `#00000066`)
`#0006` is shorter than any other option.

omgovich:
The only thing that bothers me is the browser support. As far as I know, `#RGBA` format is part of CSS Color Level 4, and I don't know a good way to check how many browsers support that.

alexander-akait:
We can implement this as feature and check support using `browserslist`, feel free to open an issue

omgovich:
If you don't mind I'll create a pull request and ask for your help with configuring the `browserlist` there.

omgovich:
Created #1109. take a look please when you'll have time.

PR #1109, opened 2021-05-19
feat(postcss-colormin): Output 4 and 8 digit hex colors

omgovich:
See discussion in #1107

1. I refactored the code a bit.
2. Rewrore the module to output 4 and 8 digit colors if they are shorter than other color formats.


alexander-akait (commented):
We need check color is supported before ouptut 4/8 digest, also can we add more tests?

omgovich:
How to perform the proper browser support check? I'm not familiar with the patterns cssnano used to add to connect the plugin to browserlist.

alexander-akait:
Here example https://github.com/cssnano/cssnano/blob/master/packages/postcss-reduce-initial/src/index.js#L23, maybe we need move it to utils...

omgovich:
@alexander-akait The PR is ready for review =)

ludofischer:
Do you know why rgba alpha is not rounded up anymore. In the integration tests,

```
rgba(0,0,0,.08)
```

becomes
```
rgba(0,0,0,.075)
```

packages/postcss-colormin/src/lib/color.js
```diff
@@ -0,0 +1,67 @@
+import { colord, extend, getFormat } from 'colord';
+import namesPlugin from 'colord/plugins/names';
+import getShortestString from './getShortestString';
+
+let minifierPlugin = (Colord) => {
+  /**
+   * Shortens a color to 3 or 4 digit hexadecimal string if it's possible.
+   * Returns the original (6 or 8 digit) hex if the it can't be shortened.
+   */
+  Colord.prototype.toShortHex = function ({ formatAlpha }) {
```
ludofischer, line 10: Have you thought of making this plugin part of the colord package instead so it could also be used outside of cssnano (maybe more people would try the minification so it would become better over time, although it could make life harder for cssnano as it would lose control of some of the minification logic, but )

packages/postcss-colormin/src/index.js
```diff
@@ -1,6 +1,7 @@
 import browserslist from 'browserslist';
+import { isSupported } from 'caniuse-api';
 import valueParser, { stringify } from 'postcss-value-parser';
-import colormin from './colours';
+import minifyColour from './colours';
```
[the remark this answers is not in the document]
omgovich: Agree. Will do)

packages/postcss-colormin/src/lib/color.js
omgovich: Yes, I have thought) But, since we're having some browser support conditions, I didn't wonder a good way to create a good plugin that can be used here.

If I create the plugin, I'll definitely come back and simplify your code =)

ludofischer:
Hi, can you rebase on master to resolve the conflicts?

omgovich:
> Do you know why rgba alpha is not rounded up anymore. In the integration tests,
> 
> ```
> rgba(0,0,0,.08)
> ```
> 
> becomes
> 
> ```
> rgba(0,0,0,.075)
> ```

- Source file contains `"rgba(0, 0, 0, .075)"`.
- `Colord.toShortString` returns `"rgba(0, 0, 0, 0.08)"`.
- Since  the initial values doesn't have a leading zero, both strings a have the same length, so plugin uses the source value.

omgovich:
Done)

ludofischer:
> I think we should add a color minification inside of this plugin (spaces and leading zeros) to produce better results, but it's a task for a separate PR I assume.

Yes, better a separate issue. In fact from what I could gather spaces handling is a bit unclear in cssnano, some plugins seem to minify spaces on their own but there's also the normalize-whitespace plugin.

ludofischer (dismissed):
Good job!

alexander-akait (commented):
I think `rgba(0,0,0,.08)` and `rgba(0,0,0,.075)` will be problem, it is different colors...

ludofischer:
> I think `rgba(0,0,0,.08)` and `rgba(0,0,0,.075)` will be problem, it is different colors...

Ouch, you're right, I thought the difference would not be visible but they look quite different depending on the background. Sorry @omgovich, I think it needs to preserve the original value in that case.

omgovich:
Alright. I'll think how to solve this. Probably, I just need to change the alpha precition inside of colord from 2 to 3.

ludofischer:
@omgovich it turns out the `0.08` was a bug because the original value is `0.075` is in the unminified CSS. So this PR actually improves that particular case, yay! Still better to add some tests to ensure it does not round the opacity too much.

omgovich:
Give me a day. I'll think about making the alpha transformations safer.

omgovich:
Hey guys! The alpha precision is improved.
I also added a condition to make sure that conversion to alpha-hex is lossless.

packages/cssnano-preset-advanced/src/__tests__/integrations/bootstrap-v3.4.0.css
```diff
@@ -4,4 +4,4 @@
...
...y:block!important}table.visible-lg{display:table!important}tr.visible-lg{display:table-row!important}td.visible-lg,th.visible-lg{display:table-cell!important}.visible-lg-block{display:block!important}.visible-lg-inline{display:inline!important}.visible-lg-inline-block{display:inline-block!important}}@media (max-width:767px){.hidden-xs{display:none!important}}@media (min-width:768px) and (max-width:991px){.hidden-sm{display:none!important}}@media (min-width:992px) and (max-width:1199px){.hidden-md{display:none!important}}@media (min-width:1200px){.hidden-lg{display:none!important}}.visible-print{display:none!important}@media print{.visible-print{display:block!important}table.visible-print{display:table!important}tr.visible-print{display:table-row!important}td.visible-print,th.visible-print{display:table-cell!important}}.visible-print-block{display:none!important}@media print{.visible-print-block{display:block!important}}.visible-print-inline{display:none!important}@media print{.visible-print-inline{display:inline!important}}.visible-print-inline-block{display:none!important}@media print{.visible-print-inline-block{display:inline-block!important}.hidden-print{display:none!important}}
```
alexander-akait, line 7: Is it possible to keep old value if it's shorter, I am afraid that with this improvement we will increase the size of compressed css with old color notations

omgovich: Not sure I understand. The plugin preserves the initial value if it's shorter.

omgovich: The reason why snapshots are different now since the updated version outputs the alpha values that the source files contain.

omgovich: Previous version converted `rgba(0,0,0,.075)` to `rgba(0,0,0,.08)` for example. It was kinda bug that is fixed now.

omgovich: So the plugin outputs the same alpha values it did before migration to colord.

alexander-akait (approved):
/cc @ludofischer I think we need release after merge, here and bug and fix

packages/cssnano-preset-advanced/src/__tests__/integrations/bootstrap-v3.4.0.css
ludofischer: I had noticed the change but I did not realize such small alpha changes would be perceptible.

PR #1118, opened 2021-05-20
fix(postcss-colormin): prevent interpreting all numbers as hex colors.

### LINKED ISSUE #1113: [Bug] : postcss-colormin transforms z-index: 999999 to z-index: #999
**Describe the bug**
postcss-colormin transforms `z-index: 999999` to `z-index: #999`

**To Reproduce**
```js
var colormin = require('postcss-colormin');

var css = 'div {z-index: 999999} div {z-index: 9999999}';
console.log(postcss(colormin()).process(css).css);
```

**Expected behavior**
Should not be transformed to a hex

**Screenshots**
Screenshot 2021-05-20 at 14 25 22

cssnano@5.0.3
postcss-colormin@5.1.0

OS: macOS Mojave 10.14.6
Node: 16.2.0

### LINKED ISSUE #1115: [Bug] : Hashes (#) were added to the RGB color format
Hashes (#) were added to the RGB color format:
https://github.com/angular/angular-cli/issues/20869#issuecomment-845029637

Expected:
image

In fact:
image

I set the values as in the first screenshot and expect to get them in the output.
We need such values so that we can further process them, for example, add transparency, like this: rgba(var(--gray-10), 0.3);
If it is just a variable with rgb(), then it will not work.

ludofischer:
The issue is that the colord parser is so lenient it accepts single numbers
as hex colors.

Fix #1113, fix #1115


ludofischer:
@omgovich do you agree with this solution? I wonder if the colord parser is not a bit _too_ lenient. The interesting part is that the cssnano code never checks if color can be at that position, I guess because it is a bit complicated with all the shorthand declarations, so it relies on the color library failing to parse to decide whether to continue or not. Because colord accepts more values than color, cssnano ends up transforming any number with three digits.

omgovich:
Hey guys. Sad to see these issues.

Colord was created to help people to type their color and fix possible mistakes there. It produces valid css colors, but parses even broken ones. I've never aimed to support any other input then possible color value.

The current workaround is good enought I guess.
I'll try to find a more elegant solution for cssnano soon. Maybe implement some `strict` option to make the parser follow CSS specs.

ludofischer:
We could use more readable integration test. The regression was visible in the generated output, and I even looked at it, but because the result is a single minified string, it is very easy to miss details. Maybe we should also run one plugin at a time on the sample CSS.

omgovich:
Already working on `strict` mode implementation for Colord.
Hope I'll release an update tomorrow and we'll be able to get rid of the workarounds and hacks.

alexander-akait:
@ludofischer Do you think we should merge this and do fast release or better to migrate on v2 in other PR?

ludofischer:
I think we should copy the tests from this in the other PR and release the
other because this one is not that well tested either, so it's better to
concentrate our attention on one PR

michaelfaith:
Just to get a sense for timing, any thoughts when this might make it into a release?  Just trying to guage whether or not i should use a workaround or wait for the fix.

omgovich:
Everything is ready on my side. See #1122

ludofischer:
Superseded by #1122

cssnano 5.0.4 with fixed color parsing is out, please give it a try.

PR #1122, opened 2021-05-21
fix(postcss-colormin): Strict color parsing

### LINKED ISSUE #1113: [Bug] : postcss-colormin transforms z-index: 999999 to z-index: #999

### LINKED ISSUE #1115: [Bug] : Hashes (#) were added to the RGB color format

### LINKED ISSUE #1120: [Bug] : postcss-colormin change to colord parser breaks rgba colors with variable alpha
**Describe the bug**
Given the folowing color definition:

```
--tw-bg-opacity:0.5;
background-color:rgba(51,153,255,var(--tw-bg-opacity))
```

Parsing with the new `colord` package gives an alpha of 1 every time (this occurs if the variable is not included in the parse aswell).

` {parsed: Object {r: 51, g: 153, b: 255, a: 1}, rgba: Object {r: 51, g: 153, b: 255, a: 1}}`

This will then trigger the plugin to go through the normal flow and turn the rgba definition into hex:

```
if (parsed.isValid()) {    
    const alpha = parsed.alpha();   
    
    if (alpha === 1) {      
        const toHex = toShorthand(parsed.toHex());      
        const toName = parsed.toName();      

        if (toName && toName.length < toHex.length) {        
              return toName;     
        } else {       
              return toHex;     
        }
```

Essentially removing the entire logic with the variable opacity.

This did not occur previously because the package used for parsing (`color`) fails to parse colors such as:

`rgba(51,153,255,var(--tw-bg-opacity))`

This triggers the catch block to execute, returning the original color definition. This way the variable opacity logic stays intact.

As far as I can tell this was introduced with the following PR:

https://github.com/cssnano/cssnano/pull/1107/commits/1d36ad1df8d098e40dc25fc0a02724f87ba1e144

The color I gave as an example is generated by tailwindcss. This possibly affects a lot of projects.

omgovich:
Fixes #1120
Fixes #1115
Fixes #1113

Colord v2 conforms to CSS Color Level Specs and ignores invalid `<color>` values, so upgrading will fix all of the issues we got last few days.

See https://github.com/omgovich/colord/pull/53 for more details.


alexander-akait (commented):
Let's add tests to ensure we don't break it again :+1:

omgovich:
How do you work with the integration tests BTW? They are HUGE and it seems impossible to check them.
They are falling (probably because they were updated during the migration to color v1.7) but I can't even see where is the error.

alexander-akait:
> How do you work with the integration tests BTW? They are HUGE and it seems impossible to check them.

Agree, we need improve them, right now you can use smart diff locally and found which characters was changed (and maybe found regression)

omgovich:
Ooof. Your integrations tests are killing me 🙃 

1. What is smart diff? Can't google the thing)
2. How to update integration snapshots properly?

ludofischer:
git diff --word-diff

omgovich:
Snapshots are fixed (tons of `z-index: #999` were fixed there), so we should be ready to go!

alexander-akait (approved):

ludofischer (approved):
Seems straightforward enough to me. Let's try and release this!

ludofischer:
Thanks a lot @omgovich for your hard work!

michaelfaith:
Thanks for the quick work on this guys.

omgovich:
Always glad to help. Let me know if you need something else :)

PR #1124, opened 2021-05-21
Publish cssnano 5.0.4

ludofischer:
* postcss-colormin 5.1.1
* cssnano-preset-default 5.1.1
* cssnano-preset-advanced 5.1.1

Please check carefully that the version numbers match.
This time I generated the changelog by hand as lerna keeps repeating the same changelog entries over and over again and I cannot understand why it bumbs version numbers for packages where nothing changes. We need to find some other solution as lerna is somewhat unmaintained.

### DIFF
--- a/packages/cssnano-preset-advanced/CHANGELOG.md
+++ b/packages/cssnano-preset-advanced/CHANGELOG.md
@@ -3,6 +3,15 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+# [5.1.1] (2021-05-21)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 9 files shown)


alexander-akait:
> lerna is somewhat unmaintained.

Maybe we can write own simple script for releasing

alexander-akait (dismissed):

ludofischer:
I published, updating and testing locally it  seems to handle numbers normally again. This PR needs re-approval because I have updated the changelog on the website.

> Maybe we can write own simple script for releasing

The main annoyance for me is updating versions across different packages correctly. For example, if I bump colormin, I need to bump update the dependency and bump cssnano, cssnano-preset-default and cssnano-preset-advanced but not cssnano-preset-lite. We could hard code the dependency graph, otherwise if we start building it by parsing `package.json` we will end up re-creating lerna.

alexander-akait (approved):

PR #1050, opened 2021-04-19
fix(postcss-svgo): make sure "new URL" does not crash on relative urls

Neunerlei:
I'm using cssnano not directly, but as a dependency through "css-minimizer-webpack-plugin". 
After the update to "css-minimizer-webpack-plugin" 2.0.0 and in turn cssnano 5.0.0 webpack fails with an error:
```
css/bundle.css from Css Minimizer
TypeError [ERR_INVALID_URL]: Invalid URL: ../assets/th-bingen-9472d38cacd57ee37863.eot?#iefix
    at W:\work\th_bingen\website_relaunch\typo\app\src\css\bundle.css:13:3
    at onParseError (internal/url.js:256:9)
    at new URL (internal/url.js:332:5)
    at W:\work\th_bingen\website_relaunch\typo\app\src\node_modules\postcss-svgo\dist\index.js:32:17
    at walk (W:\work\th_bingen\website_relaunch\typo\app\src\node_modules\postcss-value-parser\lib\walk.js:7:16)
    at ValueParser.walk (W:\work\th_bingen\website_relaunch\typo\app\src\node_modules\postcss-value-parser\lib\index.js:18:3)
    at minify (W:\work\th_bingen\website_relaunch\typo\app\src\node_modules\postcss-svgo\dist\index.js:22:23)
    at W:\work\th_bingen\website_relaunch\typo\app\src\node_modules\postcss-svgo\dist\index.js:116:9
    at W:\work\th_bingen\website_relaunch\typo\app\src\node_modules\postcss\lib\container.js:91:18
    at W:\work\th_bingen\website_relaunch\typo\app\src\node_modules\postcss\lib\container.js:74:18
    at AtRule.each (W:\work\th_bingen\website_relaunch\typo\app\src\node_modules\postcss\lib\container.js:60:16)
```

**To Reproduce**
In my Stylesheet I have a part of SASS code: 
```
$thb-icon-web-directory: "../fonts" !default
@font-face
  font-family: "th-bingen"
  src: url("#{$thb-icon-web-directory}/th-bingen.eot")
  src: url("#{$thb-icon-web-directory}/th-bingen.eot?#iefix") format("embedded-opentype"),
  url("#{$thb-icon-web-directory}/th-bingen.woff") format("woff"),
  url("#{$thb-icon-web-directory}/th-bingen.ttf") format("truetype"),
  url("#{$thb-icon-web-directory}/th-bingen.svg#th-bingen") format("svg")
  font-weight: normal
  font-style: normal
[...]

PR #1133, opened 2021-05-25
fix(postcss-svgo): skip non-SVG urls

ludofischer:
Fix #1050

### DIFF
--- a/packages/postcss-svgo/src/__tests__/index.js
+++ b/packages/postcss-svgo/src/__tests__/index.js
@@ -213,6 +213,24 @@ test(
 );
 
 test(
+  'should skip non-SVG urls',
+  passthroughCSS(`@font-face {
+  src: url("../example/dfds.woff2") format("woff2"),
+       url('data:image/svg+xml;charset=utf-8,<svg/>') format("svg");
+  }`)
+);
+
+test(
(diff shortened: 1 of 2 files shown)


packages/postcss-svgo/src/__tests__/index.js
```diff
@@ -226,4 +226,12 @@ test('should not crash on malformed urls when encoded', () => {
   expect(() => decode(svg)).not.toThrow();
 });
 
+test(
+  'should skip non-SVG urls',
+  passthroughCSS(`@font-face {
+  src: url("../example/dfds.woff2") format("woff2"),
```
alexander-akait, line 232: Can we add test with relative SVG URL?

alexander-akait (dismissed):

ludofischer:
Updated. I suspect this is another consequence of https://github.com/cssnano/cssnano/pull/1034 When we removed `is-svg` we started processing too many things that were not SVGs.

alexander-akait:
We can check it using `\.svg` extension, don't forget about `?something=bar` in query strings

alexander-akait (approved):

internalsystemerror:
@ludofischer @alexander-akait thank you both

PR #1134, opened 2021-05-25
chore: use yarn workspaces

ludofischer:
As [we've discussed](https://github.com/cssnano/cssnano/pull/1124#issuecomment-846125562), lerna does not work very well for us. The alternatives I have seen rely workspaces.
so move to yarn workspaces to prepare removing lerna completely. We are still using lerna, but with `useWorkspaces` option turned on. 

Since this recreates the whole dependency tree, it had the effect of bumping Browserslist which caused the integration tests to fail for `cssnano-preset-advanced` (also happens on the [dependabot PR](https://github.com/cssnano/cssnano/pull/1131)). So to check whether workspaces themselves break something, I have first updated all dependencies, and then moved to workspaces in separate commits.

The problem with workspaces is that since everything is hoisted to the root, a package can import a dependency which is not in its `package.json` more easily, but I could not find a solution except switching to pnpm, so I guess we need to be careful if we merge this.


ludofischer:
Closing this as I have the impression there are too many bugs with the lerna + Yarn 1 workspaces (also yarn 1 is not being developed any more, so the workspace feature might not see any bug fix).

PR #1138, opened 2021-05-28
docs: Update website for 5.0.5

ludofischer:
* Update all package versions
* Clean up console.log
* Remove prettier since the lint script in the main repo also lints the site


ludofischer:
Last time we updated the site wrong as the playground gets out of date if we update the site before publishing the new version.

alexander-akait (approved):

site/docusaurus.config.js
```diff
@@ -93,7 +93,7 @@ module.exports = {
           ],
         },
       ],
-      copyright: `Latest release v5.0.3 · Distributed under the MIT License.`,
+      copyright: `Latest release v5.0.5 · Distributed under the MIT License.`,
```
alexander-akait, line 96: Maybe load this from `package.json`?

ludofischer: Yeah, we could generate this.

PR #1144, opened 2021-06-05
fix(postcss-ordered-values): preserve columns count

### LINKED ISSUE #1123: [Bug] : count property of `columns` shortand gets removed
**Describe the bug**
`columns: 2 200px` gets transformed into `columns: 200px`, removing the 'count' property.
`columns: 200px 2` produces the same result.

**To Reproduce**

1. Run cssnano over the above CSS.

**Expected behavior**
I'd expect the count property to be preserved in the value.

**Screenshots**
If applicable, add screenshots to help explain your problem. (from [cssnano-playground](https://cssnano.co/playground/))

- CSSNANO Version 5.0.3
- preset: 'default'
- Add the output of `npx envinfo && npm ls cssnano` here
```
  System:
    OS: Linux 5.11 Ubuntu 21.04 (Hirsute Hippo)
    CPU: (12) x64 Intel(R) Core(TM) i7-10710U CPU @ 1.10GHz
    Memory: 8.02 GB / 31.20 GB
    Container: Yes
    Shell: 5.8 - /usr/bin/zsh
  Binaries:
    Node: 14.17.0 - ~/.nvm/versions/node/v14.17.0/bin/node
    npm: 7.13.0 - ~/.nvm/versions/node/v14.17.0/bin/npm
  Managers:
    Apt: 2.2.3 - /usr/bin/apt
    Composer: 2.0.13 - ~/.config/composer/composer
    pip3: 20.3.4 - /usr/bin/pip3
  Utilities:
    CMake: 3.20.2 - /snap/bin/cmake
    Make: 4.3 - /usr/bin/make
    GCC: 10.3.0 - /usr/bin/gcc
    Git: 2.30.2 - /usr/bin/git
    FFmpeg: 4.3.2 - /usr/bin/ffmpeg
  IDEs:
    Nano: 5.4 - /usr/bin/nano
    VSCode: 1.56.2 - /usr/bin/code
  Languages:
    Bash: 5.1.4 - /usr/bin/bash
    Perl: 5.32.1 - /usr/bin/perl
    PHP: 7.4.16 - /usr/bin/php
    Python3: 3.9.4 - /usr/bin/python3
  Browsers:
    Chrome: 90.0.4430.212
    Firefox: 88.0.1
```

**Additional context**

ludofischer:
Fix #1123

* rewrite columns handling logic
* add tests


ludofischer (commented):
We only want to do something when one value of the `columns` declaration has a unit (so is a width) and the other not. All other cases are either invalid or cannot be optimized through reordering. So the new code collects the widths and other types of values separately as it walks the tree, and reorders only if the condition holds at the end. If we stop as soon as two values are found, then we might make some invalid declarations valid and change the effect of the CSS.

alexander-akait (approved):

PR #1145, opened 2021-06-07
chore(dev): remove obsolete build scripts

ludofischer:
* remove buildPackages script
* remove scripts used only by buildPackages
* remove unused dev deps
* get rid of duplicate fs-extra by updating it to 9.1

`buildPackages` tries to keep the `README` by introspecting the code (it even imports the presets!) but right now it makes the docs worse (replaces a lot of plugin description with `This plugin is loaded with its default configuration.`) I think it's easier to write text by hand than fix this script.


alexander-akait:
Maybe in this case we need write steps how to do release in this case, otherwise somebody from contributors can forget

ludofischer:
> Maybe in this case we need write steps how to do release in this case, otherwise somebody from contributors can forget

I don't think that's necessary.  Ihave never used these scripts to do the release because I remember @anikethsaha already told me they were broken, so I don't think anyone is relying on them. They've probably never been updated since lerna was introduced to the repo  (the READMEs must have been updated by hand for some time, since running the script destroys some of the information). They're not documented in `CONTRIBUTORS.md` either.

alexander-akait:
hm, how you do release right now (steps)?

ludofischer:
> hm, how you do release right now (steps)?

The way it says in[ `CONTRIBUTING.md`](https://github.com/cssnano/cssnano/blob/master/CONTRIBUTING.md#releasing):
```
yarn lerna publish from-package
```
I think `buildPackages` mainly updated the `README`s. It does not even seem to keep the versions in sync in `package.json` I've kept the two other scripts `buildMetadata` and `buildSiteMarkdown` as they're  used to build the website.

alexander-akait (approved):

PR #1154, opened 2021-06-30
fix(site-playground): add missing error handling to fix WSOD (#991)

### LINKED ISSUE #991: [Bug] : your playground crashes when clicking "run" on invalid css
To reproduce:
1. go to https://cssnano.co/playground/
2. enter:
```
.crashes {
  font-size: 15pt
  background: red;
}
```
3. click "run"

Expected behaviour: editor does not turn white and need a page reload

Screen Shot 2021-02-05 at 11 56 20

sigveio:
This pull request aims to fix the playground WSOD reported in #991

I found two related issues leading to this behaviour:

1. Uncaught errors thrown by `prettier.format()` when the "Format" button is used.
2. The error object from caught postcss errors when using "Run" were being passed to `setOutput()` in order to output it to the output pane of the playground. Here it was passed in raw as `EditorProps.value` - which expected a string, not an object.

The two cases throws slightly different types of errors. 

On most syntax errors, Prettier throws a `SyntaxError` wrapping some output from `CssSyntaxError` thrown by the css parser - which is quite verbose with numbered line output and such in the message. There are also some edge cases causing Prettier to throw different types of errors.  And the line numbering from the SyntaxError does not correspond to the input lines.

While postcss throws a clean `CssSyntaxError` with good/usable information.

I therefore introduced a simple error handling function which conforms the errors to a short and concise format (using common factors), and outputs it as a red bar right above the code panes. So now the error will for the most part be identical whether one clicks "Run" or "Format"

Screenshot 2021-06-30 at 09 46 28

This is also more friendly for devices with small screen resolutions, where outputting errors to the output pane would be out of view and could cause confusion (i.e. "nothing happening" when clicking the buttons)

I tried to follow the general style/conventions of the existing code.

Closes #991

### DIFF
--- a/site/src/components/editor/editor.module.css
+++ b/site/src/components/editor/editor.module.css
@@ -40,3 +40,13 @@
   left: 50%;
   z-index: 99;
 }
+
+.inputError {
+  color: #fff;
+  background-color: #bc1515;
+  padding: 0.3em 1em;
+}
+
+.inputError[data-theme='light'] {
+  background-color: #ff5959;
+}
--- a/site/src/components/editor/inputError.js
+++ b/site/src/components/editor/inputError.js
@@ -1,0 +1,14 @@
+import React, { Component } from 'react';
+import styles from './editor.module.css';
+
+class InputError extends Component {
+  render() {
+    return (
+      <div className={styles.inputError} data-theme={this.props.theme}>
+        {this.props.error}
+      </div>
+    );
+  }
+}
+
+export default InputError;
--- a/site/src/pages/playground.js
+++ b/site/src/pages/playground.js
@@ -4,6 +4,7 @@ import prettier from 'prettier/standalone';
 import cssParser from 'prettier/parser-postcss';
 import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
 import className from 'classnames';
+import { CssSyntaxError } from 'postcss';
 import { RingSpinner as Loader } from '../components/editor/RingSpinner.js';
 import MainEditor from '../components/editor/main';
 import OutputEditor from '../components/editor/output';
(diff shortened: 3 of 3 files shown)


sigveio:
In case anyone would be curious, the rationale for choosing this error format is:

From the `SyntaxError` wrapped `CssSyntaxError` (thrown from the Prettier call), we only really have the location and the already formatted error message available:

debug2

Which looks like this:
```
CssSyntaxError: Missed semicolon (3:14)
  1 | /* write your css below */
  2 | .crashes {
> 3 |   font-size: 15pt
    |              ^
  4 |   background: red;
  5 | }
```

While in the `CssSyntaxError` thrown by postcss, the order/format of things is slightly different:

```
CssSyntaxError: <css input>:3:14: Missed semicolon
```

But... here we do have access to the different parts as properties of the error:

debug1

So the easiest way to get them to conform to a common format, without too much fiddly and unnecessary processing, was to simply discard anything after the first line on the errors from Prettier and format the `CssSyntaxError` from postcss in the same order/style.

The line/column number is anyway available in the error, and the editor itself also attempts to highlight. So discarding the verbose output with the line/column indicator doesn't really lose any valuable information. Keeping the message consistent and concise seems more important, considering the location it's displayed.

alexander-akait (approved):
/cc @ludofischer

site/src/components/editor/editor.module.css
```diff
@@ -40,3 +40,13 @@
...
   z-index: 99;
 }
+
+.inputError {
+  color: #fff;
+  background-color: #bc1515;
+  padding: 0.3em 1em;
+}
+
+.inputError[data-theme='light'] {
```
ludofischer, line 50: Nice touch!

ludofischer (approved):
LGTM

site/src/components/editor/editor.module.css
sigveio: Thanks :)

PR #1156, opened 2021-07-03
chore(git): clean up .gitignore files

sigveio:
## Description 

This pull request aims to re-organise the projects `.gitignore` to help keep it clean, and to add various common editor generated folders/files to the list. It also removes some redundant lines from the site/docusaurus `.gitignore`, which are covered by the root.

The pattern `*.bundle.js` is also deleted.
_(From what I could gather this was added ~5 years ago for a compiler that is now deprecated and no longer used)_

## Motivation / Context

I generated a `.vscode/launch.json` file for the vscode debugger while working on a PR, and thought it would be practical for developers to avoid having to stash or delete these types of files between commits. So I wanted to suggest adding common ignores for the most popular IDEs. Which seemed like a good opportunity for some housekeeping =)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,21 +1,42 @@
-.nyc_output
-coverage
-dist
-lerna-debug.log
+# Package manager
 node_modules
-npm-debug.log
-yarn-error.log
+
+# Build
+dist
 _webpackOutput
 
 # Generated metadata (only the root one is generated!)
-
 /metadata.toml
 
-
-# Phenomic things
-
-*.bundle.js
-
 # Cache
-
+.cache
 .eslintcache
+
+# Testing
+.nyc_output
+coverage
+
+# Env
+.env.local
+.env.*.local
+
+# Logging
+*.log
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+lerna-debug.log*
+.pnpm-debug.log*
+
+# Editor
+.idea
+.vscode
+.history
+*.suo
+*.ntvs*
+*.njsproj
+*.sln
+*.sw?
+
+# OS generated
+.DS_Store
--- a/site/.gitignore
+++ b/site/.gitignore
@@ -1,23 +1,6 @@
-# Dependencies
-/node_modules
-
-# Production
+# Build
 /build
 
 # Generated files
 .docusaurus
 .cache-loader
-
-# Misc
-.DS_Store
-.env.local
-.env.development.local
-.env.test.local
-.env.production.local
-
-npm-debug.log*
-yarn-debug.log*
-yarn-error.log*
-
-
-.cache


.gitignore
```diff
@@ -1,21 +1,42 @@
...
+yarn-error.log*
+lerna-debug.log*
+.pnpm-debug.log*
 
-# Cache
+# Editor
+.idea
+.vscode
+.history
+*.suo
```
ludofischer, line 35: I have never seen these extensions:
* `.ntvs*`
* `.njsproj`
* `.sln`
* `.sw?` 

What are they?

.gitignore
```diff
@@ -1,21 +1,42 @@
...
-dist
-lerna-debug.log
+# Package manager
 node_modules
-npm-debug.log
-yarn-error.log
+
+# Build
+dist
 _webpackOutput
```
ludofischer, line 6: Do you think `_webpackOutput` is still needed? What would generate this file? The only occurrence on Google for `_webpackOutput` is... this file.

ludofischer (commented):
Did you gather the list from somewhere or is it based on your own usage only?

.gitignore
sigveio: They are mostly related to Visual Studio. The two first are related to their Node.js tools. And the `.sln` is (like `*.suo`) related to [organising projects](https://docs.microsoft.com/en-us/visualstudio/extensibility/internals/solution-dot-sln-file) there.

`.sw?` takes care of vim temporary files

.gitignore
sigveio: Hmm, good catch. I somehow just assumed it was something meaningful when I saw it was webpack related - but I can't find any external references to it anywhere either. 

From the git blame it seems to have been added by Ben 5 years ago

```bash
❯ grep -rnw cssnano -e '_webpackOutput'
(...)
cssnano/packages/cssnano/src/__tests__/_webpack.config.js:8:    path: path.resolve('./_webpackOutput/'),
```

Test related?

sigveio:
> Did you gather the list from somewhere or is it based on your own usage only?

I boiled it together from a combination of my own standard `.gitignore` / dotfiles and the existing ones in this project. I can't recall where I originally gathered these code editor related extensions from - some are from own usage, and some probably inspired from other `.gitignores` I have come across.

.gitignore
```diff
@@ -1,21 +1,42 @@
...
+.idea
+.vscode
+.history
+*.suo
+*.ntvs*
+*.njsproj
+*.sln
+*.sw?
 
-.eslintcache
```
alexander-akait, line 21: Better to keep it, because eslint generate cache

sigveio: It's just been moved to another location in the file, under the `# Cache` section :-)

.gitignore
ludofischer: You're right, it's a file that the tests create.

ludofischer (approved):

PR #1160, opened 2021-07-07
refactor(postcss-reduce-initial): refactor and add tests for acquire script

### LINKED ISSUE #1159: [Feature Request] Replace html2plaintext dependency in postcss-reduce-initial
## Is your feature request related to a problem? Please describe.

I noticed that the acquire tool in `postcss-reduce-initial` which builds the from/to JSON files depends on the package `html2plaintext` which has a transitive dependency of a vulnerable `css-what` version.

So I headed over to their [repo](https://github.com/kurttheviking/html2plaintext) - and it's not been updated for 3 years.
From what I can see, it doesn't appear to be maintained.

## Describe the solution you'd like

Since it's just a helper tool for generating files for the package I suppose it's not a big issue, but for the sake of prudence I would propose replacing `html2plaintext` with [html-to-text](https://www.npmjs.com/package/html-to-text), which appears to be much more frequently used and actively maintained.

## Are you willing to work on this ?

Yes - happy to submit a PR if you agree 👍

sigveio:
## Summary

This pull request aims to refactor the `acquire` script under `postcss-reduce-initial`

The script is responsible for generating data used by the plugin to determine whether or not there is something to gain from replacing a CSS property value with either the `initial` keyword or a shorter equivalent.

The refactor sets out to do the following:
- Separate concerns into different files where reasonable
- Replace comments with more readable code
- Refactor with testability in mind
- Replace dependencies with more up-to-date and lightweight alternatives
- Add test coverage

Closes #1159 

#### Replaced:

`got` -> `node-fetch`
`html2plaintext` -> `html-to-text`

#### Removed:

`write-file` _(Replaced with node.js' writeFile)_

#### Added:

`nock` _(Adds convenient HTTP server mocking for tests)_

### Overall dependency impact:

By my not so careful calculations, this yields a net result of approx. **25 transient dependencies removed**.
_(Most of the credit for that goes to @ludofischer for pointing out that `got` is dep. heavy)_

And notably removes a vulnerable version of `css-what` - from the seemingly unmaintained `html2plaintext`

## Motivation / Context

My original motivation was to replace `html2plaintext` due to a vulnerability in one of its dependencies. After looking into it more closely and getting feedback from @ludofischer  in #1159 it seemed sensible to add a test suite so that the data files can be generated with confidence in the future (perhaps also automated as part of the build). I also saw an opportunity to make some structural changes to help aid readability/maintainability.

## Are you introducing a breaking change?
- [ ] Yes
- [x] No (refactors auxiliary script)

## Additional notes:

- Test coverage does not (as pointed out by @ludofischer in the issue) currently protect fully from the source file format changing. Although perhaps unlikely to happen, since the source is the [MDN CSS Data repository](https://github.com/mdn/data/tree/master/css), it could perhaps be prudent to add snapshot testing of `data/fromInitial.json` and `data/toInitial.json` to serve as a basic guard against unexpected changes. _(I am leaving this out until I have feedback on whether or not this is desired or not)_
- The unrelated coverage ignore pattern `!packages/postcss-colormin/src/generate.js` was removed from `jest.config.js` as the file no longer seems to exist.
- A `transform` pattern was added to `jest.config.js` to support importing ES modules with .mjs extension in tests.


alexander-akait (approved):

packages/postcss-reduce-initial/src/script/__tests__/mdnCssProps.js
```diff
@@ -0,0 +1,93 @@
+import {
```
ludofischer, line 1: The tests in here are the most useful for me.

packages/postcss-reduce-initial/src/script/lib/mdnCssProps.mjs
```diff
@@ -0,0 +1,53 @@
...
+
+export function toPlainText(value) {
+  return htmlToText(value, {
+    wordwrap: false,
+  })
+    .replace(/[\t\r\n\f\u200b]/g, '')
+    .trim();
+}
+
+export function reduceInitial(peropertyData) {
```
ludofischer, line 33: typo peroperty -> property.

packages/postcss-reduce-initial/src/script/lib/io.mjs
```diff
@@ -0,0 +1,37 @@
...
+export function toJSONString(data) {
+  return `${JSON.stringify(data, null, 2)}\n`;
+}
+
+export function write(fileFunc, filePaths, data, key) {
+  fileFunc(filePaths[key], toJSONString(data[key]), handleError);
+  return data;
+}
+
+export async function generate(jsonURL, fileFunc, filePaths) {
```
ludofischer, line 30: I feel injecting `fileFunc` makes the control flow harder to follow with no benefit since in practice `generate` only gets called once. If you want to make it testable in isolation I would  get rid of  `generate()` altogether, and call `fetch` and `writeFile` directly in `acquire.mjs`.

packages/postcss-reduce-initial/src/script/lib/io.mjs
```diff
@@ -0,0 +1,37 @@
...
+  if (error) {
+    throw error;
+  }
+}
+
+export function toJSONString(data) {
+  return `${JSON.stringify(data, null, 2)}\n`;
+}
+
+export function write(fileFunc, filePaths, data, key) {
```
ludofischer, line 25: I think this function can be misleading since in practice `key`, `filepaths` and `fileFunc` are always the same. I would get rid of it. The exact names are more important than the mechanism since they're the interface with the plugin code.

packages/postcss-reduce-initial/src/script/__tests__/io.js
```diff
@@ -0,0 +1,86 @@
...
+  test('should produce parsable JSON', () => {
+    expect(JSON.parse(toJSONString(rawData))).toEqual(rawData);
+  });
+
+  test('should have a trailing newline', () => {
+    expect(toJSONString(rawData)).toMatch(/\n$/);
+  });
+});
+
+describe('File operations', () => {
```
ludofischer, line 27: I also see this suite as a bit superfluous. It tests the `write()` interface but we don't really care about it since in the production code `fileFunc` is always the same.  Similarly it tests that the file paths are configurable, but we don't  care since in fact the plugin requires a specific name (`fromInitial` and `toInitial`).

packages/postcss-reduce-initial/src/script/lib/io.mjs
```diff
@@ -0,0 +1,37 @@
...
+  fileFunc(filePaths[key], toJSONString(data[key]), handleError);
+  return data;
+}
+
+export async function generate(jsonURL, fileFunc, filePaths) {
+  return fetch(jsonURL)
+    .then((result) => result.json())
+    .then(reduceInitial)
+    .then(validate)
```
ludofischer, line 34: The function can stop here.

ludofischer (commented):
Thank you for your PR, it actually contains an improvement in functionality since it updates the property lists.
My suggestions are

* change the JSON files to modules that export an object so the `import` statements are valid without transpilation
* keep the parsing and tranformation tests since they document the original file format
* remove the mock HTTP tests and the `nock` library, I don't think those tests failing would alert us of anything that we would not immediately notice on running the script
* un-generalize (is that a word?) the file writing, we only need files with a specific name at a specific location

ludofischer:
Also running `npx yarn-deduplicate` can be useful to further reduce the `yarn.lock` size when you update dependencies.

packages/postcss-reduce-initial/data/toInitial.json
```diff
@@ -17,17 +17,19 @@
...
+  "mask-repeat": "no-repeat",
   "mask-type": "luminance",
   "ruby-align": "space-around",
   "ruby-merge": "separate",
+  "ruby-position": "alternate",
   "text-decoration-color": "currentcolor",
   "text-emphasis-color": "currentcolor",
   "text-emphasis-position": "over right",
-  "transform-box": "border-box",
+  "transform-box": "view-box",
```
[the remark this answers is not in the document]
sigveio: From what I could find out, it changed from the 2018 spec. But this change was not picked up by the MDN data repository until [28 May 2020](https://github.com/mdn/data/commit/fcea8355abb23259a5cd456690a19b14dabc4f5a) and now us. So this has actually deviated from the spec for more than 3 years, and from MDN more than one

packages/postcss-reduce-initial/data/fromInitial.json
```diff
@@ -1,7 +1,9 @@
 {
+  "-webkit-line-clamp": "none",
```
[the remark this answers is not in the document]
sigveio: Good point! Fixed

packages/postcss-reduce-initial/src/script/__tests__/io.js
```diff
@@ -0,0 +1,86 @@
...
+
+  test('should have correct data type (string)', () => {
+    expect(typeof toJSONString(rawData)).toBe('string');
+  });
+
+  test('should produce parsable JSON', () => {
+    expect(JSON.parse(toJSONString(rawData))).toEqual(rawData);
+  });
+
+  test('should have a trailing newline', () => {
```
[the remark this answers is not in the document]
sigveio: I assumed it was a choice made by the author of the original script to comply with the POSIX standard and be compatible with tooling as you suggest. One could think that it's a bit of a moot point since it's not a file that should be edited directly, but since it's checked into source control the new line should be there for the sake of the diff tool at least

packages/postcss-reduce-initial/src/script/__tests__/io.js
```diff
@@ -0,0 +1,86 @@
...
+    garply: 'waldo',
+    'fred-plugh': 'xyzzy-thud',
+  };
+
+  test('should have correct data type (string)', () => {
+    expect(typeof toJSONString(rawData)).toBe('string');
+  });
+
+  test('should produce parsable JSON', () => {
+    expect(JSON.parse(toJSONString(rawData))).toEqual(rawData);
```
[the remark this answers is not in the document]
sigveio: My thought with this test was that... if we...

1. Validate that `toJSONString` produces parsable JSON
2. Assume that writeFile does it's thing (not our responsibility to test)
3. In the assertion on line 43 verify that `writeFile` was called with the data this function produces... 

Then we can reasonably assume that valid JSON would be written to the file. 

And indirectly that it would be transpilable, and usable by the plugin

packages/postcss-reduce-initial/src/script/__tests__/io.js
```diff
@@ -0,0 +1,86 @@
...
+    expect(fileFunc).toHaveBeenCalledWith(path, toJSONString(expected), err);
+  });
+
+  test('should handle file operation errors', () => {
+    expect(handleError).not.toThrowError();
+    expect(() => handleError(new Error('something went wrong'))).toThrowError();
+  });
+});
+
+describe('Data fetching', () => {
```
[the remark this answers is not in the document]
sigveio: The intention was to cover generate and assert that at the end of whatever happens inside the function - not actually caring about the specifics of the implementation - the file function is called twice to generate two files.

For this `nock` is useful in order to remove the HTTP request from the equation.

My thinking was that this test on line 56:
`test('should make it through promise chain and write 2 files', async () => {`

... together with the assertions mentioned in [this comment](https://github.com/cssnano/cssnano/pull/1160/files#r666879350) would serve as basic smoke tests for expected behaviour without being too much tried to the implementation.

As long as the general control flow with the injected file function is preserved (allowing the test to pass in a spy) - someone could later replace `node-fetch` with something else during refactoring without breaking the test.

The thinking behind `should throw on failed API call` I think was to ensure that if  someone replace `node-fetch` in the future it doesn't fail silently when the HTTP endpoint simulated by `nock` responds with an error. To help ensure that basic error handling works.

But you are right in that `validate()` should be tested individually - and with that the `should throw on failed API call` test would anyway become redundant since `validate()` would detect the missing data regardless of whether the HTTP client knows it has failed or not.

### In summary

I would leave the first test in this suite, delete the two throw tests, and add a test for `validate()`

packages/postcss-reduce-initial/src/script/lib/io.mjs
```diff
@@ -0,0 +1,37 @@
+import fetch from 'node-fetch';
+import { reduceInitial } from './mdnCssProps.mjs';
+
+export function validate(data) {
```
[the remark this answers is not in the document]
sigveio: Yes, I agree 👍

packages/postcss-reduce-initial/src/script/lib/io.mjs
sigveio: Injecting the `fileFunc` allows the very basic function of `generate` ("should produce two files") to be asserted by a test, through passing in a spy instead of for example `fs.writeFile`. Without having to care about the specifics of the implementation. 

It also allows for the assertion mentioned in number 3 [here](https://github.com/cssnano/cssnano/pull/1160/files#r666879350) to be made.

Also commented more in full [here](https://github.com/cssnano/cssnano/pull/1160/files#r666914354).

I'd argue that not injecting, or moving these things to `acquire.mjs` would make proper test coverage more difficult.

packages/postcss-reduce-initial/src/script/__tests__/io.js
```diff
@@ -0,0 +1,86 @@
...
+  const rawData = {
+    foo: 'bar',
+    'baz-qux': 'quux',
+    quuz: 'corge',
+    garply: 'waldo',
+    'fred-plugh': 'xyzzy-thud',
+  };
+
+  test('should have correct data type (string)', () => {
+    expect(typeof toJSONString(rawData)).toBe('string');
```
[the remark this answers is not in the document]
sigveio: I think my logic was that the intended output of the function `toJSONString` is -- in the crudest sense from it's name -- a string. So if someone were to refactor the code later and lets say... omit the `JSON.stringify` and return a `Buffer` like you mention... it would diverge from what the function is supposed to do.

That return could still be parsable by `JSON.parse` and pass the next test - but theoretically might not be compatible with whatever is responsible for saving the file (which might not be `fs.writeFile`) if that expects a string. Or some other part of the code that in some future revision relies on the output from `toJSONString` to be a string.

```js
const data = { foo: 'bar', baz: 'qux' };
const returnValue = Buffer.from(JSON.stringify(data));
console.log(JSON.parse(returnValue)); // { foo: 'bar', baz: 'qux' }
console.log(returnValue.trim()); // TypeError
```

It's possible I'm erring a bit too much on the side of caution though. What do you think?

packages/postcss-reduce-initial/src/script/lib/mdnCssProps.mjs
sigveio: Uups... thanks, good catch.

packages/postcss-reduce-initial/src/script/__tests__/mdnCssProps.js
```diff
@@ -0,0 +1,93 @@
...
+});
+
+describe('should recognise properties with unpredictable behavior', () => {
+  test.each([
+    ['nonstandard', 'align-items', true],
+    ['nonstandard', 'display', true],
+    ['standard', 'display', true],
+    ['standard', 'align-items', false],
+    ['experimental', 'aspect-ratio', false],
+  ])('isComplexSyntax(%p, %p) expected: %s', (status, key, expected) => {
```
sigveio, line 46: This should say `isUnpredictable`

packages/postcss-reduce-initial/src/script/lib/io.mjs
sigveio: `fileFunc` is injected to [aid in testability](https://github.com/cssnano/cssnano/pull/1160/files#r666925656), and I chose the structure with `filePaths` + `key` and returning the `data` in part because it allowed cleanly chaining the two file writes in the promise chain.

I felt it was cleaner if `write` only concerned itself with the task of writing a single file. And that the responsibility of directing which and how many files are created belongs more in `generate`. Alternatively that control could be moved to `acquire.mjs` - but that would obscure it from testing.

Another factor leading to the choice of passing in `filePaths` from `acquire.mjs` is that use of `__dirname` to get the base path is no longer allowed by Node in ES modules. So it was necessary to build the paths using the new `import.meta.url` - which didn't play well with `jest`. So this separation simplifies things in that regard.

The chaining and `filePaths` + `key` solution - instead of for example having `write` do both writes - also allowed for writing cleaner tests as the responsibilities were more clearly separated. The test against `generate` asserting that two files are written. The test against `write` making a single assertion to verify that it formats the data before it is passed to `writeFile`. And the `test.each()` also indirectly verifying that `write` does indeed write the specific files we target with the `key` - to make it less fragile.

I'm not sure if I understand what you mean with `"interface with the plugin code"` - could you elaborate?

packages/postcss-reduce-initial/src/script/__tests__/io.js
sigveio: It verifies that `write` passes the data through `toJSONString` when calling `writeFiles` - as for the configurable file paths, they are a result of other considerations that I have tried to explain [here](https://github.com/cssnano/cssnano/pull/1160/files#r667018251). The comment [here](https://github.com/cssnano/cssnano/pull/1160/files#r666879350) is also relevant in regards to how the test fits in with others.

I included `should handle file operation errors` for the sake of coverage and the fact that `fs.writeFile` will call this callback regardless of whether there is an error or not... so it should only throw when there is an error passed in.

sigveio:
> Thank you for your PR, it actually contains an improvement in functionality since it updates the property lists.

Glad to hear! :-) Thank you very much for taking time to give such a thorough review - it's much appreciated.

I have done my best to be thorough in replying to your comments. Of course ultimately I will defer to the experience and wisdom of yourself and @alexander-akait - I just wanted to explain my reasoning/thinking behind the various things. 

Apologies for giving you so much to read!

> Also running `npx yarn-deduplicate` can be useful to further reduce the `yarn.lock` size when you update dependencies.

Thanks for the tip. :-)

Running `yarn-deduplicate yarn.lock` or `yarn-deduplicate --list yarn.lock` produces no changes to the file or output on my end though?

packages/postcss-reduce-initial/data/toInitial.json
```diff
@@ -17,17 +17,19 @@
   "box-sizing": "content-box",
   "column-rule-color": "currentcolor",
   "font-synthesis": "weight style",
+  "image-orientation": "from-image",
   "mask-clip": "border-box",
   "mask-mode": "match-source",
   "mask-origin": "border-box",
-  "mask-repeat": "repeat",
+  "mask-repeat": "no-repeat",
```
sigveio, line 24: I traced the spec for this property all the way back to the original `W3C Last Call Working Draft` of 29 October 2013, and it's remained unchanged as `no-repeat` since then. The only occurrence I find of plain `repeat` initial value is in the [original working draft](https://www.w3.org/TR/2013/WD-css-masking-20130620/). So this is also a pretty old bug - which MDN didn't catch in their repo until [this merge](https://github.com/mdn/data/pull/105/files) back in 2017.

Perhaps I should amend the commit message to include a short summary of the changes?

For example:

```
* Fix incorrect initial value of `mask-repeat` and `transform-box`
* Update initial value for `image-orientation` and `ruby-position`
* Add misc. missing props that became stable / standard
```

packages/postcss-reduce-initial/src/script/__tests__/io.js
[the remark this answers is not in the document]
sigveio: I had a little look at this tonight, and using `jest.mock` features with `mockReturnValue` for the one remaining test was sufficient. So I will remove `nock` altogether like you suggest :-)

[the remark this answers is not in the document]
sigveio: I had a little look at this tonight, and I can replace `nock` for the one remaining test with something along the lines of:

```js
// test file: io.js
import fetch from 'node-fetch';

jest.mock('node-fetch');

// Selectively bypass mocking to use a real Response obj
const { Response } = jest.requireActual('node-fetch');

// (...)

test('should make it through promise chain and write 2 files', async () => {
  // (...)
  fetch.mockReturnValue(
    Promise.resolve(new Response(JSON.stringify(testData)))
  );
  // (...)
});
```

The disadvantage of doing so is that the test now becomes `node-fetch` specific, and thus more fragile.

What do you think?

Is removing the `nock` dev-dependency more important than the test being HTTP client agnostic?

packages/postcss-reduce-initial/data/toInitial.json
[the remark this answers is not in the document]
ludofischer: Interesting, thanks!

packages/postcss-reduce-initial/src/script/__tests__/io.js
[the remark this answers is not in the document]
ludofischer: I think this test is not that necessary, as I do not see the code to write the file changing that often. Suppose someone would change that code anyway, why would they assume they are receiving a string?

packages/postcss-reduce-initial/data/toInitial.json
```diff
@@ -17,17 +17,19 @@
...
   "ruby-merge": "separate",
+  "ruby-position": "alternate",
   "text-decoration-color": "currentcolor",
   "text-emphasis-color": "currentcolor",
   "text-emphasis-position": "over right",
-  "transform-box": "border-box",
+  "transform-box": "view-box",
   "transform-origin": "50% 50% 0",
   "vertical-align": "baseline",
   "writing-mode": "horizontal-tb"
```
alexander-akait, line 35: Will be great to look at git history of `toInitial.json` and `fromInitial.json`, some properties were removed due incompatibility between browsers

packages/postcss-reduce-initial/src/script/__tests__/io.js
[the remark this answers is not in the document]
sigveio: Roger that - I'll drop it then 👍

packages/postcss-reduce-initial/data/toInitial.json
ludofischer: Ah, so the files were edited by hand? Good to know!

alexander-akait: It seems that we did this once, so it's better to check, maybe it's worth leaving a comment about this so as not to forget next time

sigveio: Oh... yeah, good to know. But is that desirable to do anyway? 

If such manual overrides are necessary, would it perhaps be more appropriate to make changes to the script rather than the files themselves? So that one can update the content in a repeatable/predictable manner.

One solution could be to have an "override list" in a separate file that replaces the initial value from the MDN data before it's reduced and processed further. Then no extra logic beyond that would be necessary.

Having to examine the git history and try to determine what needs to be kept or not seems asking for trouble

packages/postcss-reduce-initial/data/toInitial.json
sigveio: Looking at the git history, after @alexander-akait's tip about it, I see this one was deliberately changed in [this PR](https://github.com/cssnano/cssnano/pull/679) in 2019. Didn't occur to me to check, as I didn't think these files would be manually overridden.

The [current spec](https://www.w3.org/TR/css-masking-1/) says "no-repeat" while the [editors draft](https://drafts.fxtf.org/css-masking-1/#the-mask-repeat) says "repeat"

What's appropriate to follow here?

packages/postcss-reduce-initial/src/script/__tests__/io.js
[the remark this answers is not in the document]
ludofischer: Good point about version control. But it occurred to me that the eslint and prettier should take care of files ending with a new line, if that is not the case we should fix the linter config instead of adding this unit test. 
```
yarn eslint --print-config packages/postcss-reduce-initial/data/fromInitial.json
```
shows that eslint applies the `prettier/prettier` rules to the JSON files and it seems prettier doe ensure newlines  are present (it's even[ impossible to turn off](https://github.com/prettier/prettier/issues/2575))

packages/postcss-reduce-initial/data/toInitial.json
sigveio: Something like...

```js
// lib/overrideInitial.mjs
const overrides = {
  'property-name-here': {
    comment: 'comment about why this propertys initial value is overriden',
    newInitial: 'foo',
  },
};

export default overrides;
```

Then add `overrideInitial` to the promise chain:

```js
return fetch(jsonURL)
    .then((result) => result.json())
    .then(overrideInitial)
    .then(reduceInitial)
    .then(validate)
```

And the `overrideInitial` function simply checks against the object and replaces any values that matches.

Then `overrideInitial.mjs` would also help document the overrides for the future

packages/postcss-reduce-initial/data/toInitial.json
ludofischer: I think we should base ourselves on what current browsers implement and we should avoid regressions against already filed issues. Unfortunately, which browsers cssnano supports is somewhat unclear, you can see it when people complain about things breaking in IE 11 (recently https://github.com/cssnano/cssnano/issues/1163), because some cssnano plugins use Browserslist configs to determine which transforms are allowed, others just ignore it.

packages/postcss-reduce-initial/data/toInitial.json
ludofischer: I am starting to believe the changes to the JSON files should go in a separate pull request, as checking which replacements should be allowed is going to be quite time-intensive. We need to check that most browsers out there actually adhere to the spec, not just what's allowed in theory.

packages/postcss-reduce-initial/src/script/__tests__/io.js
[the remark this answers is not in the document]
sigveio: Yeah, that's also a valid point.

If I remove the `\n` and open the file in vscode + save it, then the prettier rule does take care of the newline. However if I just generate the file without the trialing `\n` and run `yarn lint` on the project it does not complain or fix it.

Do you have some extra linting beyond that set up in CI that would take care of this? if not something probably needs to be changed to make it happen automagically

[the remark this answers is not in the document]
ludofischer: There's no extra rules in CI.  Are you sure that saving the file in VsCode adds a newline because of the project linter settings? It might be a VScode setting. It seems that the eslint config needs some improvements.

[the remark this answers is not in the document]
ludofischer: Running `yarn prettier --write packages/postcss-reduce-initial/data/fromInitial.json ` _removes_ the newline at the end of the file. I'd say definitely do not test for the presence of a newline, at this point it's not clear whether it makes any difference/which is better.

packages/postcss-reduce-initial/data/toInitial.json
sigveio: Alright - sounds sensible!

But then I would propose including (as part of that new PR) an override mechanism along the lines of what I mention above - so that the manual changes can be tracked and documented in one place and always applied when the script is run to re-generate the files. Rather than manually applied each time.

To aid in house-keeping I could include the original value in the override file too, and print a warning to the console if it doesn't match up with the data fetched from MDN (i.e. might be stale). And do the same if the data from MDN suddenly matches the override, making it redundant.

packages/postcss-reduce-initial/src/script/__tests__/io.js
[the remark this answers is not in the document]
ludofischer: > I would leave the first test in this suite, delete the two throw tests, and add a test for validate()

OK.

> Is removing the nock dev-dependency more important than the test being HTTP client agnostic?

I think using `nock` just once in the project will make it harder for future maintainers to understand the tests. It's especially not worth it in my view since the cssnano package has little to do with HTTP so `nock` is not going to see much use.

packages/postcss-reduce-initial/data/toInitial.json
ludofischer: I guess splitting the two would make it clearer which browser compatibility workarounds the plugin uses.

packages/postcss-reduce-initial/src/script/lib/io.mjs
ludofischer: > I'm not sure if I understand what you mean with "interface with the plugin code" - could you elaborate?
I mean the cssnano plugin only sees the file names, all the rest is an implementation detail. In that sense the download script 'interface' are the JSON file names.

ludofischer:
To summarize, I think you should
* open a separate PR with the JSON file changes
* test `validate()` separately and remove `nock` and the extra tests as detailed in https://github.com/cssnano/cssnano/pull/1160#discussion_r669097303

packages/postcss-reduce-initial/src/script/__tests__/io.js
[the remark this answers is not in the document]
sigveio: Strange that you are seeing that behaviour from prettier - it's the opposite here with the same command. And they also mention that "files always end with a single newline" in their official "rationale" [here](https://prettier.io/docs/en/rationale.html#empty-lines)

I agree that this is something that linting should address in general, to catch cases where it is forgotten/overlooked.

But I'm leaning towards it being appropriate in principle to follow the POSIX standard when writing the file. So that it is validly formatted for most operating systems/tools upon creation - and not relying on external tools such as a linter to update it at some point after. And in that case also leave the test, but update the name so that it documents better why the newline is there.

sigveio:
> To summarize, I think you should

Roger that 👍  Thank you very much for taking the time to follow up; I appreciate it

packages/postcss-reduce-initial/src/script/__tests__/io.js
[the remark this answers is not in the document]
ludofischer: I see it this way: suppose that one day the acquire script does not write a new line anymore. The cssnano plugin would still work (because in fact it does work without a newline), but the test would break. In that case the test failure would be a false signal, since the program in fact would still be working fine. Therefore the test failure gives no information, so the test is redundant.

sigveio:
How would you prefer that I add the changes?

I was thinking either:

1. Rebase this PR on master, drop the data files commit (currently 
b902a1b), and add changes to the two remaining commits using `git commit --fixup` to be autosquashed later. And create one additional PR for the data files.

2. Cherry pick and create two new superseding PRs (one for the refactoring, another for the data files) referencing this one

Perhaps the second option is cleaner in this case?

ludofischer:
> How would you prefer that I add the changes?

Whichever you're more comfortable with, since the commit are going to be squashed on merge it won't make much difference.

sigveio:
Alright; thanks!

Superseded by #1169

PR #1161, opened 2021-07-07
test: fix misc. test warnings

sigveio:
## Summary

This pull request aims to address some minor issues leading to warnings during test running.

A sample of the 3 different types of warnings encountered:

Screenshot 2021-07-07 at 21 29 38
Screenshot 2021-07-07 at 21 29 09
Screenshot 2021-07-07 at 21 28 48

It also adds ignores for two test helper files which had ignore patterns for test running but not coverage reporting.

## Are you introducing a breaking change?
- [ ] Yes
- [x] No (only touches tests - in the form of refactoring)

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -4,6 +4,8 @@ module.exports = {
     'packages/*/src/**/*.js',
     '!packages/postcss-colormin/src/generate.js',
     '!packages/postcss-reduce-initial/src/acquire.js',
+    '!packages/cssnano/src/__tests__/_processCss.js',
+    '!packages/cssnano/src/__tests__/_webpack.config.js',
   ],
   testPathIgnorePatterns: [
     '/node_modules/',
--- a/packages/cssnano-utils/src/__tests__/sameParent.test.js
+++ b/packages/cssnano-utils/src/__tests__/sameParent.test.js
@@ -1,8 +1,15 @@
 import postcss from 'postcss';
 import sameParent from '../sameParent';
 
+const plugin = () => {
+  return {
+    postcssPlugin: 'dummy',
+  };
+};
+plugin.postcss = true;
+
 test('should calculate same parent', () => {
-  return postcss()
+  return postcss(plugin)
     .process('h1 {} h2 {}', { from: undefined })
     .then((result) => {
       const h1 = result.root.nodes[0];
@@ -13,7 +20,7 @@ test('should calculate same parent', () => {
 });
 
 test('should calculate same parent (detached nodes)', () => {
-  return postcss()
+  return postcss(plugin)
     .process('h1 {} h2 {}', { from: undefined })
     .then((result) => {
       const h1 = result.root.nodes[0];
@@ -27,7 +34,7 @@ test('should calculate same parent (detached nodes)', () => {
 });
 
 test('should calculate same parent (at rules)', () => {
-  return postcss()
+  return postcss(plugin)
     .process('@media screen{h1 {} h2 {}}', { from: undefined })
     .then((result) => {
       const h1 = result.root.nodes[0].nodes[0];
@@ -38,7 +45,7 @@ test('should calculate same parent (at rules)', () => {
 });
 
 test('should calculate same parent (multiple at rules)', () => {
-  return postcss()
+  return postcss(plugin)
     .process('@media screen{h1 {}} @media screen{h2 {}}', { from: undefined })
     .then((result) => {
       const h1 = result.root.nodes[0].nodes[0];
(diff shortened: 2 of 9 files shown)


packages/cssnano/src/__tests__/issue927.js
```diff
@@ -13,17 +13,20 @@ p{
 const expected = `div{grid-column:span 2}p{column-count:2}`;
 
 test('it should compress the columns', () => {
-  const processor = postcss([
-    postcss.plugin('cloner', () => {
-      return (css) => {
-        css.walkAtRules((rule) => {
-          css.prepend(rule.clone());
+  const plugin = () => {
```
ludofischer, line 16: Actually I had left the old style plugins on purpose to check it still worked with those... I guess we can leave one old style and one new style?

packages/cssnano-utils/src/__tests__/sameParent.test.js
```diff
@@ -1,8 +1,15 @@
 import postcss from 'postcss';
 import sameParent from '../sameParent';
 
+const plugin = () => {
```
ludofischer, line 4: Isn't there another way to suppress the warnings? In fact we want postcss to do nothing here. I think adding more code to the test does not help with clarity.

jest.config.js
```diff
@@ -4,6 +4,8 @@ module.exports = {
     'packages/*/src/**/*.js',
     '!packages/postcss-colormin/src/generate.js',
     '!packages/postcss-reduce-initial/src/acquire.js',
+    '!packages/cssnano/src/__tests__/_processCss.js',
```
ludofischer, line 7: Good catch!

packages/cssnano/src/__tests__/issue927.js
sigveio: Ah, I see! I thought they had been forgotten/overseen since they were mostly in issue-specific tests or kind of buried in other things. Perhaps I should have asked and not assumed - sorry about that. Just wanted to save you another issue :-P

Was that the case for all of them or just this one?

ludofischer: Let's leave just this one as an old style plugin

packages/cssnano-utils/src/__tests__/sameParent.test.js
sigveio: Well... my rationale was that when I opened up that file, not knowing the code at all, it was not clear to me at first that you did in fact not want postcss to do anything. So although the dummy plugin at the top of the file does add a little bit of code - could it not also help point out the intention by explicitly passing in a "dummy" instead of nothing?

I didn't see any obvious alternative way (e.g. an option for postcss), but admittedly I also didn't look super hard. Would be happy to have another look though

ludofischer: Yeah, I think just adding a comment to explain what's going on is the best solution.

sigveio: Alright - I'll look into it a bit more thoroughly and see if I find a better solution to suppress the warnings then :-)

sigveio: Ah, so there is actually an option to suppress it:

https://github.com/postcss/postcss/blob/10d974f81d2b925a9e8c0912857dcda1bf6720c0/lib/processor.js#L19

So;

```
test('should calculate same parent', () => {
  return postcss()
    .process('h1 {} h2 {}', { from: undefined, hideNothingWarning: true })
    .then((result) => {
      (...)
    });
});
```

Works!

packages/cssnano/src/__tests__/issue927.js
sigveio: I left one old style in addition to the new style as you suggested; that way it's clear to anyone viewing it in the future that it's intentional. I also investigated if there is a way to suppress the warning; currently [there is not](https://github.com/postcss/postcss/blob/1545d8aabd2a920c76a70fdd4e7c43a9464fba5e/lib/postcss.js#L31). But I think we can live with that ^^,

packages/cssnano-utils/src/__tests__/sameParent.test.js
```diff
@@ -3,7 +3,7 @@ import sameParent from '../sameParent';
 
 test('should calculate same parent', () => {
   return postcss()
-    .process('h1 {} h2 {}', { from: undefined })
+    .process('h1 {} h2 {}', { from: undefined, hideNothingWarning: true })
```
ludofischer, line 6: Wow! You've managed to find a hidden postcss option!

ludofischer (approved):

PR #1168, opened 2021-07-18
Improve config documentation and remove 22 dependencies from cssnano

ludofischer:
This PR does two things.
First, I've reworked the documentation to mention the config file priorities and added some tests to verify the basic config loading.
In a separate commit, I've replaced cosmiconfig with [lilconfig](https://github.com/antonk52/lilconfig). This gets rid of 22 dependencies and about 1MB when doing `npm i cssnano`.  This makes cssnano use the same library as the lastest postcss-load-config (https://github.com/postcss/postcss-load-config/commit/d1478640e823b5432b7dc43a0ba6e23c58b7b981).
You can't see the removed dependencies in the repository `yarn.lock` because lerna depends on cosmiconfig.

### DIFF
--- a/packages/cssnano/package.json
+++ b/packages/cssnano/package.json
@@ -24,7 +24,8 @@
     ],
     "license": "MIT",
     "dependencies": {
-        "cosmiconfig": "^7.0.0",
+        "lilconfig": "^2.0.3",
+        "yaml": "^1.10.2",
         "cssnano-preset-default": "^5.1.3",
         "is-resolvable": "^1.1.0"
     },
--- a/packages/cssnano/src/index.js
+++ b/packages/cssnano/src/index.js
@@ -1,6 +1,7 @@
 import path from 'path';
 import postcss from 'postcss';
-import { cosmiconfigSync } from 'cosmiconfig';
+import yaml from 'yaml';
+import { lilconfigSync } from 'lilconfig';
 import isResolvable from 'is-resolvable';
 
 const cssnano = 'cssnano';
@@ -59,7 +60,7 @@ function resolvePreset(preset) {
 
 /*
  * cssnano will look for configuration firstly as options passed
- * directly to it, and failing this it will use cosmiconfig to
+ * directly to it, and failing this it will use lilconfig to
  * load an external file.
  */
 
@@ -76,7 +77,21 @@ function resolveConfig(options) {
     configPath = path.resolve(process.cwd(), options.configFile);
   }
 
-  const configExplorer = cosmiconfigSync(cssnano);
+  const configExplorer = lilconfigSync(cssnano, {
+    searchPlaces: [
+      'package.json',
+      '.cssnanorc',
+      '.cssnanorc.json',
+      '.cssnanorc.yaml',
+      '.cssnanorc.yml',
+      '.cssnanorc.js',
+      'cssnano.config.js',
+    ],
+    loaders: {
+      '.yaml': (filepath, content) => yaml.parse(content),
+      '.yml': (filepath, content) => yaml.parse(content),
+    },
+  });
   const config = configPath
     ? configExplorer.load(configPath)
     : configExplorer.search(searchPath);
--- a/site/docs/config-file.mdx
+++ b/site/docs/config-file.mdx
@@ -21,10 +21,31 @@ export const ExampleChart = () => {
 
 <ExampleChart/>
 
-## `cssnano` config files
+You can configure cssnano either in the PostCSS configuration file or in a dedicated cssnano configuration file. The PostCSS configuration file takes precedence over the dedicated cssnano configuration.
+If you don't pass any configuration, cssnano runs with the `default` preset.
 
-We use configuration for `cssnano` using file name as `cssnano.config.js` , `cssnano` property in your `package.json`, using `cssnano.config.json` and `.cssnanorc` as well.
-We are using `cosmiconfig` in order to load cssnano config. Please read [here](https://github.com/davidtheclark/cosmiconfig) for more details
+
+## Configure through PostCSS configuration files
+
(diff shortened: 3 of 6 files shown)


packages/cssnano/package.json
```diff
@@ -24,7 +24,8 @@
     ],
     "license": "MIT",
     "dependencies": {
-        "cosmiconfig": "^7.0.0",
+        "lilconfig": "^2.0.3",
+        "yaml": "^1.10.2",
```
ludofischer, line 28: I was uncertain whether to keep YAML config parsing since it was never documented explicitly, but removing would be a breaking change in theory.

site/docs/config-file.mdx
```diff
@@ -85,7 +120,3 @@ These plugins will run after once all presets operations are complete.
   }
   ```
 
-## Alternatives
```
ludofischer, line 88: The section about the PostCSS config file has been moved to the top as that is what most people will use and the PostCSS config overrides the config in the cssnano files.

site/docs/config-file.mdx
```diff
@@ -21,46 +21,81 @@ export const ExampleChart = () => {
...
 - **Type:** `Array<'string' | 'function' | ['string' | 'function', Object<Options for the plugin here>]>`
 
-> If you want to pass config to the plugin, you need to use Array's of array for the plugins i.e `plugins: [ ['autoprefixer', {}] ]`
+In addition to the preset, you can pass a list of plugins to cssnano.
+This is equivalent to adding the plugins after cssnano in the PostCSS plugins array.
+If you want to configure the individual plugins, use an array of arrays:
 
-These plugins will run after once all presets operations are complete.
+```js
+plugins: [ ['autoprefixer', {}] ]
```
sigveio, line 97: Could it be clearer for a novice user to expand the example slightly?

```js
plugins: [ ['autoprefixer', { flexbox: false }], ... ]
```

or something like this to show multiple plugins configured individually..

```js
plugins: [ 
  ['autoprefixer', { flexbox: false }],
  ['pixrem', { rootValue: 12, replace: true }],
]
```

sigveio (commented):
All-in-all good job - this documentation seems a lot clearer now 👍 

And nice one with the deps!!

site/docs/config-file.mdx
sigveio: Oh, scratch this comment. I see it's covered in the example below.

site/docs/config-file.mdx
```diff
@@ -21,46 +21,81 @@ export const ExampleChart = () => {
...
+```js
+cssnano({preset: 'default'})
+```
+
+Preset themselves can take options.
+Pass options to the preset by using an array where the first element is the preset and the second is an object with the preset options. 
+You can specify a preset with the preset name as a string or by passing the result of importing the preset package.
+
+```js
+preset: [require('cssnano-preset-default'), {discardComments: false}]
```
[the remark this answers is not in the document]
ludofischer: I've updated the examples

packages/cssnano/package.json
sigveio: Sounds like a good call to keep it for now, even if the chance of it being used is remote 👍 

Question:

Since the tests you added were paired with a `.cssnanorc.json` and you mentioned keeping YAML support, I decided to attempt running the test suite with a `.cssnanorc.yaml` and a few other variations (replacing `.cssnanorc.json` in the `config_loading` dir) for the sake of diligence. And it doesn't appear to be picked up.

It's entirely possible I'm having some form of "brainfart" and doing something wrong here - but shouldn't that work?

(Out of curiosity I had a look at the `searchItems` found by lilconfig in `lilconfigSync.search()` and it only appears to look for `package.json`, `cssnanorc.json`, `cssnano.config.js`, and `cssnanorc.js`, `.cssnanorc.cjs`, and `cssnano.config.cjs`)

ludofischer: My inner conviction is nobody is really using the YAML format, so I did not bother checking whether it works, I have just copied the instructions on the lilconfig repo. Looking at the `lilconfig` code, it seems it does not even load `.rc` (with no extension) by default.

packages/cssnano/src/index.js
```diff
@@ -76,7 +77,12 @@ function resolveConfig(options) {
     configPath = path.resolve(process.cwd(), options.configFile);
   }
 
-  const configExplorer = cosmiconfigSync(cssnano);
+  const configExplorer = lilconfigSync(cssnano, {
+    loaders: {
+      '.yaml': (filepath, content) => yaml.parse(content),
+      '.yml': (filepath, content) => yaml.parse(content),
+    },
```
sigveio, line 84: I wonder if you may also need to add something like `searchPlaces: ['.cssnanorc.yaml', '.cssnanorc.yml']` to the options in order for lilconfig to actually look for yaml files?

It seems like lilconfig uses this as basis for [merging in the loaders](https://github.com/antonk52/lilconfig/blob/c54ebad1a7916c764a423c8221013c63b2d1ce63/src/index.ts#L105).

Ref. https://github.com/cssnano/cssnano/pull/1168#discussion_r671873676

sigveio: Seems like a bit of a bug or design weakness on their part

ludofischer: I've updated the code with explicit search places. I did try renaming the file to `.cssnanorc.yml` and it worked.

sigveio: Confirmed! And perhaps not such a bad thing for it to be explicit anyway 👍

sigveio (approved):
LGTM 👍

alexander-akait (approved):

ludofischer:
> /cc @sigveio I can invite your in cssnano repo so you can help us more better smile @ludofischer What do you think?

Sure, @sigveio can certainly help.

sigveio:
Thanks @alexander-akait and @ludofischer - I appreciate the trust, and looking forward to help where I can 😄

PR #1207, opened 2021-10-18
Simplify `postcss-colormin`

omgovich:
Hey guys!

`cssnano` already uses `colord` for color conversions in `postcss-colormin` and few other packages, but we added 100+ lines around the library to perform the color modification.

Recently I releases `colord` v2.9 which has built-in color modification utilities so I decided to bring them here and make the cssnano codebase simpler.

The new minifier I implemented inside of `colord` is also way smarter and performs better compression results.

Example:
```
SOURCE
rgba(221, 221, 221, 0.5)

current pocstscss-colormin
hsla(0, 0%, 87%, 0.5)

with colord v2.9 (no spaces, no leading zero)
hsla(0,0%,87%,.5)
```

Don't know how to perform benchmarks here, but it also should work faster.

### DIFF
--- a/packages/postcss-colormin/src/__tests__/minifyColor.js
+++ b/packages/postcss-colormin/src/__tests__/minifyColor.js
@@ -4,6 +4,8 @@ function isEqual(input, output) {
   return () => expect(min(input)).toBe(output);
 }
 
+test('should lowercase keywords', isEqual('RED', 'red'));
+
 test('should convert shorthand hex to keyword', isEqual('#f00', 'red'));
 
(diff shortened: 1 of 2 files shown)


ludofischer:
> Unfortunately, I'm not able to run tests locally (don't know why). Hope you can help me.

From your error message, seems like the ESLint import plugin is failing. What happens if you run `yarn test:only`?

omgovich:
> I guess browser support can be configured on cssnano side by passing options to minify()

It already works this way =)

PR #1214, opened 2021-10-21
chore: update test matrix

ludofischer:
Test on Node.js without increasing the number of tests too much

* test on all supported Node.js versions on MacOS only
* test latest LTS on Linux and Windows
* update codecov action
* use built-in yarn cache functionality


ludofischer:
Running all tests on MacOS because it is the most difficult to check without a physical Mac and is on par with Windows in reported issues, while fewer issues report Linux as the OS.

PR #1224, opened 2021-11-07
refactor: drop one-liner dependencies

ludofischer:
After the recent vulnerability disclosures, I've looked at whether we could further reduce cssnano deps. 
1. we can remove uniqs as it's a one-liner.  The Set method should work on Nodejs 8+ and according to the measurements I have found online it is not slower than array filtering
2. we can remove vendors as it is [a single array](https://github.com/wooorm/vendors/blob/main/index.js) we only use once and it has not changed in the last 5 years, except that the latest release has become ESM-only

### DIFF
--- a/packages/postcss-merge-rules/src/lib/ensureCompatibility.js
+++ b/packages/postcss-merge-rules/src/lib/ensureCompatibility.js
@@ -11,31 +11,15 @@ const cssFirstLine = 'css-first-line';
 const cssInOutOfRange = 'css-in-out-of-range';
 const formValidation = 'form-validation';
 
-/** @type {string[]} */
-const prefixes = [
-  '-ah-',
-  '-apple-',
-  '-atsc-',
-  '-epub-',
-  '-hp-',
-  '-khtml-',
-  '-moz-',
-  '-ms-',
-  '-o-',
-  '-rim-',
-  '-ro-',
-  '-tc-',
-  '-wap-',
-  '-webkit-',
-  '-xv-',
-];
+const vendorPrefix =
+  /-(ah|apple|atsc|epub|hp|khtml|moz|ms|o|rim|ro|tc|wap|webkit|xv)-/;
 
 /**
  * @param {string} selector
  * @return {string[]}
  */
-export function filterPrefixes(selector) {
-  return prefixes.filter((prefix) => selector.indexOf(prefix) !== -1);
+function filterPrefixes(selector) {
+  return selector.match(vendorPrefix);
 }
 
 // Internet Explorer use :-ms-input-placeholder.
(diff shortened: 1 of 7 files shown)


packages/postcss-merge-rules/src/lib/ensureCompatibility.js
```diff
@@ -13,7 +12,23 @@ const cssInOutOfRange = 'css-in-out-of-range';
...
+  '-moz-',
+  '-ms-',
+  '-o-',
+  '-rim-',
+  '-ro-',
+  '-tc-',
+  '-wap-',
+  '-webkit-',
+  '-xv-',
+];
```
alexander-akait, line 31: Maybe just regexp?

alexander-akait (commented):
Looks good, less deps = less problems

packages/postcss-merge-rules/src/lib/ensureCompatibility.js
ludofischer: I've replaced it with a regex. Unfortunately, I think both with and without regex the code is not particularly clear

packages/postcss-merge-rules/src/lib/ensureCompatibility.js
```diff
@@ -11,31 +11,15 @@ const cssFirstLine = 'css-first-line';
...
-  '-o-',
-  '-rim-',
-  '-ro-',
-  '-tc-',
-  '-wap-',
-  '-webkit-',
-  '-xv-',
-];
+const vendorPrefix =
+  /-(ah|apple|atsc|epub|hp|khtml|moz|ms|o|rim|ro|tc|wap|webkit|xv)-/;
```
alexander-akait, line 15: I mean `/-(.+)-/` to avoid adding new vendor (but I think we won't have new ones)

packages/postcss-merge-rules/src/lib/ensureCompatibility.js
```diff
@@ -11,8 +11,7 @@ const cssFirstLine = 'css-first-line';
 const cssInOutOfRange = 'css-in-out-of-range';
 const formValidation = 'form-validation';
 
-const vendorPrefix =
-  /-(ah|apple|atsc|epub|hp|khtml|moz|ms|o|rim|ro|tc|wap|webkit|xv)-/;
+const vendorPrefix = /-([a-z]+)-/;
```
alexander-akait, line 14: Don't forget css case insensitivity in may places, put `/i`

packages/postcss-merge-rules/src/lib/ensureCompatibility.js
ludofischer: OK, I've updated the regex. Though, in the future there also could be a string that looks like a vendor prefix but has a  different meaning :-)

alexander-akait (dismissed):

alexander-akait (approved):

ludofischer:
Using a generic regex does not work. As the code is written now we need to match

1. a vendor prefix
2. anywhere in the string (not sure, but apparently it tests against concatenated selectors?)

If we make the regex too loose it will match `glyph-icon-something` and it we only check the beginning of a string with `^` it fails to merge some rules.

PR #1226, opened 2021-11-10
fix(postcss-normalize-url): disable URL parameter sorting

### LINKED ISSUE #812: background-image

ludofischer:
Close #843
Close #812

Sorting the parameters calls URL.searchParams.sort() under the hood,
which uses the WHATWG specification for deciding which characters to encode.
The WHATWG is different from the IETF URI spec, so sorting the parameters
causes the URL to be encoded differently.
See (https://nodejs.org/docs/latest-v16.x/api/url.html#class-urlsearchparams)

Add test to check for slashes in URL parameters.

### DIFF
--- a/packages/cssnano/src/__tests__/issue315.js
+++ b/packages/cssnano/src/__tests__/issue315.js
@@ -12,7 +12,7 @@ test('should work with postcss-font-magician', () => {
     .process(css, { from: undefined })
     .then((result) => {
       expect(result.css).toMatchInlineSnapshot(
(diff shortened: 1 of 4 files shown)


ludofischer:
It boils down to a problem with the WHATWG [URLSearchParams specificaion](https://url.spec.whatwg.org/#interface-urlsearchparams), which defines a different encoding than the IETF URI specification. `normalize-url` uses the Node.js `url.searchParams.sort()` which follows the WHATWG specification, which causes the URL parameters to change encoding.

alexander-akait (approved):

alexander-akait:
Yep, legacy workarounds and stuff, better to keep it as is

PR #1227, opened 2021-11-10
chore: replace lerna with changesets

ludofischer:
For some time, Lerna would generate garbled changelogs.
Replacing it with changesets helps because changesets lets you specify
which packages to bump intead of trying to figure it out on its own.

* replace lerna prepare with yarn workspaces run
* remove publish command from workspace root else yarn tries (and fails) to publish the root
* move cli example out of packages, it's not meant to be published and it confuses yarn
* update publishing instructions


ludofischer:
1. will require reformatting the changelogs as changesets uses prettier which uses `-` for bullets instead of `*` :-)
2. at least on my machine `yarn workspaces run` seems consistently half a second slower per package :-( 
3. as you can see from the `yarn.lock` diff, this will drop a ton of `devDependencies`

alexander-akait (approved):

alexander-akait (approved):

PR #1228, opened 2021-11-11
chore: update release scripts and documentation

ludofischer:
* remind to rebuild the packages befor publishing
* add script to rebuild and publish together
* reformat CHANGELOGs with prettier

PR #1229, opened 2021-11-11
chore: remove postcss-devtools

ludofischer:
I don't think this postcss-devtools has any effect now.
According to its README, postcss-devtools is supposed to measure the running time
of individuals plugins. postcss-devtools has not been updated since PostCSS 5.
It was used in a single test but the test does not inspect the running time
and I can't see the running time in the test output either.


alexander-akait (approved):

ludofischer:
@alexander-akait can you also approve https://github.com/cssnano/cssnano/pull/1228?

PR #1230, opened 2021-11-12
refactor(postcss-merge-longhand): remove css-color-names dep

ludofischer:
css-color-names is a JSON file with a mapping from color name to hex:

1. we don't use the hex code
2. CSS4 color spec says color names are not recommended,
   so unlikely new color names will be added
3. even if we need to add a color, it is easier to add it ourselves
   than to open a PR upstream
4. css-color-names sources the color names from w3schools instead of the spec
5. css-color-names is not future-proof because it exports a JSON file


packages/postcss-merge-longhand/src/lib/colornames.js
```diff
@@ -0,0 +1,151 @@
...
+  'teal',
+  'thistle',
+  'tomato',
+  'turquoise',
+  'violet',
+  'wheat',
+  'white',
+  'whitesmoke',
+  'yellow',
+  'yellowgreen',
```
ludofischer, line 150: There are 148 colors (150 - 1 comment line - 1 export line) as expected.

alexander-akait (approved):

PR #1237, opened 2021-11-23
chore: remove unused junit reporter

ludofischer:
Not used in the current Jest config. See https://github.com/jest-community/jest-junit If it was used, it should be configured in either `jest.config.js` or at the command line.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -49,7 +49,7 @@ jobs:
         run: yarn --frozen-lockfile
 
       - name: Run tests with coverage
-        run: yarn test:coverage --ci --reporters="default" --reporters="jest-junit"
+        run: yarn test:coverage --ci --reporters="default"
 
       - name: Publish code coverage report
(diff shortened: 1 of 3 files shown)


alexander-akait (dismissed):

alexander-akait:
hm, looks like we use `jest-junit` :smile:

ludofischer:
> hm, looks like we use `jest-junit` smile

It's in the `.github` workflow. That's why ripgrep did not catch it! But is it necessary? Or just a leftover from another CI system?

alexander-akait:
Maybe we don't need it, I don't see any profit from it

ludofischer:
The jest-junit reporter could be needed for codecov. Let's see what happens.

ludofischer:
codecov should not need the Junit reporter https://github.com/codecov/example-node The upload succeeded https://github.com/cssnano/cssnano/runs/4305961921?check_suite_focus=true but there is no diff becomes the changes are in in ignored files. I think it should be good to merge.

alexander-akait (approved):

PR #1239, opened 2021-11-24
chore(css-size): drop gzip-size and brotli-size deps

ludofischer:
Using the Node.js standard library directly is as convenient.

These packages bundle a bunch of ways to call the standard library functions (ex. on a stream, with promises, sync). They set the maximum compression level by default. Then they return `length` or `byteLength` of the result buffer (`byteLength` === `length` because a Node.js `Buffer` is a  subclass of `UInt8Array` https://nodejs.org/api/buffer.html#buffer).

### DIFF
--- a/packages/css-size/src/index.js
+++ b/packages/css-size/src/index.js
@@ -1,9 +1,23 @@
+import zlib from 'zlib';
 import prettyBytes from 'pretty-bytes';
 import postcss from 'postcss';
-import { sync as gzip } from 'gzip-size';
-import { sync as brotli } from 'brotli-size';
 import Table from 'cli-table3';
 import nano from 'cssnano';
+
+/* Returns the size of the gzipped input */
+function gzip(input) {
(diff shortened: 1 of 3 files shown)


packages/css-size/src/index.js
```diff
@@ -1,10 +1,24 @@
...
 import nano from 'cssnano';
 
+/* Returns the size of the gzipped input */
+function gzip(input) {
+  return zlib.gzipSync(input, { level: 9 }).byteLength;
+}
+
+/* Returns the size of the brotli-compressed input */
+function brotli(input) {
+  return zlib.brotliCompressSync(input, {
```
ludofischer, line 14: https://github.com/erwinmombay/brotli-size/blob/6247b27b4c9c635cce5a6a781248fa0196208426/src/index.ts#L47
The original code first converts the string to a buffer, this is so it can then get the `byteSize` of the input to use as a hint, we can just call `Buffer.byteLength()` from our existing `getBinarySize()` utility.

PR #1240, opened 2021-11-24
chore: remove unused Babel environment variables

ludofischer:
I don't think `BABEL_ENV` has any effect. 
1. Our `babel.config.json` is static 
2. preset [@babel/preset-env](https://babeljs.io/docs/en/babel-preset-env) does not use it 
3. plugin [add-module-exports](https://www.npmjs.com/package/babel-plugin-add-module-exports) does not use it

For safety I have diffed a few package `dist` directory in the repo with those inside `node_modules` on a fresh cssnano install and I see no difference.


alexander-akait (approved):

packages/postcss-reduce-initial/package.json
```diff
@@ -11,7 +11,7 @@
   "scripts": {
     "acquire": "node ./src/script/acquire.mjs",
     "prebuild": "rimraf dist",
-    "build": "cross-env BABEL_ENV=publish babel src --config-file ../../babel.config.json --out-dir dist --ignore '**/__tests__/,src/script'",
+    "build": "babel src --config-file ../../babel.config.json --out-dir dist --ignore \"**/__tests__/\"",
```
ludofischer, line 14: I've moved the `script` directory out of `src`  for `postcss-reduce-initial` to make it harder to publish the `script` directory by mistake.
The `build` script for `postcss-reduce-initial` had a more complex ignore pattern, because it was the only package with the `scripts` directory  inside the `src` directory. On Windows, removing cross-env probably changed the way some globs were interpreted, so the `publish` script placed a copy of the `script` directory inside `dist`. Jest tried to run the tests inside `dist/scripts/__tests__/ `, but the imports then did not work, causing the build to fail.

ludofischer:
This breaks some scripts as it is, need to redo.

PR #1242, opened 2021-11-25
fix(postcss-reduce-initial): regenerate data

ludofischer:
* Extract script out of published `src` directory
* Run script to update initial value data from MDN

The `script` directory does not need to be published so it's better to move it out of `src` as in `postcss-ordered-values`. This also makes the `build` script more resistant to using a different command line interpreter, as the `ignore` expression is simpler (it previously failed on windows without `cross-env`). To test everything still works I've re-run the script to update the initial values.


packages/postcss-reduce-initial/data/toInitial.json
```diff
@@ -17,17 +17,19 @@
   "box-sizing": "content-box",
   "column-rule-color": "currentcolor",
   "font-synthesis": "weight style",
+  "image-orientation": "from-image",
   "mask-clip": "border-box",
   "mask-mode": "match-source",
   "mask-origin": "border-box",
-  "mask-repeat": "repeat",
```
ludofischer, line 23: The [spec](https://drafts.fxtf.org/css-masking/#the-mask-repeat) and [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/mask-repeat) do not agree on this one. Could be browser-dependent? A familiar person raised [an issue](https://github.com/mdn/data/issues/351) but the spec changed in between...

packages/postcss-reduce-initial/data/toInitial.json
```diff
@@ -17,17 +17,19 @@
...
+  "mask-repeat": "no-repeat",
   "mask-type": "luminance",
   "ruby-align": "space-around",
   "ruby-merge": "separate",
+  "ruby-position": "alternate",
   "text-decoration-color": "currentcolor",
   "text-emphasis-color": "currentcolor",
   "text-emphasis-position": "over right",
-  "transform-box": "border-box",
```
ludofischer, line 30: Correct per [spec](https://drafts.csswg.org/css-transforms/#transform-box)

packages/postcss-reduce-initial/data/fromInitial.json
```diff
@@ -209,9 +226,9 @@
   "right": "auto",
   "rotate": "none",
   "row-gap": "normal",
-  "ruby-position": "over",
```
ludofischer, line 212: Correct per [spec](https://drafts.csswg.org/css-ruby/#rubypos)

packages/postcss-reduce-initial/data/toInitial.json
```diff
@@ -17,17 +17,19 @@
...
+  "transform-box": "view-box",
```
alexander-akait, line 32: It can be problem... I remember some browsers have wrong `mask-repeat` and `transform-box` values, so we remove them, but we should left a comment... maybe make sense too look at git history to check it, a lot of time has passed

ludofischer: mask-repeat has already been manually changed to `repeat` in the past https://github.com/cssnano/cssnano/commit/cebd6d5b789a9c700f92dc8b40cdd8ef9545441a#diff-a3304505493004c674a19e13da735a9ea783c7f76e93c67ad0bb14334a67225e So I think we should probably follow the spec and not Mozilla. Maybe add a test like https://github.com/cssnano/cssnano/pull/852/files to avoid repeating the error?

alexander-akait: Yes, let's add test and comments in tests

packages/postcss-reduce-initial/data/toInitial.json
```diff
@@ -21,7 +21,6 @@
   "mask-clip": "border-box",
   "mask-mode": "match-source",
   "mask-origin": "border-box",
-  "mask-repeat": "no-repeat",
```
ludofischer: I think the previous fix for this was incorrect because if you change `mask-repeat` to `repeat` then you also need to move it to `fromInitial` since now the value is shorter than `initial`.

packages/postcss-reduce-initial/data/toInitial.json
ludofischer: Value is `repeat` and it goes into `fromInitial`.

alexander-akait:
@ludofischer Please submit invoice on 120$ https://opencollective.com/cssnano

Regarding to this PR looks good, we can fix it if somebody reported about problem, it is fast and easy

alexander-akait (approved):

PR #1245, opened 2021-11-26
test: replace chalk with picocolors

ludofischer:
Since latest chalk is Node 12.20+ only


alexander-akait (approved):
Yep, time to remove chalk

PR #1248, opened 2021-11-27
test: switch to uvu

ludofischer:
* adapt snapshot string escapes
* replace Jest mocks with nanospy
* switch to c8 for coverage

Curious to see what happens on CI.
I was a bit disappointed as it runs only 5 seconds faster locally. Could be a lot of time is spent inside the Babel transform, but since we publish the transformed files I think it is better to apply a very similar transform on tests.


alexander-akait (commented):
Looks good, changing the testing tool is certainly a big change, but I don't against.

I think we can do more:
- drop babel, we don't use very new syntax and have tests and CI, so we can rewrite `import` or `require` (and use `import` for the next major release), I don't see any profit from babel for us, babel just rewrites `import` on `require`, and no more, I don't think we need so many big and slow tool to do this
- if we will add types, we can use ts jsdocs (https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html), it is easy maintenance and we don't need ts or babel or something else to transpile our code

ludofischer:
On Windows it's only running the tests inside `cssnano_utils`. Likely something wrong with interpretation of paths, although we don't change the default test to recognize test files.

alexander-akait:
Yes, it is strange...

alexander-akait:
Maybe try to get paths in `""`, I faced with this sometimes

ludofischer:
>     * drop babel, we don't use very new syntax and have tests and CI, so we can rewrite `import` or `require` (and use `import` for the next major release), I don't see any profit from babel for us, babel just rewrites `import` on `require`, and no more, I don't think we need so many big and slow tool to do this
> 
>     * if we will add types, we can use ts jsdocs (https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html), it is easy maintenance and we don't need ts or babel or something else to transpile our code

I also thought about changing import to require. Our `import` is 'fake' anyway, many files do not work when you run them in Node directly (JSON imports, bad import paths and so on). Then I would first drop Babel without also changing testing tool. But maybe first make a release with the few updates we have now before we break every package at once :-)

alexander-akait:
Make sense, feel free to do it :+1:

ludofischer:
Fixed Windows CI. You need to use double quotes instead of single quotes (at least on GitHub CI, that's why some of the packages were not building correctly either before)!

alexander-akait (approved):

PR #1254, opened 2021-11-30
chore: update remaining dev dependencies

ludofischer:
Pin eslint-plugin-import at 2.24.2 because >=2.25 bug.


site/yarn.lock
```diff
@@ -1771,6 +1771,13 @@
   dependencies:
     "@types/node" "*"
 
+"@types/cssnano@^4.0.1":
+  version "4.0.1"
+  resolved "https://registry.yarnpkg.com/@types/cssnano/-/cssnano-4.0.1.tgz#67fa912753d80973a016e7684a47fedf338aacff"
+  integrity sha512-hGOroxRTBkYl5gSBRJOffhV4+io+Y2bFX1VP7LgKEVHJt/LPPJaWUIuDAz74Vlp7l7hCDZfaDi7iPxwNwuVA4Q==
+  dependencies:
+    postcss "5 - 7"
```
alexander-akait, line 1779: Hm, interesting, I think we really need types here, as minimum for public functions

alexander-akait (approved):

site/yarn.lock
ludofischer: Does it mean we need to add TypeScript with TSDoc comments?
I've also wondered why that package appeared. Seems like `css-minimizer-webpack-plugin` is pulling it in.
=> Found "@types/cssnano@4.0.1"
info Reasons this module exists
   - "@docusaurus#core#css-minimizer-webpack-plugin" depends on it
   - Hoisted from "@docusaurus#core#css-minimizer-webpack-plugin#@types#cssnano"

alexander-akait: > Does it mean we need to add TypeScript with TSDoc comments?

Yes, it will be enough

PR #1255, opened 2021-11-30
chore(postcss-normalize-url): inline is-absolute-url dep

ludofischer:
is-absolute-url has become ESM only. It boils down to two
short regexes, so we can inline it.
The license is MIT which is the same as cssnano.

### DIFF
--- a/packages/postcss-normalize-url/src/index.js
+++ b/packages/postcss-normalize-url/src/index.js
@@ -1,11 +1,28 @@
 import path from 'path';
 import valueParser from 'postcss-value-parser';
 import normalize from 'normalize-url';
-import isAbsolute from 'is-absolute-url';
 
 const multiline = /\\[\r\n]/;
 // eslint-disable-next-line no-useless-escape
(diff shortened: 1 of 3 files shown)


packages/postcss-normalize-url/src/index.js
```diff
@@ -1,12 +1,29 @@
...
 import normalize from 'normalize-url';
-import isAbsolute from 'is-absolute-url';
 
 const multiline = /\\[\r\n]/;
 // eslint-disable-next-line no-useless-escape
 const escapeChars = /([\s\(\)"'])/g;
 
+// Scheme: https://tools.ietf.org/html/rfc3986#section-3.1
+// Absolute URL: https://tools.ietf.org/html/rfc3986#section-4.3
+const ABSOLUTE_URL_REGEX = /^[a-zA-Z][a-zA-Z\d+\-.]*?:/;
```
ludofischer, line 11: The regexes are simple and I could not see any vulnerabilities reports or significant changes in the last two years.

alexander-akait (approved):

Issue #1262, opened 2021-12-13
[Feature Request]: Allow colormin to take full advantage of colord options


Irevall:
### What should be improved?

Colormin in current state is [just a wrapper](https://github.com/cssnano/cssnano/blob/f93026181d2cf707f308016ffbecab1ceb8aa1d2/packages/postcss-colormin/src/minifyColor.js#L25) for [colord](https://github.com/omgovich/colord) package, but for some reason allows to take control of only 2 out of [6 options](https://github.com/omgovich/colord#color-utilities).
For example for my project I'd like to turn off the `name` transformation, e.g. `#ff0000 -> red`, but I can only change `alphaHex` and `transparent` settings.

### Describe the solution you would like

Instead of passing
```
{
  alphaHex: settings.supportsAlphaHex,
  transparent: settings.supportsTransparent,
  name: true,
}
```
We should pass entire `settings` object with added backwards compatibility  for `supportsAlphaHex` and `supportsTransparent`.

### Are you willing to work on this?

- [X] Yes, I would like to help

Irevall:
Whoops turns out I misread a bit and there is no options support whatsoever. 😅
Added PR.

ludofischer:
Fixed in #1263

PR #1263, opened 2021-12-14
feat: allow options in colormin

Irevall:
Allow passing `colord` lib options to `postcss-colormin`.
Created from #1262

### DIFF
--- a/packages/postcss-colormin/src/index.js
+++ b/packages/postcss-colormin/src/index.js
@@ -66,24 +66,29 @@ function transform(value, options) {
   return parsed.toString();
 }
 
-function pluginCreator() {
+function addPluginDefaults(options, browsers) {
+  const defaults = {
+    transparent: browsers.some(hasTransparentBug) === false, // Does the browser support 4 & 8 character hex notation
+    alphaHex: isSupported('css-rrggbbaa', browsers), // Does the browser support "transparent" value properly
+    name: true,
+  };
+  return { ...defaults, ...options };
+}
+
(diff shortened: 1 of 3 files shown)


packages/postcss-colormin/src/index.js
```diff
@@ -66,24 +66,29 @@ function transform(value, options) {
   return parsed.toString();
 }
 
-function pluginCreator() {
+function addPluginDefaults(options, browsers) {
+  const defaults = {
+    transparent: browsers.some(hasTransparentBug) === false, // Does the browser support 4 & 8 character hex notation
```
ludofischer, line 71: Nice to check for feature support!

ludofischer (approved):
I think overall it is an improvement. Should even fix minification when the browsers do not support some features.

packages/postcss-colormin/src/index.js
Irevall: This was already in the code, I just moved it around. 🙈

Irevall:
I guess I should also add this new information to ReadMe... anywhere else?

ludofischer:
> I guess I should also add this new information to ReadMe... anywhere else?

I think it's enough to add it to the ReadMe. I'll figure out if other docs need to be updated, I think they are automatically generated though.

packages/postcss-colormin/src/index.js
ludofischer: Ah, true, but it still a bugfix because previously the options did not get passed down to colord.

PR #1265, opened 2021-12-14
chore: switch from yarn to pnpm

ludofischer:


ludofischer:
cssnano preset resolution might be really broken with alternative package managers...

PR #1266, opened 2021-12-14
test: declare all test dependencies in devDependencies

ludofischer:
Add missing devDependencies or tests will fail with stricter package managers.


packages/css-size/src/__tests__/index.js
```diff
@@ -3,7 +3,6 @@ import { spawn } from 'child_process';
 import path from 'path';
 import * as assert from 'uvu/assert';
 import { test } from 'uvu';
-import colors from 'colors/safe';
```
ludofischer, line 6: Look like a `chalk` ancestor that  somehow still ends up on the module resolution path, better to drop it completely rather than add a new dependency.

alexander-akait (approved):

PR #1269, opened 2021-12-17
Pnpm

ludofischer:

PR #1270, opened 2021-12-17
fix(cssnano): replace is-resolvable dep

ludofischer:
is-resolvable cannot find modules when installed in a pnpm monorepo . It's a tiny [wrapper around require.resolve](https://github.com/shinnn/is-resolvable),
so we can drop it easily. This solves the issue when running cssnano inside a pnpm monorepo (see [before](https://github.com/cssnano/cssnano/actions/runs/1579937249) and [after](https://github.com/cssnano/cssnano/actions/runs/1593701210)).

### DIFF
--- a/packages/cssnano/package.json
+++ b/packages/cssnano/package.json
@@ -25,7 +25,6 @@
     "license": "MIT",
     "dependencies": {
         "cssnano-preset-default": "^5.1.9",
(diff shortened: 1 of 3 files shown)


alexander-akait (approved):

ludofischer:
Already done in https://github.com/cssnano/cssnano/pull/1269

PR #1279, opened 2022-01-06
refactor: remove cssnano-utils getMatch

ludofischer:
Replace the `getMatch` function from cssnano-utils. `getMatch()` takes an array and returns a corresponding string. Replace it with a map that uses stringified arrays as keys. Also refactor some tests so that the fixtures do not derive from the main source code.

This is a breaking change for the `cssnano-utils` package.
 
### Motivation
The data structure `[[string, string[]]]` makes the intent unclear.  For example, the keys are the second values in every nested array of `[string, string[]]`. 

### Solution
Replace `getMatch()` with a standard `Map()`.
`getMatch()` is always used to return a shorthand for a sequence of values, so the key is always 1 level deep. Therefore comparing the arrays is the same as comparing the result of calling `toString()` on the arrays. 
Even when the array elements are numbers, looking up based on the `toString()` return value works because `toString()` normalizes the numbers, for example `2.3e-1` becomes `'0.23'`, so comparing the strings works the same as comparing the numbers directly, at least for the kind of numbers in a CSS transform.


packages/postcss-normalize-display-values/src/__tests__/index.js
```diff
@@ -1,26 +1,40 @@
...
 test('should pass through "block ruby"', passthroughCSS('display:block ruby;'));
 
 test('should pass through single values', passthroughCSS('display:block;'));
-
-Object.keys(data).forEach((key) => {
-  const actual = data[key];
-  const expected = key;
-
+/* source: https://www.w3.org/TR/css-display-3/#the-display-properties */
+[
```
ludofischer, line 14: Use a hard coded map because if the test uses the same map as the main code, the test cannot detect mistakes in the map from the main code (for example a typo).

packages/postcss-normalize-repeat-style/src/__tests__/index.js
```diff
@@ -38,10 +35,17 @@ function suite(fixture, expected) {
     ]);
 }
 
-Object.keys(data).forEach((conversion) => {
-  const fixture = data[conversion];
-
-  test(fixture, suite(fixture, conversion));
+[
```
ludofischer, line 38: Same as for the other test, this test could pass even if the mapping in the main code was wrong.

packages/cssnano-utils/src/__tests__/getMatch.test.js
```diff
@@ -1,17 +0,0 @@
-import { test } from 'uvu';
-import * as assert from 'uvu/assert';
-import getMatchFactory from '../getMatch';
-
-test('should get match', () => {
```
ludofischer, line 5: You can see from this test that this code was meant to work like a custom map implementation.

packages/cssnano-utils/src/getMatch.js
```diff
@@ -1,13 +0,0 @@
-export default function getMatchFactory(map) {
-  return function getMatch(args) {
```
ludofischer, line 2: I find this function a bit hard to parse.

alexander-akait (approved):

PR #1285, opened 2022-01-15
chore: speed up build by building packages in parallel

ludofischer:


ludofischer:
Does not seem to be particularly faster on CI overall, but it does run a lot faster on my laptop...

alexander-akait (approved):

PR #1287, opened 2022-01-15
refactor(postcss-discard-comments): use map instead of object as cache

ludofischer:
Help TypeScript infer the right type for the cache.

### DIFF
--- a/packages/postcss-discard-comments/src/index.js
+++ b/packages/postcss-discard-comments/src/index.js
@@ -3,17 +3,17 @@ import commentParser from './lib/commentParser';
 
 function pluginCreator(opts = {}) {
   const remover = new CommentRemover(opts);
-  const matcherCache = {};
-  const replacerCache = {};
+  const matcherCache = new Map();
+  const replacerCache = new Map();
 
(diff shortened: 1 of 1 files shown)


packages/postcss-discard-comments/src/index.js
```diff
@@ -41,7 +41,7 @@ function pluginCreator(opts = {}) {
 
     const result = space(parsed).join(' ');
 
-    replacerCache[key] = result;
+    replacerCache.set(key, result);
 
     return result;
   }
```
alexander-akait, line 47: Ideally we can use WeakMap here and store AST nodes, so it reduces memory usage,  but let's think about it late

ludofischer: I don't think storing the AST nodes in a WeakMap would be equivalent, because right now it is returns a cached result for different nodes with the same structure, while if we use the node as a key it will only return a cached result if it visits the same node again. In fact since the plugin uses `OnceExit` it should not see the same node twice, so using the AST node as a cache key would never return a cache hit.

alexander-akait (approved):

PR #1293, opened 2022-01-19
refactor: replace objects with maps and avoid excessively dynamic code

ludofischer:
* use maps instead of objects when the object is only used as a map
* make function dispatching a little more explicit.
* test: add one test for uncovered case in background positions

### DIFF
--- a/packages/postcss-normalize-positions/src/index.js
+++ b/packages/postcss-normalize-positions/src/index.js
@@ -3,8 +3,14 @@ import valueParser, { unit } from 'postcss-value-parser';
 const directionKeywords = ['top', 'right', 'bottom', 'left', 'center'];
 
 const center = '50%';
-const horizontal = { right: '100%', left: '0' };
-const verticalValue = { bottom: '100%', top: '0' };
+const horizontal = new Map([
+  ['right', '100%'],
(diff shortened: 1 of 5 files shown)


alexander-akait (approved):

packages/postcss-normalize-string/src/index.js
```diff
@@ -248,22 +252,16 @@ function pluginCreator(opts) {
       const cache = new Map();
 
       css.walk((node) => {
```
ludofischer, line 254: The code picks a property name based on the node type, then uses the property name to select a function and the property to update on the node. I find using a map is not necessary since the node types are fixed

packages/postcss-normalize-positions/src/index.js
```diff
@@ -140,12 +146,10 @@ function transform(value) {
         nodes[2].value = nodes[1].value = '';
       }
 
-      const map = Object.assign({}, horizontal, {
-        center,
-      });
+      const map = new Map([...horizontal, ['center', center]]);
```
ludofischer, line 149: The test I have added should add coverage for this line. Previously, all tests passed even when I had made a mistake here.

PR #1297, opened 2022-01-21
refactor: replace alphanum-sort with built-in sort

ludofischer:
The order inside selector lists and atrule parameters has no meaning. The reason to sort them is to improve compression. But then the sorting method does not matter as long as it is the same everywhere.

The main changes is that uppercase now comes before lowercase, and numbers are ordered as letters. Also letters come before other symbols like `*` (of course the order inside a single selector still does not change).

Passing this collator to `sort()` gives almost the same result as `alphanum-sort`:

```javascript
new Intl.Collator('en', {sensitivity: 'base', numeric: true, ignorePunctuation: true}).compare
```
But why use a more complicated sort method when the order does not matter anyway?


packages/postcss-minify-selectors/src/__tests__/index.js
```diff
@@ -96,10 +96,10 @@ test(
 );
 
 test(
-  'should sort using natural sort',
```
alexander-akait, line 99: I think natural sort has better compression... I think we should check it

ludofischer: I am getting the opposite result, natural sort is slightly worse. I am checking with:
```
gzip -6 -c packages/cssnano-preset-default/src/__tests__/integrations/*css | wc -c
```
I get 32411 bytes for natural sort and 32407 bytes for the built-in sort.

ludofischer: Gzipping all files to disk gives the same result for both: 356K

alexander-akait (approved):

TrySound:
😢

PR #1299, opened 2022-01-21
Publish cssnano 5.0.16

ludofischer:


ludofischer:
I had forgotten to update the lockfile after bumping the package versions. Should be fine now.

alexander-akait (approved):

PR #1303, opened 2022-01-24
refactor(postcss-convert-values): remove unnecessary steps

ludofischer:
- lower case unit once
- if unit is one no need to multiply or divide

### DIFF
--- a/packages/postcss-convert-values/src/lib/convert.js
+++ b/packages/postcss-convert-values/src/lib/convert.js
@@ -1,19 +1,19 @@
-const lengthConv = {
-  in: 96,
-  px: 1,
-  pt: 4 / 3,
-  pc: 16,
-};
+const lengthConv = new Map([
+  ['in', 96],
+  ['px', 1],
+  ['pt', 4 / 3],
+  ['pc', 16],
+]);
 
-const timeConv = {
-  s: 1000,
-  ms: 1,
-};
+const timeConv = new Map([
+  ['s', 1000],
+  ['ms', 1],
+]);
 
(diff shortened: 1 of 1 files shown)


packages/postcss-convert-values/src/lib/convert.js
```diff
@@ -31,41 +31,32 @@ function dropLeadingZero(number) {
...
-    if (conversion[u] === 1) {
-      one = u;
-    }
-    return lowerCasedUnit !== u;
+function transform(number, originalUnit, conversions) {
+  let conversionUnits = [...conversions.keys()].filter((u) => {
+    return originalUnit !== u;
   });
 
-  if (lowerCasedUnit === one) {
```
ludofischer, line 44: I don't think this `one` variable has any purpose any more.
if `lowerCasedUnit === one` then  `conversion[lowerCasedUnit] === 1` since `conversion[one] === 1`.
So `base / conversion[lowerCasedUnit] === base * conversion[lowerCasedUnit]`, correct?

packages/postcss-convert-values/src/lib/convert.js
```diff
@@ -31,41 +31,32 @@ function dropLeadingZero(number) {
   return value;
 }
 
-function transform(number, unit, conversion) {
-  const lowerCasedUnit = unit.toLowerCase();
```
ludofischer, line 35: We already converted the unit to lower case to know whether to call this function, so we can pass the lower case unit directly instead of converting it once more.

alexander-akait (approved):

PR #1304, opened 2022-01-24
fix: fix bad parameters and conversions

ludofischer:
- `postcssValueParser.unit()` [returns the number as a string](https://github.com/TrySound/postcss-value-parser/blob/11a4436e74930f4eaf9cbdb2063656bcce8ebf6a/lib/unit.js#L117), so checking against `0` will always fail

- It's `rawSpaceAfter` with just one `s`. It seems to make no difference in the tests though
- in postcss-zindex, `LayerCache` constructor does not take parameters

### DIFF
--- a/packages/postcss-zindex/src/index.js
+++ b/packages/postcss-zindex/src/index.js
@@ -4,7 +4,7 @@ function pluginCreator(opts = {}) {
   return {
     postcssPlugin: 'postcss-zindex',
     prepare() {
-      const cache = new LayerCache(opts);
+      const cache = new LayerCache();
       return {
         OnceExit(css) {
           const nodes = [];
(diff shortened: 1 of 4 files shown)


packages/postcss-zindex/src/index.js
```diff
@@ -4,7 +4,7 @@ function pluginCreator(opts = {}) {
   return {
     postcssPlugin: 'postcss-zindex',
     prepare() {
-      const cache = new LayerCache(opts);
+      const cache = new LayerCache();
```
ludofischer, line 7: opts is not used in the constructor, it's only a parameter for the `optimizeValues` method used below.

alexander-akait (approved):

PR #1305, opened 2022-01-25
fix(postcss-minify-gradients): handle node values as strings

ludofischer:
Explicitly convert numbers to strings before assigning them to a
node value, so it's clearer how to compare them.

### DIFF
--- a/packages/postcss-minify-gradients/src/index.js
+++ b/packages/postcss-minify-gradients/src/index.js
@@ -83,7 +83,7 @@ function optimise(decl) {
           }
 
           if (lastStop && thisStop && isLessThan(lastStop, thisStop)) {
-            arg[2].value = 0;
+            arg[2].value = '0';
           }
 
           lastStop = thisStop;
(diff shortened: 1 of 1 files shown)


alexander-akait (approved):

packages/postcss-minify-gradients/src/index.js
```diff
@@ -162,7 +162,7 @@ function optimise(decl) {
           color = color.toLowerCase();
 
           const colorStop =
-            stop || stop === 0
```
ludofischer, line 165: `stop` is either `undefined` or contains a node value which in general is a string, so it should never be the number `0`, unless someone assigned the number `0` to the node value by mistake. Anyway, I think the intent was to check whether `stop` had been initialized, so I have made it explicit.

packages/postcss-minify-gradients/src/index.js
```diff
@@ -83,7 +83,7 @@ function optimise(decl) {
           }
 
           if (lastStop && thisStop && isLessThan(lastStop, thisStop)) {
-            arg[2].value = 0;
```
ludofischer, line 86: This works because the nodes get all serialized anyway, but the node values can't really be literal numbers.

PR #1308, opened 2022-01-25
refactor(stylehacks): use standard JavaScript subclasses

ludofischer:
- replace the mix of closures and `.apply()` with `extends`
- replace rest parameters with single arguments because
  plugin methods are only called with one node as argument

### DIFF
--- a/packages/stylehacks/src/plugin.js
+++ b/packages/stylehacks/src/plugin.js
@@ -1,59 +1,59 @@
-export default function plugin(targets, nodeTypes, detect) {
-  class Plugin {
-    constructor(result) {
-      this.nodes = [];
-      this.result = result;
-      this.targets = targets;
-      this.nodeTypes = nodeTypes;
-    }
-
-    push(node, metadata) {
-      node._stylehacks = Object.assign({}, metadata, {
-        message: `Bad ${metadata.identifier}: ${metadata.hack}`,
-        browsers: this.targets,
-      });
-
-      this.nodes.push(node);
-    }
-
-    any(node) {
-      if (this.nodeTypes.includes(node.type)) {
-        detect.apply(this, arguments);
-
-        return !!node._stylehacks;
-      }
-
-      return false;
-    }
-
-    detectAndResolve(...args) {
-      this.nodes = [];
-
-      detect.apply(this, args);
-
-      return this.resolve();
-    }
-
-    detectAndWarn(...args) {
-      this.nodes = [];
-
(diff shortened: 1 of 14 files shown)


packages/stylehacks/src/plugin.js
```diff
@@ -1,59 +1,53 @@
-export default function plugin(targets, nodeTypes, detect) {
```
ludofischer, line 1: I think this looks like a custom implementation of subclassing. The function returns creates a closure to add custom versions of `targets`, `nodeTypes` and `detect` to the `Plugin` class.

packages/stylehacks/src/plugins/bodyEmpty.js
```diff
@@ -1,33 +1,41 @@
...
 import exists from '../exists';
 import isMixin from '../isMixin';
-import plugin from '../plugin';
+import Plugin from '../plugin';
 import { FF_2 } from '../dictionary/browsers';
 import { SELECTOR } from '../dictionary/identifiers';
 import { RULE } from '../dictionary/postcss';
 import { BODY } from '../dictionary/tags';
 
-function analyse(ctx, rule) {
```
ludofischer, line 10: This is going to receive `this` as the `ctx` argument when called inside the class, so it is the same as an instance method.

packages/stylehacks/src/plugin.js
```diff
@@ -1,59 +1,53 @@
...
-      }
+  any(node) {
+    if (this.nodeTypes.includes(node.type)) {
+      this.detect(node);
 
-      return false;
+      return !!node._stylehacks;
     }
 
-    detectAndResolve(...args) {
```
ludofischer, line 29: This function only work with a single `Node`. I suspect rest parameters were used to obtain an array to pass to `apply`.

packages/stylehacks/src/plugin.js
```diff
@@ -1,59 +1,59 @@
...
-      this.nodeTypes = nodeTypes;
-    }
-
-    push(node, metadata) {
-      node._stylehacks = Object.assign({}, metadata, {
-        message: `Bad ${metadata.identifier}: ${metadata.hack}`,
-        browsers: this.targets,
-      });
+export default class BasePlugin {
+  constructor(targets, nodeTypes, result) {
```
ludofischer, line 2: We need to place `result` last because sometimes plugins are constructed without passing one, so it could be undefined.

alexander-akait (approved):

PR #1309, opened 2022-01-26
refactor: conform to type signatures

ludofischer:
- selector parser callback is not expected to return a node
 - unit() returns false if it cannot parse
 - don't assign an array of nodes to a string

### DIFF
--- a/packages/postcss-merge-longhand/src/lib/decl/columns.js
+++ b/packages/postcss-merge-longhand/src/lib/decl/columns.js
@@ -60,10 +60,10 @@ function explode(rule) {
 
     values.forEach((value, i) => {
       let prop = properties[1];
-
+      const dimension = unit(value);
       if (value.toLowerCase() === auto) {
         prop = properties[i];
-      } else if (unit(value).unit) {
+      } else if (dimension && dimension.unit !== '') {
         prop = properties[0];
       }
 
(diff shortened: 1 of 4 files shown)


packages/postcss-merge-longhand/src/lib/decl/columns.js
```diff
@@ -60,10 +60,10 @@ function explode(rule) {
 
     values.forEach((value, i) => {
       let prop = properties[1];
-
+      const dimension = unit(value);
       if (value.toLowerCase() === auto) {
         prop = properties[i];
-      } else if (unit(value).unit) {
+      } else if (dimension && dimension.unit !== '') {
```
ludofischer, line 66: I don't know whether the original code skipped this branch on purpose if `unit === ''` or if it was just an oversight. I have written the check like this to preserve the same output.

packages/postcss-minify-gradients/src/isColorStop.js
```diff
@@ -29,9 +29,8 @@ function isCSSLengthUnit(input) {
 }
 
 function isStop(str) {
-  let stop = !str;
```
ludofischer, line 32: - If `str` is falsy, then `stop` is `true`, so the next branch is skipped and the function returns `true`
- If `str` is truthy then `stop` is `false`, so the function returns the value assigned to `stop` in the `if` branch.

packages/postcss-unique-selectors/src/index.js
```diff
@@ -23,7 +23,7 @@ function pluginCreator() {
               sel.remove();
               return;
             } else {
-              return sel;
+              return;
```
ludofischer, line 26: selector parser callback is not expected to return a node.

PR #1312, opened 2022-01-27
docs: downgrade css-declaration-sorter in playground

### LINKED ISSUE #1311: [Bug]: https://cssnano.co/playground/ fails to minify input
### Describe the bug

Default and Advanced presets:
image

### Expected behaviour

Playground minifies input successfully

### Steps to reproduce

1. Open https://cssnano.co/playground/
2. Press "Play" button

### Version

5.0.16?

### Preset

default

### Environment

```shell
Google Chrome 97.0.4692.71
```

### Package details

```shell
Whatever playground uses
```

### Additional context

_No response_

ludofischer:
Fix #1311

Somehow the last patch release does not run when bundled with webpack.

cc @Siilwyn

### DIFF
--- a/site/package.json
+++ b/site/package.json
@@ -25,6 +25,9 @@
     "react": "^17.0.2",
     "react-dom": "^17.0.2"
   },
+  "pnpm": {
+    "overrides": {"css-declaration-sorter": "6.1.3" }
+  },
   "devDependencies": {
     "mdast-util-heading-range": "^3.1.0",
     "remark": "^14.0.2",
--- a/site/pnpm-lock.yaml
+++ b/site/pnpm-lock.yaml
@@ -1,5 +1,8 @@
 lockfileVersion: 5.3
 
+overrides:
+  css-declaration-sorter: 6.1.3
+
 importers:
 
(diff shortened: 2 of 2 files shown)


alexander-akait (approved):

Siilwyn:
Oof sorry about this, looks like a odd regression in the way rollup transpiles the code?
The only difference I can find in the CommonJS build output is:

6.1.3: (rollup 2.56.3)
```
    return Promise.resolve().then(function () { return /*#__PURE__*/_interopNamespace(require(`../orders/${order}.cjs`)); 
```
6.1.4: (rollup 2.59.0)
```
    return (function (t) { return Promise.resolve().then(function () { return /*#__PURE__*/_interopNamespace(require(t)); }); })(`../orders/${order}.cjs`)
```

Guessing webpack doesn't properly transform the latter since the path is in a variable now?

It might be worth looking into switching to using the ESM exports, which uses a dynamic import:
```
return import(`../orders/${order}.mjs`)
```

Shall I downgrade rollup and publish a new patch version (6.1.5)?

PR #1315, opened 2022-02-05
chore: convert source to CommonJS

ludofischer:
Get rid of Babel as imagined in https://github.com/cssnano/cssnano/pull/1248#pullrequestreview-817134292
Same as https://github.com/postcss/postcss-calc/pull/168 but for cssnano

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,11 +1,9 @@
 {
   "parserOptions": {
-    "sourceType": "module",
     "ecmaVersion": 2021
   },
   "extends": [
     "eslint:recommended",
-    "plugin:import/recommended",
     "plugin:prettier/recommended"
   ],
   "env": {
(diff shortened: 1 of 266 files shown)


.eslintrc.json
```diff
@@ -33,15 +31,14 @@
...
-    "import/no-duplicates": "error",
-    "import/no-mutable-exports": "error",
-    "import/no-named-as-default": "error",
-    "import/no-named-as-default-member": "error",
-    "import/order": "error",
-    "import/prefer-default-export": "error"
-  }
+    "yoda": "error"
+  },
+  "overrides": [
```
ludofischer, line 36: There's some helpers file (to build the site, fetch data from MDN) that are real ESM, so we need to handle them differently.

.github/workflows/test.yml
```diff
@@ -28,6 +28,26 @@ jobs:
       - name: Run lint
         run: pnpm lint
 
+  test_helpers:
```
ludofischer, line 31: Need an extra step because we cannot run the tests for the files that are real ESM in all node versions.

alexander-akait (approved):

PR #1320, opened 2022-02-06
fix: remove package.json references to broken or deleted files

ludofischer:
- fix(cssnano): remove broken RunKit quickstart
The code in quickstart.js throws an error. I think
it was meant to be used with the service that today
has become RunKit, but it's useless in its current state.

- chore(postcss-discard-unused): remove non-existing files from package.json
- chore: remove empty scripts sections


packages/cssnano/package.json
```diff
@@ -31,10 +31,8 @@
     "repository": "cssnano/cssnano",
     "files": [
         "src",
-        "LICENSE-MIT",
-        "quickstart.js"
+        "LICENSE-MIT"
     ],
-    "tonicExampleFilename": "quickstart.js",
```
ludofischer, line 37: This is the file that appears on `npm.runkit.com/cssnano` when you click `Try on RunKit` on the npm website. But it just shows an error message. It seems RunKit used to be called tonic.dev, hence the field name.

alexander-akait (approved):

PR #1325, opened 2022-02-08
chore: lint code to check old Node.js compatibility

ludofischer:
Since we're still targeting Node.js 10, set the language level
to a compatible version, except for modules we don not publish.
Also remove the Jest environment.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,6 +1,6 @@
 {
   "parserOptions": {
-    "ecmaVersion": 2021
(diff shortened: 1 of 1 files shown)


alexander-akait (approved):

Siilwyn:
👋 since Node.js 12 is almost at it's EOL and I see modules switch to ESM only it got me wondering why cssnano is still targeting unsupported (security-wise) versions?

PR #1333, opened 2022-02-17
More types

ludofischer:
Another batch of TypeScript conversions, I am trying to limit the amount of files changes in a single pull request to make it possible to review.


packages/postcss-convert-values/src/lib/convert.js
```diff
@@ -31,19 +34,33 @@ function dropLeadingZero(number) {
...
+ * @param {Map<string, number>} conversions
+ * @return {string}
+ */
 function transform(number, originalUnit, conversions) {
   let conversionUnits = [...conversions.keys()].filter((u) => {
     return originalUnit !== u;
   });
 
-  const base = number * conversions.get(originalUnit);
+  const base = number * /** @type {number} */ (conversions.get(originalUnit));
```
ludofischer, line 48: TypeScript could have caught a legitimate issue here. This would return `NaN` if the unit was unknown.

packages/postcss-reduce-initial/src/index.js
```diff
@@ -1,18 +1,22 @@
 'use strict';
 const browserslist = require('browserslist');
 const { isSupported } = require('caniuse-api');
-const fromInitial = require('../data/fromInitial.json');
-const toInitial = require('../data/toInitial.json');
+const fromInitial = require('./data/fromInitial.json');
```
ludofischer, line 4: Need to move the data files because TypeScript needs each subproject in a composite build to be inside a single root.

packages/postcss-convert-values/src/index.js
```diff
@@ -89,20 +102,31 @@ function clampOpacity(node) {
...
   return (
     (decl.value.includes('%') &&
       (lowerCasedProp === 'max-height' || lowerCasedProp === 'height')) ||
-    (parent.parent &&
-      parent.parent.name &&
-      parent.parent.name.toLowerCase() === 'keyframes' &&
+    (parent &&
+      parent.parent &&
+      parent.parent.type === 'atrule' &&
+      /** @type {import('postcss').AtRule} */ (
```
ludofischer, line 118: I've had to check whether `parent` is defined first and I've replaced the check for `parent.name` with `type == 'atrule'`, else TypeScript won't narrow the type to `AtRule`.

packages/postcss-minify-gradients/src/index.js
```diff
@@ -55,9 +63,14 @@ function optimise(decl) {
...
+          node.nodes[0].value =
+            angles[
+              /** @type {'top'|'right'|'bottom'|'left'}*/ (
+                node.nodes[0].value.toLowerCase()
+              )
+            ];
         }
-
+        /** @type {valueParser.Dimension | null | false} */
         let lastStop = null;
```
ludofischer, line 74: `lastStop` is going to contain the result of `valueParser.unit()`, which is either `Dimension` or `false`. Not sure whether this `null` is intentional or not because the same pattern is repeated elsewhere but with `lastStop` initialized to `false`. But in this case there's an explicit `if (lastStop === null)` instead of `(!lastStop)`.

ludofischer (commented):
In my view TypeScript does highlight potential bugs, but for now I've used casts almost anywhere as I don't want to introduce regressions by trying to fix the types...

PR #1335, opened 2022-02-25
chore: fix license of of Mozilla data

ludofischer:
They switched to Creatice Commons Zero in 2018 https://github.com/mdn/data/blob/main/LICENSE

I cannot find that the CSSDATA template mentioned in the original README
is used anywhere in the code right now and the template does not exist
on MDN any more either.

### DIFF
--- a/packages/postcss-reduce-initial/README.md
+++ b/packages/postcss-reduce-initial/README.md
@@ -85,10 +85,7 @@ See [CONTRIBUTORS.md](https://github.com/cssnano/cssnano/blob/master/CONTRIBUTOR
 
 ## License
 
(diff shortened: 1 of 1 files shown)


alexander-akait (approved):
I am afraid we can't do it, because we get infromatio

ludofischer:
> I am afraid we can't do it, because we get infromatio

Mozilla themselves changed the license, it should be OK I think?

alexander-akait:
Hope yes, in theory we can inline code and remove script, and just check it or adding them by request

ludofischer:
> Hope yes, in theory we can inline code and remove script, and just check when we need it or adding them by request

But then minification would not work without a connection to the public Internet? Seems a bit annoying for the users.

PR #1336, opened 2022-02-25
chore: fix various issues that TypeScript found

ludofischer:
* missing undefined or null checks
* missing checks on the appropriate node type
* functions that return a mix of boolean and falsy values

I think it might be easier to review if I separate the changes that affect the running code from comments and generated types.

### DIFF
--- a/packages/postcss-discard-unused/src/index.js
+++ b/packages/postcss-discard-unused/src/index.js
@@ -54,7 +54,9 @@ function hasFont(fontFamily, cache, comma) {
 function filterFont({ atRules, values }, comma) {
   values = [...new Set(values)];
   atRules.forEach((r) => {
-    const families = r.nodes.filter(({ prop }) => prop === 'font-family');
+    const families = r.nodes.filter(
+      (node) => node.type === 'decl' && node.prop === 'font-family'
+    );
 
     // Discard the @font-face if it has no font-family
     if (!families.length) {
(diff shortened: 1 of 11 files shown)


alexander-akait (approved):

packages/postcss-discard-unused/src/index.js
```diff
@@ -54,7 +54,9 @@ function hasFont(fontFamily, cache, comma) {
 function filterFont({ atRules, values }, comma) {
   values = [...new Set(values)];
   atRules.forEach((r) => {
-    const families = r.nodes.filter(({ prop }) => prop === 'font-family');
+    const families = r.nodes.filter(
+      (node) => node.type === 'decl' && node.prop === 'font-family'
```
ludofischer, line 58: Same issue, `nodes` is of type `ChildNode`, so we need to check whether it's a delcaration, even if in this case we know it's the child of a Rule.

packages/postcss-discard-unused/src/index.js
```diff
@@ -94,26 +96,27 @@ function pluginCreator(opts) {
         OnceExit(css, { list }) {
           const { comma, space } = list;
           css.walk((node) => {
-            const { type, prop, selector, name } = node;
+            const { type } = node;
```
ludofischer, line 99: Need to destructure these later as right now  TypeScript cannot know the node type.

packages/postcss-merge-longhand/src/lib/decl/borders.js
```diff
@@ -132,16 +133,12 @@ function isCloseEnough(mapped) {
...
-  return mapped.reduce((a, b) => {
-    a = Array.isArray(a) ? a : [a];
-
-    if (!a.includes(b)) {
-      a.push(b);
-    }
-
-    return a;
-  });
+  return [...new Set(mapped)];
```
ludofischer, line 141: All the previous code does in deduplicating an array.

packages/postcss-merge-longhand/src/lib/decl/borders.js
```diff
@@ -220,6 +217,7 @@ function merge(rule) {
 
           return true;
         }
+        return false;
```
ludofischer, line 220: This is to avoid the `boolean|undefined` return type.

packages/postcss-merge-longhand/src/lib/validateWsc.js
```diff
@@ -16,7 +16,7 @@ const styles = new Set([
 ]);
 
 function isStyle(value) {
-  return value && styles.has(value.toLowerCase());
+  return value !== undefined && styles.has(value.toLowerCase());
```
ludofischer, line 19: Technically the old code is different as it would return `value` if value was falsy, but really the only case is `undefined`. I don't think it ever gets assigned to `null`.

packages/postcss-minify-selectors/src/index.js
```diff
@@ -33,26 +33,28 @@ function attribute(selector) {
...
-  selector.raws.spaces.operator = {
-    before: '',
-    after: '',
-  };
-
-  selector.raws.spaces.value = {
-    before: '',
-    after: selector.insensitive ? ' ' : '',
-  };
+  if (selector.raws.spaces) {
```
ludofischer, line 36: Need to check whether `raws.spaces` is defined before assigning values to it.

packages/postcss-normalize-positions/src/index.js
```diff
@@ -156,25 +156,27 @@ function transform(value) {
       return;
     }
 
-    if (firstNode === 'center' && directionKeywords.has(secondNode)) {
-      nodes[0].value = nodes[1].value = '';
+    if (secondNode !== null) {
```
ludofischer, line 159: `secondNode` gets explicitly assigned to `null` in some cases before, so all the following code needs this check.

PR #1337, opened 2022-02-25
chore: replace do-while and other refactors to simplify types

ludofischer:
chore: replace do-while with while

The first loop execution would crash if the while condition was false.

chore: fix bollean return type

chore: remove redundant null initializations

chore: make variables const

chore: correct return types

### DIFF
--- a/packages/postcss-merge-longhand/src/lib/decl/borders.js
+++ b/packages/postcss-merge-longhand/src/lib/decl/borders.js
@@ -54,7 +54,7 @@ function getLevel(prop) {
 }
 
 const isValueCustomProp = (value) =>
-  value && value.search(/var\s*\(\s*--/i) !== -1;
+  value !== undefined && value.search(/var\s*\(\s*--/i) !== -1;
 
 function canMergeValues(values) {
   return !values.some(isValueCustomProp);
@@ -160,7 +160,7 @@ function explode(rule) {
           insertCloned(decl.parent, decl, { prop: direction });
         });
 
-        return decl.remove();
+        decl.remove();
       }
     }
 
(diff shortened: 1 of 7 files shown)


packages/postcss-merge-longhand/src/lib/decl/borders.js
```diff
@@ -160,7 +160,7 @@ function explode(rule) {
           insertCloned(decl.parent, decl, { prop: direction });
         });
 
-        return decl.remove();
+        decl.remove();
```
ludofischer, line 163: `rule.walkDecls` callback return type should be `false|undefined`.

packages/postcss-minify-font-values/src/lib/minify-family.js
```diff
@@ -33,9 +33,9 @@ const regexSimpleEscapeCharacters = /[ !"#$%&'()*+,.\/;<=>?@\[\\\]^`{|}~]/;
 
 function escape(string, escapeForString) {
   let counter = 0;
-  let character = null;
-  let charCode = null;
-  let value = null;
+  let character;
```
ludofischer, line 36: There's no checks against `null` specifically so assigning to `null` just adds one possible value to the type for no gain.

packages/stylehacks/src/plugins/leadingStar.js
```diff
@@ -23,7 +23,7 @@ module.exports = class LeadingStar extends BasePlugin {
           });
         }
       });
-      let { before } = node.raws;
+      const { before } = node.raws;
```
ludofischer, line 26: TypeScript seems to lose track of these if they are not `const`.

alexander-akait (approved):

PR #1340, opened 2022-03-01
feat: add CSS types to remaining packages

ludofischer:
Closes #1333 

This PR should contain only changes to comments, `package.json` and generated types. Types are strict everywhere, but I've defined some explicit `any` parameters in the cssnano package itself, because after trying to fit the types with the code and the documentation, I think not all configuration options work in practice, so I don't want to restrict what's being passed or change the code for the moment.


alexander-akait (approved):
Sometimes I think prettier should supports jsdocs

PR #1342, opened 2022-03-01
Publish cssnano 5.1.0

ludofischer:


pnpm-lock.yaml
```diff
@@ -62,9 +62,9 @@ importers:
   packages/cssnano:
     specifiers:
       autoprefixer: ^10.3.7
-      cssnano-preset-advanced: ^5.1.12
-      cssnano-preset-default: ^5.1.12
-      cssnano-preset-lite: ^2.0.3
+      cssnano-preset-advanced: ^5.2.0
```
ludofischer, line 65: Maybe we should configure dependencies inside the workspace with the workspace:^` option, I always forget to update the lockfile after updating the version number.

alexander-akait (approved):

PR #1344, opened 2022-03-02
docs: switch website to mostly static website

ludofischer:
In the last few months I have been playing with a few static website generators, and I think this website version might be more better than the current website. This moves most of the rendering to [Eleventy](https://www.11ty.dev/).  I've uploaded this new version here: https://621f9f84e7ff77309b8bb63d--telephone-operator-bertie-55061.netlify.app/

### Documentation pages improvements
We save a ton of dependencies, and we also need less configuration, because eleventy can directly read external files like the changelog and data files like the plugin metadata. The website is automatically in sync with the changelog and we don't need to regenerate the pages itself for a new release.
Building the docs takes about a second compared to tens of second with the previous system. At the same time the website is lighter, Lighthouse measures first contentful paint after 0.8s compared to 1.8s previously. 
Since generating pages from JSON is easier,  I've added back the examples to the optimisation pages. documentation, that was already in the metadata but not displayed. The uses cssnano itself to minify the CSS. The styling itself is a mixture of the [Astro](https://astro.build/) documentation template, the Docusaurus styles and the techniques in [EveryLayout](https://every-layout.dev/).

### Playground improvement
I've also removed React completely as the CodeMirror editor is responsible for most of the UI and the code to run PostCSS has not UI. Since the webpack configuration is not shared with Docusaurus, I've added some code splitting to the playground, so the only the minimize button is disabled and shows a loading message when loading cssnano itself. The playground becomes interactive faster under a simulated 3G connection. Lighthouse gives 90 in performance and 98 in accessibility compared to 68 and 90 for the previous version. 

### Missing features
Some features from the Docusaurus versions are still missing: the navigation does not collapse to a drop down on small screens, so on iPhone SE you would need to scroll to read the content, and there is no single page index. I could re-add them if necessary.


ludofischer:
> There are a lot of changes (hard to review), but yes, let's move ahead

I've removed one set of changes and split the remaining into three commits, hopefully it's easier to review now.

alexander-akait (approved):
Looks good

PR #1348, opened 2022-03-07
fix: preserve comment removal with PostCSS > 8.4.6

ludofischer:
Since PostCSS 8.4.6, the AST has begun changing a little when the source contains comments. I've also run the tests before updating PostCSS to check it works with the previous AST too: https://github.com/cssnano/cssnano/actions/runs/1947829642


alexander-akait (approved):
Sounds like a breaking change in postcss

PR #1352, opened 2022-03-10
fix: preserve 1 space in empty custom property values

### LINKED ISSUE #1330: [Bug]: CSS custom properties in @supports cannot remove spaces
### Describe the bug

```css
@supports (--foo:    ) {
      html { background: green; }
}
```

### Expected behaviour

Browser page shows green background.
After compression, keep at least one space after `--foo: `, otherwise not valid in Chrome.

See also: https://lea.verou.me/2020/10/the-var-space-hack-to-toggle-multiple-values-with-one-custom-property/

```css
@supports (--foo: ){html{background:green}}
```

### Steps to reproduce

https://cssnano.co/playground/#eyJpbnB1dCI6IkBzdXBwb3J0cyAoLS1mb286ICAgICkge1xuICAgICAgaHRtbCB7IGJhY2tncm91bmQ6IGdyZWVuOyB9XG59IiwiY29uZmlnIjoiY3NzbmFuby1wcmVzZXQtZGVmYXVsdCJ9

### Version

5.0

### Preset

default

### Environment

```shell
/
```

### Package details

```shell
/
```

### Additional context

_No response_

ludofischer:
Fix  #1330
I had to change two plugins because `postcss-minify-params` handles spaces inside AtRules separately. I think the AST inn `postcss-minify-params` does not make sense because `postcss-minify-params` uses `postcss-value-parser` to parse the AtRule parameters, but they do not conform to the value syntax.
I've thought of refactoring more but at the moment I haven't got any idea of how to better parse the AtRule parameters.
I've added tests for a few different cases and all the previous tests do not need changes.

### DIFF
--- a/packages/postcss-minify-params/test/index.js
+++ b/packages/postcss-minify-params/test/index.js
@@ -184,6 +184,22 @@ test(
 );
 
 test(
+  'should normalize space in custom property values',
+  processCSS(
+    '@supports (--foo:  ){html{background:green}}',
+    '@supports (--foo: ){html{background:green}}'
+  )
+);
+
+test(
+  'should minimize custom properties with multiple conditions',
+  processCSS(
+    '@supports ((--foo:  ) or (--bar: green )){html{background:green}}',
+    '@supports ((--foo: ) or (--bar:green)){html{background:green}}'
+  )
+);
+
+test(
   'should not throw on empty parentheses',
   passthroughCSS('@media (){h1{color:blue}}')
 );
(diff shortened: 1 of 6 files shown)


packages/postcss-minify-params/test/index.js
```diff
@@ -183,6 +183,22 @@ test(
   )
 );
 
+test(
+  'should normalize space in custom property values',
+  processCSS(
+    '@supports (--foo:  ){html{background:green}}',
```
alexander-akait, line 189: It should be only one space, i.e. if we have empty values, keep as is, if we have empty, let print empty space

PR #1356, opened 2022-03-11
fix(postcss-merge-longhand): don't explode custom properties

### LINKED ISSUE #675: mergeLonghand is unsafe with border and CSS variables

### LINKED ISSUE #1354: [Bug]: invalid border shorthand with custom properties
### Describe the bug

the following:

```css
h1 {
  --width: 8px 2px;
  border-width: var(--width);
  border-style: solid;
  border-color: hotpink;
}
```

results in (default preset):

```css
h1{--width:8px 2px;border:var(--width) solid hotpink}
```

which is invalid.

### Expected behaviour

when a custom property is encountered, don't collapse border properties into shorthand.

### Steps to reproduce

1. paste the above snippet in the cssnano playground
2. click minimize

### Version

whatever is currently on the cssnano playground, 5.1.0 probably

### Preset

default

### Environment

```shell
System:
    OS: Linux 5.13 KDE neon 5.24
    CPU: (4) x64 Intel(R) Core(TM) i5-6400T CPU @ 2.20GHz
    Memory: 19.96 GB / 31.30 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
```

### Package details

```shell
=> Found "postcss@8.4.5"
```

### Additional context

_No response_

ludofischer:
Fix #1354
Fix #675

- postcss-merge-longhand first 'explodes' longhand declarations to the smallest
unit (for example border-width -> border-top-width).
If the declaration value contains a custom property, it is not possible
to safely explode width/size/color, as the custom property might be equivalent to multiple
space-separated values.

- this plugin lowercases property names as a side-effect of exploding,
so to preserve the old behaviour we lowercase the property that cannot be
exploded

- remove incorrect tests and add tests for new behaviour
- helps with #1044 , but only in the case where the declarations contain a custom property. The issue mentioned in https://github.com/cssnano/cssnano/issues/1044#issuecomment-840731071 is not fixed

### DIFF
--- a/packages/postcss-merge-longhand/src/lib/decl/borders.js
+++ b/packages/postcss-merge-longhand/src/lib/decl/borders.js
@@ -231,6 +231,10 @@ function explode(rule) {
         return false;
       }
 
+      if (isCustomProp(decl)) {
+        decl.prop = decl.prop.toLowerCase();
+        return false;
+      }
       parseTrbl(decl.value).forEach((value, i) => {
         insertCloned(
           /** @type {import('postcss').Rule} */ (decl.parent),
--- a/packages/postcss-merge-longhand/test/borders.js
+++ b/packages/postcss-merge-longhand/test/borders.js
@@ -1154,22 +1154,6 @@ test(
 );
 
 test(
-  'Should correctly merge borders with custom properties (#652)',
-  processCSS(
-    'h1{border-width:var(--a);border-style:var(--b);border-color:var(--c)}',
-    'h1{border:var(--a) var(--b) var(--c)}'
-  )
-);
-
-test(
-  'Should correctly merge borders with custom properties (#652) (uppercase)',
-  processCSS(
-    'h1{BORDER-WIDTH:VAR(--A);BORDER-STYLE:VAR(--B);BORDER-COLOR:VAR(--C)}',
-    'h1{border:VAR(--A) VAR(--B) VAR(--C)}'
-  )
-);
-
-test(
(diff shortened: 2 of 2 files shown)


alexander-akait (approved):

alexander-akait (commented):
@ludofischer Can we add test case from https://github.com/vercel/next.js/issues/35219 to ensure we fix it (I think we fix it, just test case)

alexander-akait:
@ludofischer Let's do release today, thanks

ludofischer:
> @ludofischer Let's do release today, thanks

Should be available https://www.npmjs.com/package/cssnano

alexander-akait:
Big thanks

PR #1358, opened 2022-03-11
chore: use workspace protocol for local dependencies

ludofischer:
In this way the lockfile does not have to change for every version bump.
I've tested updating the versions with changesets and it seems compatible
as advertised. It's also clearer which packages are in the workspace and which not.


alexander-akait (approved):

stof:
should releases really use `^*` constraints ? Those are unbound constraints.

ludofischer:
> should releases really use `^*` constraints ? Those are unbound constraints.

You're right, I wanted to rely on the pnpm [workspace protocol](https://pnpm.io/workspaces) to automatically replace with the version number on release but I have used the wrong syntax.

ludofischer:
I've opened a pull request to fix this: https://github.com/cssnano/cssnano/pull/1372

PR #1359, opened 2022-03-11
fix(postcss-merge-longhand): preserve custom property case

### LINKED ISSUE #847: mergeLonghand lowercases values with multiple var references
This could be considered a reopening of #648, or a bug with the implemented fix in #649. That did fix simple cases but it still fails when values contain more than a single `var(--X)` reference. An example:

```
border: 1px solid hsla(var(--HUE), var(--SATURATION), var(--LUMINANCE), 0.5)
```

incorrectly transforms to:

```
border: 1px solid hsla(var(--hue), var(--saturation), var(--LUMINANCE), 0.5)
```

The culprit appears to be the greedy leading `.*` in the regular expression at https://github.com/cssnano/cssnano/blob/3580e7be5a4d32ed2b122d13b87380d2b150c705/packages/postcss-merge-longhand/src/lib/parseWsc.js#L6

This toLower logic could probably be fixed to handle strings with multiple vars. However, I'd question whether lowercasing should be happening in parseWsc.js at all. I feel like the mergeLonghand plugin should be limiting itself to only merging, and leave lowercasing up to other specialized plugins that can handle that in a more robust way.

ludofischer:
Fix #847

Locate every custom property inside a value, then build a
new string by preserving the custom properties and lowercasing
the rest.

### DIFF
--- a/packages/postcss-merge-longhand/src/lib/parseWsc.js
+++ b/packages/postcss-merge-longhand/src/lib/parseWsc.js
@@ -4,14 +4,29 @@ const { isWidth, isStyle, isColor } = require('./validateWsc.js');
 
 const none = /^\s*(none|medium)(\s+none(\s+(none|currentcolor))?)?\s*$/i;
 
-const varRE = /(^.*var)(.*\(.*--.*\))(.*)/i;
-/** @type {(p: RegExpExecArray) => string} */
-const varPreserveCase = (p) =>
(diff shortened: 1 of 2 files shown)


alexander-akait (approved):
Will be great to rewrite this code on good parser function, no regexps, but it can be tricky, so looks good

ludofischer:
> Will be great to rewrite this code on good parser function, no regexps, but it can be tricky, so looks good

I've also thought about that, but I think the lowercasing should probably be removed from this plugin eventually.

PR #1366, opened 2022-03-16
fix: add grid-column and grid-row checks to reduce-idents grid-template

jordydhoker:


packages/postcss-reduce-idents/src/lib/grid-template.js
```diff
@@ -41,7 +41,7 @@ module.exports = function () {
         });
 
         declCache.push(node);
-      } else if (node.prop.toLowerCase() === 'grid-area') {
+      } else if (/^(grid-area|grid-column|grid-row)/i.test(node.prop)) {
```
ludofischer, line 44: I think this includes the `-start` and `-end` properties as well. Wouldn't it be better to list the properties explicitly instead of relying on a regular expression? That way there are no surprises if a new property is introduced or if someone uses non-standard syntax. Something like

```js
const gridProperties = new Set(['grid-somethingg'....']);
if (gridProperties.has(node.prop.toLowerCase()) {
```
Also avoids recreating a RegEx object (although the performance gain is likely undetectable).

packages/postcss-reduce-idents/test/index.js
```diff
@@ -249,40 +249,48 @@ test(
...
 
 test(
-  'should rename grid-template-areas and grid-area (uppercase)',
+  'should rename grid-template-areas, grid-area, grid-column and grid-row (uppercase)',
   processCSS(
     [
       'body{GRID-TEMPLATE-AREAS:"head head" \n"nav  main"\n"nav  foot";}',
       'header { GRID-AREA: head }',
       'nav{GRID-AREA:nav}',
       'main{GRID-AREA:main}',
```
ludofischer, line 282: Could be worthwhile to add more tests for more properties like `grid-row-end` so nobbody forgets them in the future.

packages/postcss-reduce-idents/src/lib/grid-template.js
```diff
@@ -20,6 +20,16 @@ module.exports = function () {
   /** @type {import('postcss').Declaration[]} */
   let declCache = [];
 
+  const gridChildProperties = [
```
ludofischer, line 23:
```suggestion
  const gridChildProperties = new Set([
```

packages/postcss-reduce-idents/src/lib/grid-template.js
```diff
@@ -20,6 +20,16 @@ module.exports = function () {
...
+    'grid-area',
+    'grid-column',
+    'grid-row',
+    'grid-column-start',
+    'grid-column-end',
+    'grid-row-start',
+    'grid-row-end',
+  ];
```
ludofischer, line 31:
```suggestion
  ]);
```

packages/postcss-reduce-idents/src/lib/grid-template.js
```diff
@@ -65,7 +75,7 @@ module.exports = function () {
               node.value = node.value.replace(/\s+/g, ' '); // merge white-spaces
             }
 
-            if (decl.prop.toLowerCase() === 'grid-area' && !isNum(node)) {
+            if (gridChildProperties.includes(decl.prop.toLowerCase()) && !isNum(node)) {
```
ludofischer, line 78:
```suggestion
            if (gridChildProperties.has(decl.prop.toLowerCase()) && !isNum(node)) {
```

packages/postcss-reduce-idents/src/lib/grid-template.js
```diff
@@ -41,7 +51,7 @@ module.exports = function () {
         });
 
         declCache.push(node);
-      } else if (node.prop.toLowerCase() === 'grid-area') {
+      } else if (gridChildProperties.includes(node.prop.toLowerCase())) {
```
ludofischer, line 54:
```suggestion
      } else if (gridChildProperties.has(node.prop.toLowerCase())) {
```

ludofischer (commented):
In  my view a set better because it has far fewer methods than an array; it makes it clearer that you're going to only use it for checking membership. `has()` is faster than `includes()` although probably not noticeable with so few elements.

ludofischer (approved):
Thanks!

PR #1368, opened 2022-03-19
Publish cssnano-preset-advanced and postcss-reduce-idents

ludofischer:
Release the changes in https://github.com/cssnano/cssnano/pull/1366


ludofischer:
After a second look at the code I think I should check a few more things before releasing.

alexander-akait (approved):

ludofischer:
There's a bug in the current postcss-reduce-idents, it transforms upppercase reserved words (so for example `AUTO` becomes `c`). See https://github.com/cssnano/cssnano/pull/1369 for the fix

PR #1369, opened 2022-03-19
fix(postcss-reduce-idents): fix handling of reserved keywords

ludofischer:
* fix issue where uppercase special values (AUTO) were transformed
* skip useless parsing of `grid-template-rows` and `grid-template-columns` as no substitutions occur there (the code always checks whether the token type is `string` before considering a candidate for replacing, so there was not bug but the intents was confusing)

### DIFF
--- a/packages/postcss-reduce-idents/src/lib/grid-template.js
+++ b/packages/postcss-reduce-idents/src/lib/grid-template.js
@@ -58,7 +58,10 @@ module.exports = function () {
         declCache.push(node);
       } else if (gridChildProperties.has(node.prop.toLowerCase())) {
         valueParser(node.value).walk((child) => {
-          if (child.type === 'word' && !RESERVED_KEYWORDS.has(child.value)) {
+          if (
+            child.type === 'word' &&
+            !RESERVED_KEYWORDS.has(child.value.toLowerCase())
+          ) {
             addToCache(child.value, encoder, cache);
           }
         });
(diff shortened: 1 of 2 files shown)


packages/postcss-reduce-idents/src/lib/grid-template.js
```diff
@@ -53,7 +58,10 @@ module.exports = function () {
         declCache.push(node);
       } else if (gridChildProperties.has(node.prop.toLowerCase())) {
         valueParser(node.value).walk((child) => {
-          if (child.type === 'word' && !RESERVED_KEYWORDS.has(child.value)) {
+          if (
+            child.type === 'word' &&
+            !RESERVED_KEYWORDS.has(child.value.toLowerCase())
```
ludofischer, line 63: @jordydhoker in the previous PR, we forgot that the child value could be an uppercase reserved keyword)

alexander-akait (approved):

packages/postcss-reduce-idents/src/lib/grid-template.js
```diff
@@ -66,7 +74,7 @@ module.exports = function () {
       declCache.forEach((decl) => {
         decl.value = valueParser(decl.value)
           .walk((node) => {
-            if (/(grid-template|grid-template-areas)/i.test(decl.prop)) {
```
ludofischer, line 69: `grid-template` already matches `grid-template-areas` but also matches more than intended (namely `grid-template-rows` and `grid-template-columns`).

packages/postcss-reduce-idents/src/lib/grid-template.js
```diff
@@ -20,23 +35,13 @@ module.exports = function () {
   /** @type {import('postcss').Declaration[]} */
   let declCache = [];
 
-  const gridChildProperties = new Set([
```
ludofischer, line 23: Moving this outside of the function as there's no need to re-create it for every call.

PR #1376, opened 2022-03-27
fix: conflict detection needs to consider border-left-color vs border-color

### LINKED ISSUE #1375: [Bug]: mergeLonghand border with variables
### Describe the bug

Consider this input:

`h1{border:1px solid;border-color:var(--BORDER);border-left-style:none;}`

With this input, starting with postcss-merge-longhand@5.1.1, the output becomes

`h1{border:1px solid;border-left:1px}`

Note that the border color is dropped entirely!

### Expected behaviour

With version 5.1.0 and earlier, the output is,

`h1{border:1px solid;border-color:var(--BORDER);border-left:1px var(--BORDER);}`

It's slightly questionable whether this is an improvement over the original input, but it at least works.

### Steps to reproduce

Go to https://cssnano.co/playground/, paste the input from above, should see the incorrect output come up.

### Version

5.1.4

### Preset

(no preset)

### Environment

```shell
System:
    OS: Linux
    CPU: x64 Intel
    Shell: bash
```

### Package details

```shell
├── cssnano@5.1.4
└── postcss@8.3.6
```

### Additional context

_No response_

imirkin:
The previous algorithm was to pop values off the end and comparing to
the other given property. Change the detection to instead check that all
the components of the second value are in the first value.

Fixes #1375

Please note that I'm new to this codebase, so I may very well have missed something. The existing tests do pass though.

### DIFF
--- a/packages/postcss-merge-longhand/src/lib/mergeRules.js
+++ b/packages/postcss-merge-longhand/src/lib/mergeRules.js
@@ -9,17 +9,23 @@ const getRules = require('./getRules.js');
  * @return {boolean}
  */
 function isConflictingProp(propA, propB) {
-  if (!propB.prop || propB.important !== propA.important) {
+  if (!propB.prop || propB.important !== propA.important ||
+      propA.prop === propB.prop) {
     return false;
   }
 
-  const parts = propA.prop.split('-');
+  const partsA = propA.prop.split('-');
+  const partsB = propB.prop.split('-');
 
-  return parts.some(() => {
-    parts.pop();
+  /* Be safe: check that the first part matches. So we don't try to
+   * combine e.g. border-color and color.
+   */
+  if (partsA[0] !== partsB[0]) {
+    return false;
+  }
 
-    return parts.join('-') === propB.prop;
-  });
+  const partsASet = new Set(partsA);
+  return partsB.every((partB) => partsASet.has(partB));
 }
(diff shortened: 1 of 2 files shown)


ludofischer (commented):
Thanks for this PR! Looks good, but could you fix the lint errors? Run `pnpm lint` and fix what it reports.

ludofischer:
> Thanks for this PR! Looks good, but could you fix the lint errors? Run `pnpm lint` and fix what it reports.

You can also run `pnpm fixlint` to try to fix the errors automatically.

imirkin:
> Thanks for this PR! Looks good, but could you fix the lint errors? Run `pnpm lint` and fix what it reports.

I think this is done. When I run lint, it also updates index.d.ts definitions, which seem to get a lot of updates for me. Probably some difference in local environment though, or other screwup?

ludofischer:
> > Thanks for this PR! Looks good, but could you fix the lint errors? Run `pnpm lint` and fix what it reports.
> 
> I think this is done. When I run lint, it also updates index.d.ts definitions, which seem to get a lot of updates for me. Probably some difference in local environment though, or other screwup?

Yes, I think I did not commit the type definitions changes at some point.

ludofischer (approved):

alexander-akait (approved):

PR #1377, opened 2022-03-27
chore(cssnano-preset-default): update css-declaration-sorter

ludofischer:
Drop timsort dependency and fix issues with imports. [Changelog](https://github.com/Siilwyn/css-declaration-sorter/blob/master/changelog.md)

### DIFF
--- a/packages/cssnano-preset-default/package.json
+++ b/packages/cssnano-preset-default/package.json
@@ -11,7 +11,7 @@
   ],
   "license": "MIT",
   "dependencies": {
-    "css-declaration-sorter": "^6.0.3",
+    "css-declaration-sorter": "^6.2.1",
     "cssnano-utils": "workspace:^",
     "postcss-calc": "^8.2.3",
     "postcss-colormin": "workspace:^",
--- a/packages/cssnano-preset-default/package.json
+++ b/packages/cssnano-preset-default/package.json
@@ -11,7 +11,7 @@
   ],
   "license": "MIT",
   "dependencies": {
-    "css-declaration-sorter": "^6.2.1",
+    "css-declaration-sorter": "^6.2.2",
     "cssnano-utils": "workspace:^",
     "postcss-calc": "^8.2.3",
     "postcss-colormin": "workspace:^",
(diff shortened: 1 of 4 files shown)


Siilwyn:
Can I help with this? Seems that something is wrong with the TS type?

ludofischer:
> Can I help with this? Seems that something is wrong with the TS type?

Yes, I think I know what the problem is. cssnano expects the plugins to conform to the postcss `PluginCreator` interface: https://github.com/postcss/postcss/blob/e5c10e009717c14c00b03867344eaefe960968da/lib/postcss.d.ts#L196

which means the function that returns the plugin must have the `postcss` property set to true.

css-declaration-sorter does set the property in JS:
https://github.com/Siilwyn/css-declaration-sorter/blob/82461efa9d49d44b611f2d53e85ab6988b5dc41c/src/main.mjs#L37
but then it's not declared in the TypeScript types.

Siilwyn:
I see! Just pushed 6.2.2 which uses `PluginCreator`, I also reverted the typing to export the default CommonJS entry. 

Since having a named export *and* default export for both ESM and CJS in a TS type definition file is not possible, I think I'll push a major bump later so I can just use a named export. 🤔

ludofischer:
> I see! Just pushed 6.2.2 which uses `PluginCreator`, I also reverted the typing to export the default CommonJS entry.

Thank! Seems fine now.

Siilwyn:
There is still an override in the site package.json right?

https://github.com/cssnano/cssnano/blob/master/site/package.json#L25-L28

(don't have much experience with pnpm)

ludofischer:
> There is still an override in the site package.json right?

1. you're right
2. it had no effect anyway since it's getting cssnano from the parent directory
3. the playground seems to be still broken with the latest css-declaration-sorter
```
Cannot find module '../orders/alphabetical.cjs'
```

Siilwyn:
That is odd, the CJS build inlines the orders so this seems like old code?

Or Webpack is loading the ESM build and coverts imports to cjs?

Thoughts?

ludofischer:
> That is odd, the CJS build inlines the orders so this seems like old code?
> 
> Or Webpack is loading the ESM build and convers imports to cjs?
> 
> Thoughts?

Works fine now after deleting all `node_modules`. I think webpack was still bundling the previous css-declaration-sorter version lying around in the pnpm cache. Might be an issue with webpack's caching.

Siilwyn (approved):
Looks good! ✨

alexander-akait (approved):

PR #1383, opened 2022-04-01
docs: update website dependencies

ludofischer:
Update both the playground dependencies and the build tools. Browsersync (eleventy dependencies) updated socket.io from 2 to 4 in a patch release, which gets rid of a bunch of unmaintained dependencies, which explains the diff in the lockfile.

### DIFF
--- a/site/src/playground.njk
+++ b/site/src/playground.njk
@@ -31,7 +31,7 @@
       <span class="center">Loading CSS editor…</span>
     </div>
     <script type="module" src="/js/runtime.bundle.js"></script>
-    <script type="module" src="/js/628.bundle.js"></script>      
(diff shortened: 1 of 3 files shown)


site/src/playground.njk
```diff
@@ -31,7 +31,7 @@
       <span class="center">Loading CSS editor…</span>
     </div>
     <script type="module" src="/js/runtime.bundle.js"></script>
-    <script type="module" src="/js/628.bundle.js"></script>      
+    <script type="module" src="/js/932.bundle.js"></script>
```
ludofischer, line 34: We could use the webpack manifest plugin to update this automatically, but it does not change that often (it changed when bumping webpack from 5.70 to 5.71)

alexander-akait (approved):

PR #1385, opened 2022-04-08
Don't convert `min-width: 0%` to `min-width: 0`

nex3:
This transformation isn't safe on IE11. See https://codepen.io/nex3/full/XWVqJrJ as an example of `min-width: 0` breaking IE11's flexbox behavior where `min-width: 0%` does not.

### DIFF
--- a/packages/postcss-convert-values/src/index.js
+++ b/packages/postcss-convert-values/src/index.js
@@ -36,6 +36,13 @@ const keepWhenZero = new Set([
   'line-height',
 ]);
 
+// Can't remove the % on these properties when they're 0
+const keepZeroPercent = new Set([
+  'max-height',
+  'height',
+  'min-width',
+]);
+
 /**
  * Numbers without digits after the dot are technically invalid,
  * but in that case css-value-parser returns the dot as part of the unit,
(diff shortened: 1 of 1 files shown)


ludofischer (commented):
I trust the problem is real  although I could not test in IE 11 because codepen asks me to complete a CAPTCHA which does not display on IE 11 :-) Since there are no tests, how did you check that this change solves it (it looks like it would but as there is always the chance that some other transform is also removing the %)

alexander-akait:
We should avoid it only when you have IE11 in browserslist, because it is flexbox bug in IE11

ludofischer:
> We should avoid it only when you have IE11 in browserslist, because it is flexbox bug in IE11

Currently `convert-values` does not use Browserslist at all, so we need to add it as a dependency and retrieve the browsers somewhere.

nex3:
I just tested manually by removing `postcss-convert-values` from my stack. I've added a test.

I'm happy to add a browserslist dependency if you can point me to an existing package that does so, so I can copy the pattern.

ludofischer (approved):
I think we can merge this and add the browserslist check in another commit.

alexander-akait (approved):

PR #1387, opened 2022-04-16
fix: correct file paths in scripts that fetch external data

ludofischer:
* correct the file paths that changed when transpiling was
removed and TypeScript checks added
* switch fetch library to undici since its got zero dependencies
  and it's going to be the basis for built-in Node.js fetch
* did not commit changes to the data files from running the scripts
  again because they need a separate review for browser compatibility

### DIFF
--- a/packages/postcss-ordered-values/script/buildListStyleType.js
+++ b/packages/postcss-ordered-values/script/buildListStyleType.js
@@ -3,7 +3,7 @@
 
 const { writeFile } = require('fs');
 const { resolve } = require('path');
-const fetch = require('node-fetch');
+const { fetch } = require('undici');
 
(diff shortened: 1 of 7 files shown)


packages/postcss-ordered-values/script/buildListStyleType.js
```diff
@@ -26,15 +26,4 @@ const listTypeURL =
       }
     }
   );
-
-  writeFile(
```
ludofischer, line 30: This block of code is duplicated below. It was copied to update  the file path but then the original code was not deleted, which caused a crash when trying to run the script.

alexander-akait (approved):

PR #1397, opened 2022-05-20
fix(postcss-merge-longhand): preserve fallback for color() function

### LINKED ISSUE #1396: [Bug]: `border-color` fallback incorrectly removed by `mergeLonghand`
### Describe the bug

On the default preset, `border-color` is incorrectly having a fallback removed, while `background-color` does not. This does not happen on the `lite` preset in the Playground. I was able to narrow it down to `mergeLonghand` locally (turning it off fixes this) but I have not put together a proper repro proving that that's the cause.

### Input

```
.border-grey-1000 {
  border-color: rgb(37, 45, 49);
  border-color: color(display-p3 0.1451 0.1765 0.1922 / 1);
}

.bg-green-500 {
  background-color: rgb(72, 213, 151);
  background-color: color(display-p3 .2824 .8353 .5922 / 1);
}
```

### Output (whitespace added back in for readability):

```
.border-grey-1000 {
  border-color: color(display-p3 .1451 .1765 .1922/1)
}

.bg-green-500{
  background-color: #48d597;
  background-color: color(display-p3 .2824 .8353 .5922/1)
}
```

### Expected behaviour

I expect to see the `border-color` fallback left in place.

### Steps to reproduce

See [Playground](https://cssnano.co/playground/#eyJpbnB1dCI6Ii5ib3JkZXItZ3JleS0xMDAwIHtcbiAgYm9yZGVyLWNvbG9yOiByZ2IoMzcsIDQ1LCA0OSk7XG4gIGJvcmRlci1jb2xvcjogY29sb3IoZGlzcGxheS1wMyAwLjE0NTEgMC4xNzY1IDAuMTkyMiAvIDEpO1xufVxuXG4uYmctZ3JlZW4tNTAwIHtcbiAgYmFja2dyb3VuZC1jb2xvcjogcmdiKDcyLCAyMTMsIDE1MSk7XG4gIGJhY2tncm91bmQtY29sb3I6IGNvbG9yKGRpc3BsYXktcDMgLjI4MjQgLjgzNTMgLjU5MjIgLyAxKVxufSIsImNvbmZpZyI6ImNzc25hbm8tcHJlc2V0LWRlZmF1bHQifQ==)

### Version

Whatever is running in the Playground, probably 5.1.8

### Preset

default

### Environment

```shell
Playground
```

### Package details

```shell
Playground
```

### Additional context

_No response_

ludofischer:
Fix #1396

### DIFF
--- a/packages/postcss-merge-longhand/src/lib/decl/borders.js
+++ b/packages/postcss-merge-longhand/src/lib/decl/borders.js
@@ -20,7 +20,8 @@ const { isValidWsc } = require('../validateWsc.js');
 
 const wsc = ['width', 'style', 'color'];
 const defaults = ['medium', 'none', 'currentcolor'];
-const colorMightRequireFallback = /hsla\(|rgba\(|color\(/i;
+const colorMightRequireFallback =
+  /(hsla|rgba|color|hwb|lab|lch|oklab|oklch)\(/i;
 
 /**
  * @param {...string} parts
--- a/packages/postcss-merge-longhand/src/lib/decl/borders.js
+++ b/packages/postcss-merge-longhand/src/lib/decl/borders.js
@@ -20,6 +20,7 @@ const { isValidWsc } = require('../validateWsc.js');
 
 const wsc = ['width', 'style', 'color'];
 const defaults = ['medium', 'none', 'currentcolor'];
+const colorMightRequireFallback = /hsla\(|rgba\(|color\(/i;
 
 /**
  * @param {...string} parts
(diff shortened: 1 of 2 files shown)


packages/postcss-merge-longhand/src/lib/decl/borders.js
```diff
@@ -20,6 +20,7 @@ const { isValidWsc } = require('../validateWsc.js');
 
 const wsc = ['width', 'style', 'color'];
 const defaults = ['medium', 'none', 'currentcolor'];
+const colorMightRequireFallback = /hsla\(|rgba\(|color\(/i;
```
alexander-akait, line 23: Looks good, but there are more color functions https://www.w3.org/TR/css-color-4/#color-syntax

ludofischer: Whoops, you're right. There's also an `isColor` function somewhere in the `postcss-merge-longhand` plugin, but updating that one has no effect on this bug, so I've left it alone as I don't understand what it does. I'll add the color functions to the regex.

ludofischer: I've added more color functions. To be consistent with how other features are handled, we could check if the browsers in browserslist support a color before merging it, on the other hand if the user has added a fallback, we could assume they know the color does not work on all target browsers. To do the merging properly we would need more precise parsing of the color values. For example, according to CanIUse, even IE6 supports `rgb()` with integer values, but no IE version supports `rgb` with float values.

alexander-akait: Yes, you are right, we can fix it in the next PR

alexander-akait (approved):

PR #1416, opened 2022-06-13
fix(postcss-merge-longhand): preserve hyphenated custom property case

ingmarh:
### Summary

Extends the regex that is used to locate custom properties inside a
value so that their case is preserved to also preserve the case of
custom property names that contain hyphen-minus characters.

For example, the case of the custom property in the following rule
should be preserved:

```css
border: 1px solid rgba(var(--colors-secondaryColor));
```

### Motivation

In our projects, we use CSS custom property names with hyphen-minus characters. These custom property names were lowercased via `postcss-merge-longhand` when minifying for production. Since hyphen-minus characters are valid in CSS custom property names, I think they should be allowed in the corresponding regex so that the case of hyphenated custom properties is preserved.

- [ ] Yes
- [x] No

### Checklist

### Documentation
Please check if any of these apply:

- [ ] My proposed change require an update to the documentation
- [ ] I have added or updated the relevant documentation


ludofischer (approved):
Looks good!

PR #1423, opened 2022-07-06
chore: update development dependencies

ludofischer:


ludofischer:
> Ci is failed...

I think uvu broke compatibility with Node.js in their latest patch release. I've opened an issue on the uvu repository https://github.com/lukeed/uvu/issues/213 as they still test on Node.js 10, so this is supposed to work.

alexander-akait (approved):

PR #1425, opened 2022-08-08
Fix `cssnanoPlugin` type to return Processor not Plugin

isker:
Invoking `postcss` gives you a Processor, not a Plugin.

### DIFF
--- a/packages/cssnano/src/index.js
+++ b/packages/cssnano/src/index.js
@@ -125,7 +125,7 @@ function resolveConfig(options) {
 /**
  * @type {import('postcss').PluginCreator<Options>}
  * @param {Options=} options
- * @return {import('postcss').Plugin}
+ * @return {import('postcss').Processor}
  */
 function cssnanoPlugin(options = {}) {
   if (Array.isArray(options.plugins)) {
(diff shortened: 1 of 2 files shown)


ludofischer:
> Ideally CI would have caught this problem - I'm not quite sure why it does not.

Me neither, running `tsc` currently gives no errors. How did you find out about the wrong type? Did you use PostCSS programmatically in TypeScript? After poring over the PostCSS type definitions, it does seem to me your change is correct.

isker:
> Did you use PostCSS programmatically in TypeScript?

Yes.  Correct programmatic usage of cssnano currently errors in TS.

ludofischer (approved):

ludofischer:
Could you show a sample where the type checking fails? I am convinced this change should be correct, but I cannot make TypeScript throw an error with the current types. Here's what I've tried:

```typescript
import * as cssnano from 'cssnano';
import postcss, { Result } from 'postcss';

import { readFile, writeFile } from 'fs';

readFile('./input.css', (err, css) => {
  postcss([cssnano({preset: 'default'})])
    .process(css, { from: './input.css', to: 'dest/app.css' })
    .then((result: Result) => {
      writeFile('dest/app.css', result.css, () => true)
      if ( result.map ) {
        writeFile('dest/app.css.map', result.map.toString(), () => true)
      }
    })
})
```
Compiled this with `tsc --strict`,  PostCSS 4.1.16, TypeScript 4.7.4, cssnano 5.1.12

PR #1430, opened 2022-08-27
chore: update TypeScript to 4.8.2

ludofischer:
I think all the type changes are in types used internally.


packages/postcss-merge-longhand/types/lib/colornames.d.ts
```diff
@@ -1,14 +1,2 @@
...
-    forEach(callbackfn: (value: string, value2: string, set: Set<string>) => void, thisArg?: any): void;
-    has(value: string): boolean;
-    readonly size: number;
-    entries(): IterableIterator<[string, string]>;
-    keys(): IterableIterator<string>;
-    values(): IterableIterator<string>;
-    [Symbol.iterator](): IterableIterator<string>;
-    readonly [Symbol.toStringTag]: string;
-};
+declare const _exports: Set<string>;
```
ludofischer, line 1: Looks like an improvement. Before it exported an object with all the methods of a Set instead.

alexander-akait (approved):

Issue #1436, opened 2022-10-08
[Bug]: reduceInitial ignore pointer-events not working


jamacoe:
### Describe the bug

In gulp 4: disabling reduceInitial from default preset work, so obv. I'm addressing the thing correct. But { ignore : ["pointer-events"] } doesn't have any effect on the output. **pointer-events: initial** will be translated to **pointer-events: auto**.

```
...
.pipe(postcss([ cssnano({
        preset: ['default', { 
               // reduceInitial: false
               reduceInitial: { ignore : ["pointer-events"] }
        }]
}) ]))
...
```

### Expected behaviour

I expect **pointer-events: initial** to be left untouched.

### Steps to reproduce

CSS:

.test {
   pointer-events: initial;
}

### Version

5.1.13

### Preset

default

### Environment

```shell
System:
    OS: Windows 10 10.0.19044
    CPU: (16) x64 AMD Ryzen 7 4800H with Radeon Graphics
    Memory: 53.05 GB / 63.42 GB
```

### Package details

```shell
+-- autoprefixer@10.4.12
| `-- postcss@8.4.17
+-- cssnano@5.1.13
| +-- cssnano-preset-default@5.2.12
| | +-- css-declaration-sorter@6.3.1
| | | `-- postcss@8.4.17 deduped
| | +-- cssnano-utils@3.1.0
| | | `-- postcss@8.4.17 deduped
| | +-- postcss-calc@8.2.4
| | | `-- postcss@8.4.17 deduped
| | +-- postcss-colormin@5.3.0
| | | `-- postcss@8.4.17 deduped
| | +-- postcss-convert-values@5.1.2
| | | `-- postcss@8.4.17 deduped
| | +-- postcss-discard-comments@5.1.2
| | | `-- postcss@8.4.17 deduped
| | +-- postcss-discard-duplicates@5.1.0
| | | `-- postcss@8.4.17 deduped
| | +-- postcss-discard-empty@5.1.1
| | | `-- postcss@8.4.17 deduped
| | +-- postcss-discard-overridden@5.1.0
| | | `-- postcss@8.4.17 deduped
| | +-- postcss-merge-longhand@5.1.6
| | | +-- postcss@8.4.17 deduped
| | | `-- stylehacks@5.1.0
| | |   `-- postcss@8.4.17 deduped
| | +-- postcss-merge-rules@5.1.2
| | | `-- postcss@8.4.17 deduped
| | +-- postcss-minify-font-values@5.1.0
[...]

PR #1440, opened 2022-10-23
chore(deps)!: upgrade svgo to v3

dargmuesli:
### Summary

Upgrade dependency svgo to v3 and increase the minimum supported node version to v14.

### Motivation

Keep up to date with the svgo upstream library.

- [x] Yes
- [ ] No

### Checklist

### Documentation
Please check if any of these apply:

- [ ] My proposed change require an update to the documentation
- [ ] I have added or updated the relevant documentation


dargmuesli:
This PR is meant to get work started to support svgo v3 and thus in draft state. I'm not sure what else needs to be done as I'm new to this library's source. Feel free to add or suggest necessary changes. I'm open to improvement suggestions!

packages/postcss-svgo/package.json
```diff
@@ -28,16 +28,15 @@
...
     "postcss-value-parser": "^4.2.0",
-    "svgo": "^2.7.0"
+    "svgo": "^3.0.0"
   },
   "bugs": {
     "url": "https://github.com/cssnano/cssnano/issues"
   },
   "engines": {
-    "node": "^10 || ^12 || >=14.0"
+    "node": ">=14.0"
```
dargmuesli, line 37: https://github.com/svg/svgo/releases/tag/v3.0.0 lists this as a breaking change.

packages/postcss-svgo/package.json
```diff
@@ -28,16 +28,15 @@
...
   },
   "devDependencies": {
-    "@types/svgo": "^2.6.2",
```
dargmuesli, line 40: Types are now included in the base library.

packages/postcss-svgo/src/index.js
```diff
@@ -43,12 +43,9 @@ function minifySVG(input, opts) {
   svg = svg.replace(escapedQuotes, '$1="$2"');
 
   const result = optimize(svg, opts);
-  if (result.error) {
```
dargmuesli, line 46: All errors are now thrown. I think this makes this error check unnecessary.

tsconfig.json
```diff
@@ -12,7 +12,8 @@
     "newLine": "lf",
     "allowSyntheticDefaultImports": true,
     "forceConsistentCasingInFileNames": true,
-    "strict": true
+    "strict": true,
+    "skipLibCheck": true
```
dargmuesli, line 16: Without this option, the `tsc` command would complain about typing inside `.ts` files inside the `node_modules` directory.

CharleeWa:
Thank you for this PR, I found the questions in the Nuxtjs.

When I add some lib on the Nuxt projects, The command line print this warning:

`stable has been deprecated`.

So, If this PR is merged, Can I give a PR to Nuxt.js?  

Thank you so much!

dargmuesli:
Hey @CharleeWa, thank you for your support.
I'm afraid, I don't know what "questions" and "some lib" is referring to. I also don't know where the warning originates from. Would you create a [minimal reproduction](https://stackblitz.com/github/nuxt/starter/tree/v3-stackblitz)?

Regarding Nuxt: `cssnano` is a subdependency of `@nuxt/vite-builder`, which is a subdependency of `nuxt`. If this PR is merged as a breaking change, another PR would be required for `@nuxt/vite-builder`.

ludofischer (commented):
Thank you for your PR! I think the only issue is that we would need to release a major version of postcss-svgo, cssnano, cssnano-preset-default and cssnano-preset-advanced. Updating to a major version can be inconvenient for those users who install cssnano as a transitive dependency. I've done a quick check :
1. `css-minimizer-webpack-plugin` (4371000 downloads) already requires Node.js 14.15, but the most downloaded version is still the 3.x which only requires Node.js 12.x 
2. Parcel: only requires Node.js 12.x but I think they use lightningcss by default
3. vue-cli: Only requires Node.js 12.x, but is deprecated

What do you think @alexander-akait ? Can we drop Node < 14 and release cssnano 6 in the near future? If we release 6 I would at least also get rid of the YAML parser, since it's only necessary for the (udocumented) YAML configuration feature which I think nobody is using (if they do they can just turn it into JSON)

ludofischer:
The postcss-svgo tests need to be updated because the result is slightly different but it looks equivalent (different attribute order and minified identifiers)

alexander-akait:
@ludofischer I think it is time to drop old nodes, so I am fine with the next v6

CharleeWa:
@dargmuesli 

Sorry, I wasn't specific enough. If you use the pnpm package manager, the installation dependency will have this prompt. Of course it's not a big deal, just a warning, but it doesn't look very comfortable. So, I think we should wait for cssnano to release a new version, nuxt will update cssnano's version, and the problem will not exist.

截屏2022-10-25 13 58 01

ludofischer:
Thank you for updating the tests. It might take a bit longer before we release 6.x  New cssnano releases take a long time to reach users, so I would like to do at least another 5.x release with the updates we've just merged on master. Maybe it's also possible to solve some merge-rules bugs on 5.x before we switch to 6.  
I'm also thinking we could use the 6 release to rewrite the configuration loading and parsing (see https://github.com/cssnano/cssnano/issues/1436 for an example where the current behaviour is a bit unintuitive).

tsconfig.json
TrySound: This is fixed in https://github.com/svg/svgo/releases/tag/v3.0.1

dargmuesli: Thank you for the notice!
It now gives
```
node_modules/.pnpm/svgo@3.0.1/node_modules/svgo/plugins/plugins-types.ts:1:28 - error TS7016: Could not find a declaration file for module 'csso'. '/mnt/data/WorkingDirectory/Git/dargmuesli/cssnano/node_modules/.pnpm/csso@5.0.5/node_modules/csso/cjs/index.cjs' implicitly has an 'any' type.
  Try `npm i --save-dev @types/csso` if it exists or add a new declaration (.d.ts) file containing `declare module 'csso';`

1 import type * as csso from 'csso';            
```
Is that something that should be fixed by cssnano or svgo?

TrySound: You need to add `@types/csso` to dev dependencies

dargmuesli:
I updated to v3.0.1 and reverted the `skipLibCheck` addition.

alexander-akait:
@dargmuesli can you rebase to solve pnpm merge

dargmuesli:
Rebased as requested ✅

zdm:
When this PR will be merged?

silkfire:
Bump, when can we merge this PR? @ludofischer

mahnunchik:
I'm looking forward to have it merged

dargmuesli:
Please upvote the PR by giving a :+1: on the first comment here and do not add "me too" comments, thanks!

zdm:
Oh, is is possible to merge it finally?
PLEASE.

packages/postcss-svgo/package.json
```diff
@@ -28,16 +28,16 @@
   "repository": "cssnano/cssnano",
   "dependencies": {
     "postcss-value-parser": "^4.2.0",
-    "svgo": "^2.7.0"
+    "svgo": "^3.0.1"
```
VIKTORVAV99, line 31: If we bump this to 3.0.2 we can get rid of the `@types/csso` package as well.
```suggestion
    "svgo": "^3.0.2"
```

packages/postcss-svgo/package.json
```diff
@@ -28,16 +28,16 @@
...
   "bugs": {
     "url": "https://github.com/cssnano/cssnano/issues"
   },
   "engines": {
-    "node": "^10 || ^12 || >=14.0"
+    "node": "^14 || ^16 || >= 18"
   },
   "devDependencies": {
-    "@types/svgo": "^2.6.2",
+    "@types/csso": "^5.0.0",
```
VIKTORVAV99, line 40: This can be removed now as they are no longer needed since v3.0.2 of SVGO.

VIKTORVAV99 (commented):
SVGO has released a new [patch version](https://github.com/svg/svgo/releases/tag/v3.0.2) so the csso types package should no longer be needed.

packages/postcss-svgo/package.json
dargmuesli: Removed :)

packages/postcss-svgo/package.json
dargmuesli: Updated! :rocket:

eddiemonge:
Considering Node14 is almost out of maintenance schedule, getting this in sooner would be much appreciated. I get moving slow with major updates, but considering users _do not have to upgrade unless they want to_, doing a major update shouldn't take so long or be so painful.

alexander-akait:
@ludofischer Maybe it is time to do major release and drop old nodejs support?

ludofischer:
What features would you want from SVGO 3? There seem to be at least one unsolved regression from SVGO 2: https://github.com/svg/svgo/issues/1746. I don't know how impactful these are though.
I was hoping to find the time to improve the configuration reading as I think some configuration options right now have no effect (as that would be another breaking change), but obviously it seems I never manage to get around to doing it. So we might as well release cssnano 6.0 just with SVGO bump, dropping the YAML support and possibly dropping normalize-url.

ludofischer (approved):
I think we can merge this.

PR #1444, opened 2022-10-26
chore: update browserslist and autoprefixer

ludofischer:
```
git diff --diff-algorithm=minimal --word-diff=color HEAD^
```
gives a good overview of the changes. Changes include removing the `-moz-test-decoration`, removing `-webkit-keyframes`.


ludofischer:
Converting to draft because there are other unexpected test failures besides the fixtures.

ludofischer:
The issue was that tests failed because they relied on IE 11 being part of the default browserslist query, and browserslist moved IE11 to 'dead' in a release 4.21. Also some `-moz` and `-webkit` prefix disappeared, but I did not try too hard to bring them back (I tried a few configurations but they seemed to result in an empty list of browsers).

package.json
```diff
@@ -60,6 +61,9 @@
     "ie9": [
       "IE 9"
     ],
+    "ie11": [
```
ludofischer, line 64: A test was using this environment, but it was not defined.

alexander-akait (approved):

PR #1446, opened 2022-10-27
fix: update initial property values

ludofischer:
This takes a long time to check.It makes no sense to compare with MDN, so for each property you need to check the spec. (I have opened a PR for some mistakes I found on MDN: https://github.com/mdn/data/pull/622)


packages/postcss-reduce-initial/src/data/fromInitial.json
```diff
@@ -12,16 +12,17 @@
...
   "azimuth": "center",
   "backdrop-filter": "none",
   "background-attachment": "scroll",
   "background-blend-mode": "normal",
   "background-image": "none",
   "background-position": "0% 0%",
-  "background-position-x": "left",
-  "background-position-y": "top",
+  "background-position-x": "0%",
+  "background-position-y": "0%",
```
alexander-akait, line 25: Good catch

alexander-akait (approved):

Issue #1000, opened 2021-02-23
[Bug] : Merging properties and rules with custom properties results in wrong border


hatsumatsu:
**Describe the bug**

Using short and long formats for border properties AND custom properties results wrong border rule. Seems related to but different from https://github.com/cssnano/cssnano/issues/864

**To Reproduce**

1. Use this CSS in the playground:

```css
:root {
	--border--width--thick: 10px;
}

.a {

/**
 * works without custom properties:
 * border-top: 10px solid blue;    
 */
    border-top: var(--border--width--thick) solid blue;    
    border-width: 1px;    
}

.a > .b {
	/**
     * works without custom properties:
	 * border-left: 10px solid blue;    
	 */
    border-left: var(--border--width--thick) solid blue;
    border-width: 1px;
}	
```
2. The result is: 

```css
:root{--border--width--thick:10px}.a{border-top:solid #00f}.a,.a>.b{border-width:1px}.a>.b{border-left:solid #00f}
```

3. See how `.b` should have a border width of 1px. Instead `border-left:solid #00f` is produced which falls back to a width of `medium`.

**Expected behavior**
Border rules width custom properties should be megred correctly.

**Desktop (please complete the following information):**
- CSSNANO Version 4.1.10
-   `{"preset" : "default"}`

**Additional context**
This bug occured using [nextjs](https://github.com/vercel/next.js) in which configuration to cssnano is not exposed. Looks like a broad user base is affected by this bug.

ludofischer:
Confirmed with 5.0.0-rc.2

ludofischer:
The specific error (with disappearance of the `var()`) seems to come from the between merge-rules and another plugin, since it does not reproduce with merge-rules alone. But I think the merge rules output is still wrong:
```css
.a { border-top:·var(--border-width-thick) solid blue; }
.a, .a > .b { border-width: 1px; }
.a > .b {border-left: var(--border-width-thick) solid blue; }

```
`border-left` on `.a > .b`  is now 10px when originally it was 1px.

ludofischer:
Seems fixed in cssnano 8.05, the border rules are not merged.

PR #1450, opened 2022-11-04
fix(merge-rules): avoid merging conflicting borders

ludofischer:
The previous logic assumed properties did not conflict if their
names were not equal but contained the same number of dash-separated parts.
This works for margin and padding but does not work for border,
because for example border-left and border-color can conflict.

This change prevents a valid optimization when the conflicting properties
re-declared the same value (for example border-width: 10px and
border: 10px solid blue), but that's the same as for other properties.

Related to #1000 and #701

I think a similar problem is going to happen for `place-items` and `align-items` or `justify-items` since they conflict but escape the current conflict detection algorithm (it assumes no conflict if the names property names start with a different word).


packages/postcss-merge-rules/src/index.js
```diff
@@ -229,22 +243,18 @@ function partialMerge(first, second) {
...
       .slice(indexOfDecl + 1)
       .filter((d) => isConflictingProp(d.prop, decl.prop));
-    if (!nextConflictInFirst.length) {
+    if (nextConflictInFirst.length === 0) {
       return true;
     }
     const nextConflictInIntersection = intersection
       .slice(intersectIndex + 1)
       .filter((d) => isConflictingProp(d.prop, decl.prop));
-    if (!nextConflictInIntersection.length) {
```
ludofischer, line 245: Next condition ensures that `nextConflictInIntersection.length !== 0` since already `nextConflictInFirst.length !== 0`, but  in fact it even does not matter, as long as the number of conflicts is the same we can still hope the merge will work.

packages/postcss-merge-rules/src/index.js
```diff
@@ -166,11 +167,24 @@ function isConflictingProp(propA, propB) {
...
     return false;
   }
+
   // Conflict if rest-count mismatches
   if (a.rest.length !== b.rest.length) {
     return true;
   }
 
+  /* Do not merge conflicting border properties */
+  if (a.base === 'border') {
```
ludofischer, line 177: I find this function uses some non-obvious terminology. `base` is the first part of the property name excluding the vendor prefix. `rest` all the rest of the name.
I could not find an alternative to using a special case, CSS syntax is just not that regular. I think  in fact `conflict if  count mismatches or if every part is the same` is only true for the `margin- ` and `padding-` properties.

alexander-akait (commented):
I am fine with it, but I am afraid it will not fix all possible cases, I think we need rewrite logic, I want to implement this in `swc` too, so in theory we can rewrite the whole logic together (it will be just different languages, but logic will be same)

ludofischer:
> I am fine with it, but I am afraid it will not fix all possible cases, I think we need rewrite logic, I want to implement this in `swc` too, so in theory we can rewrite the whole logic together (it will be just different languages, but logic will be same)

I would rather rewrite the logic when we drop Node 12 and 14 and release 6.0, this is fix for people who will remain on 5.x. Do you think there's a more general method to merge the rules? I think you can likely do more than merging only groups of adjacent two rules, but you still need to make special cases for the properties with different names but can override each other

ludofischer:
'Evil' case:
```css
.a { place-content: center; justify-content: start; } .b { justify-content: start; place-content: center; }
```

ludofischer:
I've added a fix for the `place-` `justify` `align-` conflict.

alexander-akait:
@ludofischer 

> I would rather rewrite the logic when we drop Node 12 and 14 and release 6.0, this is fix for people who will remain on 5.x. Do you think there's a more general method to merge the rules? I think you can likely do more than merging only groups of adjacent two rules, but you still need to make special cases for the properties with different names but can override each other

Yeah, I am fine with it after drop Node 12 and Node 14

> .a { place-content: center; justify-content: start; } .b { justify-content: start; place-content: center; }

I think we should `order` them before merge, and if reorder is `failed` (i.e. keeped) we just don't merge them, because I found our partial logic is rely on things which should not, for perf reason, we can `reorder` only potential to merge candidates

ludofischer:
> I think we should order them before merge, and if reorder is failed (i.e. keeped) we just don't merge them, because I found our partial logic is rely on things which should not, for perf reason, we can reorder only potential to merge candidates

Does this reorder logic exist some inside cssnano? What I did was detecting if one of the properties was place-*, then don't merge. https://github.com/cssnano/cssnano/pull/1450/commits/ea403e230e505183586aa7564fb9ee96b962bbb9#diff-106bd8df0f811c5c47fcb5841936b1a954ea692da68c1b246384bec798328d5fR168 From looking at the list of all shorthands, I think `place` is the only case where a property can override another property which starts with a different word. But yes, it is kind of a hack.

alexander-akait:
> Does this reorder logic exist some inside cssnano? What I did was detecting if one of the properties was place-*, then don't merge. 

`css-declaration-sorter` should do it, even if developer disable use it directly, we can create `fake` root with required properties and sort them and then check it


I think if we move it https://github.com/cssnano/cssnano/blob/master/packages/cssnano-preset-default/src/index.js#L124 before merge-rules we will fix most of cases

ludofischer:
> I think if we move it https://github.com/cssnano/cssnano/blob/master/packages/cssnano-preset-default/src/index.js#L124 before merge-rules we will fix most of cases

I see. What do you want to do with the current PR? Merge it for 5.x or just leave it out?

alexander-akait:
@ludofischer Let's add TODO/create an issue about it to not lose context and we can merge as patch to fast fix

ludofischer:
I have create an issue for a better fix as you've suggested: https://github.com/cssnano/cssnano/issues/1466 Do you think we can merge this? Then I can make a new release with the other fixes that have been contributed recently.

alexander-akait:
@ludofischer yeah, let's do it :+1:

alexander-akait (approved):

PR #1477, opened 2023-03-15
test: remove Node.js 10 and 12 from CI

ludofischer:
Remove unmaintained Node.js from CI. As we're not testing against Node.js 10 and 12 I think we need to bump the minimal supported Node version on all packages, because we cannot guarantee that we won't introduce an incompatibility in the future.

### DIFF
--- a/.changeset/forty-hornets-lie.md
+++ b/.changeset/forty-hornets-lie.md
@@ -1,0 +1,6 @@
+---
+'cssnano-preset-advanced': major
+'cssnano-preset-default': major
+---
+
+chore: bump node versions for packages depending on svgo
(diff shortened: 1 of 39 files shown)


ludofischer:
@alexander-akait I think you need to update the required checks on master if we want to get rid of the Node.js 10 and 12 tests because I haven't got access to that setting.

alexander-akait:
WIP

alexander-akait:
Done :+1:

alexander-akait (approved):

ludofischer:
> WIP

Are you sure you want to make 19 required? Then we'll need  to change the settings again in June when it becomes EOL and replaced by 20.

alexander-akait:
On the one hand, I agree that this is not necessary, on the other hand, we can catch all potential compatibility gaps (to be honest, I don’t think they will, but I have met very different situations), so let's keep it, if we will need to do release rapidly, we can ignore it and merge

PR #1478, opened 2023-03-15
chore: update TypeScript to 5.0.2

ludofischer:
I've updated to 5.0.2 directly since it's been released in the meantime. No changes except one parameter type change.

### DIFF
--- a/packages/postcss-merge-rules/src/index.js
+++ b/packages/postcss-merge-rules/src/index.js
@@ -307,7 +307,7 @@ function partialMerge(first, second) {
 
   /**
    * @param {function(import('postcss').Declaration):void} callback
-   * @this {import('postcss').Rule}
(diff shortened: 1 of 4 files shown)


packages/postcss-merge-rules/src/index.js
```diff
@@ -307,7 +307,7 @@ function partialMerge(first, second) {
 
   /**
    * @param {function(import('postcss').Declaration):void} callback
-   * @this {import('postcss').Rule}
+   * @this void
```
ludofischer, line 310: Now TypeScript sees it as void. I think it could be right since it's arrow functions all the way, so maybe `this` never gets bound, but I don't think it matters since `this` it's never used in the callbacks either.

alexander-akait (approved):

PR #1491, opened 2023-04-22
fix(postcss-reduce-idents): fix line names grid templates

### LINKED ISSUE #529: [Bug] named grid-template-columns are not renamed while grid-area does
version: 3.10.0
preset: default

input: 
```scss
.test{
  grid-template-columns: [image] 100px [info] auto;
}

.test a{
  grid-area: image;
}
```

cssnano output:
```scss
.test {
  grid-template-columns: [image] 100px [info] auto
}

.test a {
  grid-area: a
}
```

ludofischer:
Handle grid and column names enclosed in square brackets. Cover both the case where the brackets contain a single grid line name and the case where the brackets contain multiple separate grid line names.

Fix #529 (again)

### DIFF
--- a/packages/postcss-reduce-idents/src/lib/grid-template.js
+++ b/packages/postcss-reduce-idents/src/lib/grid-template.js
@@ -14,6 +14,8 @@ const RESERVED_KEYWORDS = new Set([
 const gridTemplateProperties = new Set([
   'grid-template',
   'grid-template-areas',
+  'grid-template-columns',
+  'grid-template-rows',
 ]);
 
 const gridChildProperties = new Set([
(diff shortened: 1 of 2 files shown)


packages/postcss-reduce-idents/src/lib/grid-template.js
```diff
@@ -53,6 +55,19 @@ module.exports = function () {
               }
             });
           }
+          if (child.type === 'word') {
+            const word = child.value;
+            if (word.startsWith('[') && word.endsWith(']')) {
```
ludofischer, line 60: Maybe there's a better way to do it, but postcss-value-parser does not understand list of elements in brackets at all.

ludofischer:
@alexander-akait do you think we can merge this?

alexander-akait (approved):

PR #1499, opened 2023-05-04
chore: replace Node.js 19 with Node.js 20 in CI

ludofischer:
Since Nodee 19 support will stop on June 1. Also update action and npm version.


alexander-akait (approved):

ludofischer:
@alexander-akait I think you need to change the required check from Node 19 to 20 before we can merge this. I don't have access to that setting.

alexander-akait:
@ludofischer Done :+1:

PR #1504, opened 2023-05-29
chore: update dev deps and workflow actions

ludofischer:


ludofischer:
I've raised an issue about this failure on the pnpm repo, since the pnpm-lock version update might not be compatible as intended https://github.com/pnpm/pnpm/issues/6609

alexander-akait (approved):

PR #1506, opened 2023-06-01
fix: replace lookbehind in regex for safari support

kojiwakayama:
Lookbehind assertions are not supported in Safari, so we need to modify the regular expression to achieve the same result without using lookbehind. The alternative approach uses capturing groups and backreferences.


alexander-akait (commented):
@ludofischer Do we officialy support our code without bundling/transpiling in a browser

Issue #1460, opened 2022-12-15
[Bug]: `reduceIdents: false` not working


argyleink:
### Describe the bug

I have some keyframes being transformed into a broken state:
```css
@keyframes author-scroll-effect {
  exit -5%, 0% {
    transform: scale(1);
  }
  exit 50%, exit 100% {
    transform: scale(0.8) translateY(10vh);
  }
}
```

becoming

```css
@keyframes author-scroll-effect {
  exit -5%, 0% {
    transform: scale(1);
  }
  exit 50%, exit to {
    transform: scale(0.8) translateY(10vh);
  }
}
```

I'm pretty savvy with build tools and nothing I did could remove that transform, makes me think it's got a bug? or let me know how I'm holding it wrong. it's strange too because i'm not passing a preset (well, i wasnt at first but i've tried passing presets) and the only way i can get rid of the transform is by using the "lite" preset. This is odd because the default preset shouldnt even include this transform..?

![Screen Shot 2022-12-15 at 3 29 31 PM](https://user-images.githubusercontent.com/1134620/207988462-892a4083-4cc7-45c9-b1ec-8661f53a921f.png)

I was hoping this simple config would do it, but i've tried all the presets and stuff.

```js
cssnano({
  reduceIdents: false
})

cssnano({
  reduceIdents: { keyframes: false }
})
```

### Expected behaviour

I expected being able to disable the transform but found I could only do it by heavily downgrading the optimizations to the lite preset. 

### Steps to reproduce

1. use latest node
2. use latest cssnano and postcss
3. try and remove the `reduceIdents` transform

### Version

latest

### Preset

default

### Environment

```shell
System:
    OS: macOS 12.6.1
    CPU: (16) x64 Intel(R) Xeon(R) W-2140B CPU @ 3.20GHz
    Memory: 476.42 MB / 64.00 GB
    Shell: 3.2.57 - /bin/bash
```

### Package details

```shell
cssnano and postcss
```

ludofischer:
If the broken transformation is `100%` changing to `to`, I don't think the `reduce-idents` plugin is responsible. `reduce-idents` should not touch anything besides custom identifiers (in this case, it should shorten the keyframe name), the culprit is likely a different plugin.

PR #1518, opened 2023-08-15
fix: prevent mangling of 'named timeline range names and percentages' in keyframe selectors

johannesodland:
This PR fixes an issue reported in https://github.com/cssnano/cssnano/issues/1460

Keyframe selectors from scroll-driven animations ([named timeline range name and percentage](https://www.w3.org/TR/scroll-animations-1/#named-range-keyframes)) are mangled by replacing `100%` with `to` in keyframe selectors where it is not valid.

```css

@keyframes test {
   entry 100% {
      color: red;
   }
}

/*
The keyframe-selector with a named timeline range name and percentage is mangled to:

@keyframes test {
   entry to {
      color: red;
   }
}
*/
```

This PR fixes the issue by only replacing `100%` with `to` when the parent selector only has one child node.

### DIFF
--- a/packages/postcss-minify-selectors/src/index.js
+++ b/packages/postcss-minify-selectors/src/index.js
@@ -163,7 +163,7 @@ const tagReplacements = new Map([
 function tag(selector) {
   const value = selector.value.toLowerCase();
 
-  if (tagReplacements.has(value)) {
+  if (tagReplacements.has(value) && selector.parent && selector.parent.nodes.length === 1) {
     selector.value = /** @type {string} */ (tagReplacements.get(value));
   }
 }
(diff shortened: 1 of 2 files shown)


packages/postcss-minify-selectors/src/index.js
```diff
@@ -163,7 +163,7 @@ const tagReplacements = new Map([
 function tag(selector) {
   const value = selector.value.toLowerCase();
 
-  if (tagReplacements.has(value)) {
+  if (tagReplacements.has(value) && selector.parent && selector.parent.nodes.length === 1) {
```
ludofischer, line 166: Apologies for the long delay in answering. It looks fine, but could you maybe write a comment explaining what the `selector.parent` etc. check is trying to achieve?

johannesodland: Sure. I refactored and added comments to clarify the handling of simple vs complex tag selectors. https://github.com/cssnano/cssnano/pull/1518/commits/e7596d65d03a74aafde86424322fec84712c10ca

I'm not sure if checking parent node length is the best way to differentiate between a complex or a simple selector. Do you know if there is a better way?

ludofischer (approved):

Issue #1519, opened 2023-08-18
[Bug]: Incorrect font-value after transform


smelukov:
### Describe the bug

Hi there! Just found a bug:

This declaration: `font: .8em "YS Text", Arial, Helvetica, sans-serif;` ✅ correct declaration 
Will be transformed into: `font:.8em YS Text,Arial,Helvetica,sans-serif` ✅ correct declaration 

But this one (just remove a whitespace after `em`): `font: .8em"YS Text", Arial, Helvetica, sans-serif;` ✅ it also a [correct declaration](https://github.com/cssnano/cssnano/assets/6654581/63aa9670-c822-48a0-838e-408cfa0cbb86)
![image](https://github.com/cssnano/cssnano/assets/6654581/ec4ecbe0-afa8-458b-b294-b98ef6c6f87e)
Will be transformed into: `font:.8em"YS Text"Arial,Helvetica,sans-serif` ❌ incorrect declaration (missing comma after `"YS Text"`)
![image](https://github.com/cssnano/cssnano/assets/6654581/fa41aeb1-ba1c-4e22-a480-9295b977ffcc)

I've fixed it in the PR (#1520)

### Expected behaviour

`font: .8em"YS Text", Arial, Helvetica, sans-serif;` should be transformed correctly

### Steps to reproduce

1. open https://cssnano.co/playground/
2. paste
```
body{
  font: .8em"YS Text", Arial, Helvetica, sans-serif;
}
```
into source-field
3. see incorrect declaration on result-field

### Version

6.0.1

### Preset

default

### Environment

```shell
Does not depend on the OS
```

### Package details

```shell
No packages
```

PR #1520, opened 2023-08-18
Add additional space before font-family if needed

smelukov:
`.8em"YS Text", Arial, Helvetica, sans-serif` => `.8em YS Text, Arial, Helvetica, sans-serif`

See #1519

### DIFF
--- a/packages/postcss-minify-font-values/src/lib/minify-font.js
+++ b/packages/postcss-minify-font-values/src/lib/minify-font.js
@@ -56,9 +56,31 @@ module.exports = function (nodes, opts) {
     }
   }
 
-  familyStart += 2;
+  let needAdditionalSpace = false;
 
-  family = minifyFamily(nodes.slice(familyStart), opts);
+  // handle case .8em"Times New Roman"
+  if (nodes[familyStart + 1].type === 'space') {
+    familyStart += 2;
+  } else {
+    // @ts-ignore
+    if (nodes[familyStart + 1].quote) {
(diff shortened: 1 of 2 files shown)


smelukov:
friendly ping ☺️

SedykhArtem (approved):

packages/postcss-minify-font-values/src/lib/minify-font.js
```diff
@@ -56,9 +56,31 @@ module.exports = function (nodes, opts) {
...
 
-  familyStart += 2;
+  let needAdditionalSpace = false;
 
-  family = minifyFamily(nodes.slice(familyStart), opts);
+  // handle case .8em"Times New Roman"
+  if (nodes[familyStart + 1].type === 'space') {
+    familyStart += 2;
+  } else {
+    // @ts-ignore
```
ludofischer, line 65: I don't think we use `ts-ignore` anywhere else. Could you find a way to solve this through an additional check or a TypeScript cast?

ludofischer (changes requested):
Thank you for your PR and sorry for taking so long to review it. It looks fine but could you solve the TypeScript issue?

packages/postcss-minify-font-values/src/lib/minify-font.js
ludofischer: The tests do not pass either, there's likely too many assumptions about the shape of the nodes here.

PR #1523, opened 2023-09-30
chore: update dev deps

ludofischer:


packages/postcss-colormin/types/index.d.ts
```diff
@@ -6,5 +6,5 @@ export = pluginCreator;
  */
 declare function pluginCreator(config?: Record<string, boolean>): import('postcss').Plugin;
 declare namespace pluginCreator {
-    const postcss: true;
+    let postcss: true;
```
ludofischer, line 9: Small change in TypeScript output, seems correct as we just assign this property to the `pluginCreator` object so it is in fact mutable.

ludofischer:
Hi @alexander-akait, in case you have time, could you review this PR and https://github.com/cssnano/cssnano/pull/1524?

PR #1533, opened 2023-12-02
deps: update to svgo v3.0.5 and update docs

SethFalco:
I've updated the documentation to use the SVGO v3 syntax, and upgraded to SVGO v3.0.5 which includes many bug fixes and a few more optimizations.

Consider reading the changelogs:

* https://github.com/svg/svgo/releases

Does the following:

* Update the documentation to use the SVGO v3 syntax, the original documentation was outdated.
* Update to SVGO v3.0.5 which has many bug fixes and some optimizations.
* Drop the types for the `plugin` property as SVGO provides its own already which is more elaborate.
* Replace the link to the SVGO plugins directory in the GitHub repo with the link to the [SVGO documentation](https://svgo.dev/).

## Chores

* Drop the Sticker Mule link from support page, as this link redirects to the homepage.


ludofischer:
There's one test failure in the css-size package.

SethFalco:
> There's one test failure in the css-size package.

@ludofischer Hmm, seems to be specific to Node.js v20.10.0, and is not related to the changes introduced by this PR.

In all previous versions until then, the test passes as-is. Only when run in Node.js v20.10.0 which is the latest LTS release, both in CI and locally, the result changes to `68 B`.

This is related to the Node.js version, though, and occurs in `master` and so should be affecting all PRs.

Not sure what the right way to approach this is tbh. 🤔

ludofischer (approved):

PR #1570, opened 2024-03-02
fix: prevent crash where first preset is already invoked

colinrotherham:
Thanks for such a brilliant utility 😊

I've fixed a couple of lines in [**packages/cssnano/src/index.js**](https://github.com/cssnano/cssnano/blob/a352df203eebff22f8b3b2ac458fe296474e9bf1/packages/cssnano/src/index.js#L36-L42) where arrays of presets are handled

See how the first preset in an array is assigned to `fn` and `options`?

https://github.com/cssnano/cssnano/blob/a352df203eebff22f8b3b2ac458fe296474e9bf1/packages/cssnano/src/index.js#L36-L42

The lines that follow use `preset` by mistake, rather than `fn`

```patch
+  if (preset.plugins) {
+    return preset.plugins;
-  if (fn.plugins) {
-    return fn.plugins;
   }
```

This affects the third example below but is fixed by this PR

## Single preset, already invoked

```js
cssnano({ // Works ✅
  preset: cssnanoPresetDefault({
    cssDeclarationSorter: false
  })
})
```

## Preset in array, separate config

```js
cssnano({ // Works ✅
  preset: [cssnanoPresetDefault, {
    cssDeclarationSorter: false
  }]
})
```

## Preset in array, already invoked

```js
cssnano({ // Error: Cannot load preset ❌
  preset: [
    cssnanoPresetDefault({
      cssDeclarationSorter: false
    })
  ]
})
```

### DIFF
--- a/packages/cssnano/src/index.js
+++ b/packages/cssnano/src/index.js
@@ -42,8 +42,8 @@ function resolvePreset(preset) {
   }
 
   // For JS setups where we invoked the preset already
-  if (preset.plugins) {
-    return preset.plugins;
+  if (fn.plugins) {
+    return fn.plugins;
   }
 
   // Provide an alias for the default preset, as it is built-in.
--- a/packages/cssnano/test/plugins_config.js
+++ b/packages/cssnano/test/plugins_config.js
@@ -28,6 +28,28 @@ test('should run the plugins in the preset', () => {
     });
 });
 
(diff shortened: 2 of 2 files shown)


ludofischer:
Thanks! It's rather extraordinary, I think that code has been in there for years, you must be the first person who tries the third configuration form.

alexander-akait (approved):

Issue #1513, opened 2023-07-24
[Bug]: browserslist config is not found in workspaces


anomiex:
### Describe the bug

When using workspaces, a browserslist config in a workspace is not used by cssnano. Only the browserslist config in the project root (if any) is used.

### Expected behaviour

When following the reproduction steps, I expect to see this output:
```
.foo{color:#00000026}
```
Instead I see
```
.foo{color:rgba(0,0,0,.15)}
```

### Steps to reproduce

1. Extract this zip file: [test.zip](https://github.com/cssnano/cssnano/files/12148819/test.zip)
2. Inside the extracted directory, run `npm install` or `pnpm install` or `yarn set version stable && yarn install`.
3. In the `packages/a/` subdirectory, run `npm exec postcss input.css` or `pnpm exec postcss input.css` or `yarn exec postcss input.css`

### Version

6.0.1

### Preset

default

### Environment

```shell
System:
    OS: Linux 6.3 Debian GNU/Linux trixie/sid
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
    Memory: 24.65 GB / 31.13 GB
    Container: Yes
    Shell: 5.2.15 - /bin/bash
```

### Package details

```shell
test@ /tmp/test
└─┬ a@1.0.0 -> ./packages/a
  ├─┬ cssnano@6.0.1
  │ ├─┬ cssnano-preset-default@6.0.1
  │ │ ├─┬ css-declaration-sorter@6.4.1
  │ │ │ └── postcss@8.4.27 deduped
  │ │ ├─┬ cssnano-utils@4.0.0
  │ │ │ └── postcss@8.4.27 deduped
  │ │ ├─┬ postcss-calc@9.0.1
  │ │ │ └── postcss@8.4.27 deduped
  │ │ ├─┬ postcss-colormin@6.0.0
  │ │ │ └── postcss@8.4.27 deduped
  │ │ ├─┬ postcss-convert-values@6.0.0
  │ │ │ └── postcss@8.4.27 deduped
  │ │ ├─┬ postcss-discard-comments@6.0.0
  │ │ │ └── postcss@8.4.27 deduped
  │ │ ├─┬ postcss-discard-duplicates@6.0.0
  │ │ │ └── postcss@8.4.27 deduped
  │ │ ├─┬ postcss-discard-empty@6.0.0
  │ │ │ └── postcss@8.4.27 deduped
[...]

PR #1571, opened 2024-03-02
fix: add missing type declarations to plugins with options

colinrotherham:
While looking at https://github.com/cssnano/cssnano/pull/1570 and https://github.com/alphagov/govuk-frontend/issues/4810 I noticed that plugin type definitions needed updating

This PR fixes a few missing things:

1. Plugin [`postcss-convert-values`](https://www.npmjs.com/package/postcss-convert-values) option `precision` should be optional
1. Plugin [`postcss-colormin`](https://www.npmjs.com/package/postcss-colormin) option `colormin` (via [`cssnano-preset-default`](https://www.npmjs.com/package/cssnano-preset-default)) is typed without property names
1. Plugin [`postcss-reduce-initial`](https://www.npmjs.com/package/postcss-reduce-initial) option `reduceInitial` (via [`cssnano-preset-default`](https://www.npmjs.com/package/cssnano-preset-default)) is not typed
1. Plugin [`css-declaration-sorter`](https://www.npmjs.com/package/css-declaration-sorter) option `cssDeclarationSorter` (via [`cssnano-preset-default`](https://www.npmjs.com/package/cssnano-preset-default)) is not typed
1. Plugins are not typed with Browserslist `path` and PostCSS context `env` where supported

These were spotted whilst fixing https://github.com/cssnano/cssnano/issues/1513 but I'll open another PR

## After

All plugin options are now included

Type definitions newly updated

### DIFF
--- a/packages/cssnano-preset-advanced/package.json
+++ b/packages/cssnano-preset-advanced/package.json
@@ -12,6 +12,7 @@
   "license": "MIT",
   "dependencies": {
     "autoprefixer": "^10.4.17",
+    "browserslist": "^4.23.0",
     "cssnano-preset-default": "workspace:^",
     "postcss-discard-unused": "workspace:^",
     "postcss-merge-idents": "workspace:^",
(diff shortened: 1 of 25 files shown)


colinrotherham:
I'll probably drop commit https://github.com/cssnano/cssnano/pull/1571/commits/ef1cb540057caa1be6550a22e3cf4ab8bc74e524 with those new defaults from `pnpm@8`

I was trying to prevent `tsc -b` outputting `import('../package1/node_modules/package2'` sometimes

alexander-akait (approved):

PR #1572, opened 2024-03-03
fix: tests looping `toInitial.json` data should run

colinrotherham:
This PR fixes the `Promise.all()` calls in **packages/postcss-reduce-initial/test/index.js** so they resolve

https://github.com/cssnano/cssnano/blob/c4be0f52a776da12caf0d37e86fc0c194d384e49/packages/postcss-reduce-initial/test/index.js#L20-L25

Two property tests for `'writing-mode'` and `'transform-box'` failed but these were handled in:

* https://github.com/cssnano/cssnano/pull/929
* https://github.com/cssnano/cssnano/pull/1242


colinrotherham:
@ludofischer Rebased to resolve the conflict 👍

ludofischer (approved):

PR #1573, opened 2024-03-04
feat: add preset and plugin options for browserslist

colinrotherham:
This PR fixes https://github.com/cssnano/cssnano/issues/1513 and https://github.com/alphagov/govuk-frontend/issues/4810

All plugins with Browserslist now support:

1. Add new options `overrideBrowserslist` and `path`
2. Detection of [PostCSS `from` context](https://github.com/postcss/postcss-load-config?tab=readme-ov-file#options) to find the `.browserslistrc` config
3. Detection of [webpack **postcss-loader** `file` context](https://webpack.js.org/loaders/postcss-loader/) to find the `.browserslistrc` config

## Preset options
Presets can now be configured with these options too (some plugins already could) to match Autoprefixer

```js
module.exports = {
  plugins: [ 
    cssnano({
      preset: cssnanoPresetDefault({
        env: 'development',
        overrideBrowserslist: 'supports es6-module',
        path: join(__dirname, 'path/to/config/.browserslistrc'),
        stats: {}
      })
    })
  ]
}
```


colinrotherham:
Thanks for merging https://github.com/cssnano/cssnano/pull/1570 and https://github.com/cssnano/cssnano/pull/1571 🙌

I've given this PR a rebase but commit https://github.com/cssnano/cssnano/pull/1573/commits/2967a504429d6604218cf69c1c46099e59008a52 will stay temporarily should #1572 be merged

packages/cssnano-preset-advanced/src/index.js
```diff
@@ -21,24 +22,51 @@ const autoprefixer = require('autoprefixer');
...
+      // Skip unsupported Browserslist "my stats" strings etc
+      // https://github.com/browserslist/browserslist/pull/237
+      stats:
+        typeof sharedOpts.stats !== 'string'
+          ? sharedOpts.stats // Autoprefixer supports stats object only
+          : undefined,
+    },
+  };
+
+  return deepmerge(defaults, opts);
```
ludofischer, line 59: Is it possible to avoid adding a dependency? Since it's just used once, we can probably code a solution for this particular case without needing a generic package.

tsconfig.json
```diff
@@ -7,6 +7,7 @@
     "allowJs": true,
     "checkJs": true,
     "declaration": true,
+    "declarationMap": true,
```
ludofischer, line 10: Good, I did not know about this option.

ludofischer (dismissed):
This looks fine. You seem to use the types and know about it more than I do. Did passing the autoprefixer options always work or is it a new feature?

packages/postcss-merge-rules/src/index.js
```diff
@@ -408,26 +409,30 @@ function selectorMerger(browsers, compatibilityCache) {
...
-      const browsers = browserslist(null, {
-        stats: resultOpts.stats,
-        path: __dirname,
-        env: resultOpts.env,
+      const { stats, env, from, file } = result.opts || {};
+      const browsers = browserslist(opts.overrideBrowserslist, {
+        stats: opts.stats || stats,
+        path: opts.path || dirname(from || file || __filename),
+        env: opts.env || env,
+        throwOnMissing: true,
```
colinrotherham, line 435: This was left in for testing, I'm going to remove it 🤦‍♂️

packages/cssnano-preset-advanced/test/browserslist/.browserslistrc
```diff
@@ -0,0 +1,12 @@
+defaults
+
+[legacy]
+IE <=11
+Chrome <=51
```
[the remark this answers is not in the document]
colinrotherham: Yeah I did, it was to find a nice support gap to test [postcss-colormin](https://www.npmjs.com/package/postcss-colormin) and [postcss-reduce-initial](https://www.npmjs.com/package/postcss-reduce-initial) since it was hard to find ranges that would clearly flip between `[modern]` and `[legacy]` environments etc

E.g. Chrome 47 to 61 via [**Browserslist** `supports css-placeholder-shown and not supports css-rrggbbaa`](https://browsersl.ist/#q=supports+css-placeholder-shown+and+not+supports+css-rrggbbaa)

I can't remember the exact query now, sorry

colinrotherham:
Thanks for the approval @ludofischer

I've made [one last push](https://github.com/cssnano/cssnano/compare/ecb5b4623076bcf3942d0266421bbb388ee16776..6cf152c0a2d72cd55345078583270162e4ddaf57) to address https://github.com/cssnano/cssnano/pull/1573#pullrequestreview-1915532061

colinrotherham:
>Did passing the autoprefixer options always work or is it a new feature?

@ludofischer The Browserslist `stats` or `env` options aren't new, but we're now applying them consistently

For example, these three only had it in plugin `opts`:

* Plugin `postcss-colormin` via `pluginCreator(opts)` only
* Plugin `postcss-convert-values` via `pluginCreator(opts)` only
* Plugin `postcss-minify-params` via `pluginCreator(opts)` only

With the remaining four only via PostCSS `result.opts`:

* Plugin `postcss-merge-rules` via PostCSS `result.opts` only
* Plugin `postcss-normalize-unicode` via PostCSS `result.opts` only
* Plugin `postcss-reduce-initial` via PostCSS `result.opts` only
* Plugin `stylehacks` via PostCSS `result.opts` only

There were probably scenarios where users configured plugins but the options didn't get used

## Test helpers

Your [test helpers](https://github.com/cssnano/cssnano/blob/master/util/testHelpers.js#L33-L38) also seem to follow the older PostCSS behaviour before [process options were split out](https://github.com/postcss/postcss/releases/tag/6.0.0) where the plugin options are merged with the process options

I'm not 100% sure this matches real world behaviour with tools like [postcss-load-config](https://www.npmjs.com/package/postcss-load-config) or [postcss-loader](https://www.npmjs.com/package/postcss-loader)

So for consistency, all plugins that support Browserlist will now:

• Apply options when found in PostCSS `result.opts`
• Optionally override options when found in PostCSS plugin `opts`

This follows the approach used by Autoprefixer:

```js
plugins: [
  autoprefixer({
    env: 'modern'
  })
]
```

So that Browserslist options can be overridden at plugin-level too:

```js
preset: cssnanoPresetDefault({
  env: 'modern'
})
```

Might be nice in future to loop `result.processor.plugins` and grab the Autoprefixer options from there too

packages/postcss-colormin/test/index.js
```diff
@@ -235,14 +238,62 @@ test(
...
   processCSS('h1{color:rgba(100% 50% 0% / 50%)}', 'h1{color:#ff800080}', {
-    env: 'chrome62',
+    overrideBrowserslist: 'Chrome 62',
   })
 );
 
 test(
   'should convert long color to 4-digit hex when supported',
   processCSS('h1{color:hsla(0 100% 50% / 40%)}', 'h1{color:#f006}', {
-    env: 'chrome62',
```
ludofischer, line 245: So `env` did not change meaning, it's just that `chrome62` is not an environment anymore because of the changes to the Browserslist configuration  in the tests, right?

packages/postcss-ordered-values/types/rules/grid.d.ts.map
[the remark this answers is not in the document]
colinrotherham: Yes please, would it help to split them out into another PR?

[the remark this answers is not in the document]
ludofischer: Yes, I think its better to leave the source maps in a different PR, because it is a fairly mechanical change but it affects too many files unrelated to this PR.

packages/postcss-colormin/test/index.js
colinrotherham: I've used this option to add test coverage for going straight to Browserslist. It skips both the **.browserslist** config and **package.json** environments

Using `from` (files) or `path` (directories) will locate the nearest config though

packages/postcss-ordered-values/types/rules/grid.d.ts.map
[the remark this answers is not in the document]
colinrotherham: Dropped the commit, will open a PR at some point. Thanks @ludofischer

ludofischer (approved):

ludofischer:
I've merged this, if you want you can open a PR with the source map changes.

PR #1574, opened 2024-03-05
chore: remove css-size from repository

ludofischer:
The package is not part of cssnano, has little functionality and there is little sign that it has any real users.


ludofischer:
See the discussion https://github.com/cssnano/cssnano/discussions/1566

PR #1575, opened 2024-03-05
feat: enable “go to definition” via declaration maps

colinrotherham:
This PR adds declaration maps to link types back to their original sources

E.g. Using "Go to Definition" to jump to the original source code rather than a `*.d.ts` file

>[Declaration Map - `declarationMap`](https://www.typescriptlang.org/tsconfig#declarationMap)
>Generates a source map for `.d.ts` files which map back to the original `.ts` source file. This will allow editors such as VS Code to go to the original `.ts` file when using features like _Go to Definition_.
>
>You should strongly consider turning this on if you’re using project references.

This feature was most useful whilst working on recent Browserslist changes in:

* https://github.com/cssnano/cssnano/pull/1573


ludofischer (approved):
This adds about 120 files to the package, I guess they're really small though.

PR #1650, opened 2024-07-20
Better coverage reports

cenfun:
Better coverage reports with `c8 --experimental-monocart`

- console-details
image

- v8 (coverage/index.html)
Native v8 coverage report
image

- codecov (coverage/codecov.json)
Native codecov format, better than lcov

### DIFF
--- a/.c8rc.json
+++ b/.c8rc.json
@@ -5,5 +5,5 @@
     "packages/postcss-reduce-initial/script/lib/mdnCssProps.mjs"
   ],
   "all": true,
-  "reporter": ["text", "lcovonly"]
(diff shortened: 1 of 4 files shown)


ludofischer:
Hello, thank you for your PR. At this moment I am not sure I would want to merge it. I I have got a few questions:
1. Why are these coverage reports better?
2. What does `experimental` mean? Are we going to need to update it more frequently?
3. I was hoping to get rid of c8 at some point and just use the built-in Node.js coverage reporter if it becomes stable. Is this going to cause a problem?

cenfun:
Thank you let me know it. If the built-in Nodejs coverage reporter is the next plan, I think we can close this PR.

Issue #1660, opened 2024-08-14
[Bug]: `CanvasText` is not the same as `initial`, should not be replaced


krokodok:
### What should be improved?

I am trying to make a website more accessible for Windows High Contrast Mode / forced-colors media query. I want to style my elements with the system colors from CSS Color Module Level 4: https://www.w3.org/TR/css-color-4/#css-system-colors

But currently my keywords get transformed to `initial`, which is not the same. (But not all of them?)

```css
// Input
.my-element {
    color: CanvasText;
    background: Canvas;
}

// Output
.my-element {
    color: initial;
    background: Canvas;
}
```

Currently I am relying on a workaround using a custom property:
```css
.my-element {
    --CanvasText: CanvasText;
    color: var( --CanvasText );
    background: Canvas;
}
```

### Describe the solution you would like

It would be great for the list of CSS Color Module Level 4 System Color keywords to be always respected and not transformed to `initial`.

### Additional context

I am not too familiar with cssnano. There might be a way to already do this, or an easier workaround. But I could not find any.

### Are you willing to work on this?

- [X] Yes, I would like to help

krokodok:
It seems that this is a problem in `postcss-reduce-initial`. I will open a new issue there.

ludofischer:
We can keep this open since `postcss-reduce-initial` is in this same repository.

krokodok:
Never mind! I used an outdated version of the package…

krokodok:
Sorry! I was confused with my setup! Now I found the issue again!

It is here in this line:
https://github.com/cssnano/cssnano/blob/104c3035a99af1a23fd53b315d105fdd8df030e3/packages/postcss-reduce-initial/src/data/toInitial.json#L18

`canvastext` is _not_ the same as `initial`. An easy example is, if I want to change the styling of an element of which the initial color is something different than `canvastext`, for example a `mark` element:

```css
// Remove mark highlight in forced-colors mode
@media (forced-colors: active) {
    mark {
        background: Canvas;
        color: CanvasText;
    }
}
```

If `Canvas` is black and the initial color for `mark` is black my text won't be readable.

krokodok:
Won't fix, see discussion in PR: https://github.com/cssnano/cssnano/pull/1661

PR #1661, opened 2024-08-14
Remove conversion of `canvastext` to `initial`

krokodok:
Fixes: https://github.com/cssnano/cssnano/issues/1660

### DIFF
--- a/packages/postcss-reduce-initial/src/data/toInitial.json
+++ b/packages/postcss-reduce-initial/src/data/toInitial.json
@@ -15,7 +15,6 @@
   "border-right-color": "currentcolor",
   "border-top-color": "currentcolor",
   "box-sizing": "content-box",
-  "color": "canvastext",
   "column-rule-color": "currentcolor",
   "font-synthesis": "weight style small-caps position",
   "image-orientation": "from-image",


ludofischer:
Hello, thank you for your pull request. I would prefer that we try first to fix it upstream before changing it here. The data in `toInitial.json` is fetched on build from the  [MDN data ](https://github.com/mdn/data) repository. The initial value for color comes from this line in their repository  https://github.com/mdn/data/blob/ab6bf8a2eb9bd8123aa025ba6c8758eaeb0ca2ef/css/properties.json#L4119 . Since you have already researched the spec, would you have time to open an issue or a PR on https://github.com/mdn/data ? You can probably do a better a job of explaining why they should change it than I can, because I am not familiar with this aspect of the spec.

ludofischer:
Currently, it's the color initial value is also canvastext on the MDN website, https://developer.mozilla.org/en-US/docs/Web/CSS/color, so changing it in MDN data would help avoid confusion in the future.

krokodok:
Thank you for your quick reply! I will follow this up over there!

krokodok:
So this seems to be an edge case. It is correct, that `initial` defaults to `canvastext`, but on Chrome there is a browser stylesheet overlay that sets `color` only for `<mark>` elements to `marktext`. I will close this for now and not pursue it further.

PR #1688, opened 2025-03-13
Add support for selector order preservation to postcss-minify-selectors

ezzak:
feat(minify-selectors): Add the ability to disable sorting when minifying selectors

### DIFF
--- a/packages/postcss-minify-selectors/src/index.js
+++ b/packages/postcss-minify-selectors/src/index.js
@@ -197,11 +197,15 @@ const reducers = new Map(
   ])
 );
 
+/** @typedef {object} Options
+ *  @property {boolean=} preserveOrder
+ */
 /**
  * @type {import('postcss').PluginCreator<void>}
+ * @param {Options} opts
  * @return {import('postcss').Plugin}
  */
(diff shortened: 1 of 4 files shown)


packages/postcss-minify-selectors/src/index.js
```diff
@@ -233,7 +237,9 @@ function pluginCreator() {
             }
           }
         });
-        selectors.nodes.sort();
+        if (opts.preserveOrder !== true) {
```
alexander-akait, line 240: Please avoid using `no`/`preserve`/etc, mostly such names are very misleading, let's use `sort: boolean`, by default `true` and when it is `false` we will do nothing

alexander-akait (changes requested):

alexander-akait (dismissed):

alexander-akait:
Looks like some tests are failed, can you look?

ezzak:
Tests/presets were broken because by default an empty option object is passed to plugins, and default preset was not passing `{sort: true}`. Fixed.

ludofischer (approved):

Issue #1694, opened 2025-05-02
[Bug]: Invalid initial-value for custom CSS percentage property


aramikuto:
### Describe the bug

The `initial-value` of `0%` in a custom CSS percentage property is incorrectly minified to `0`, dropping the percent sign. The value `0` without the percent sign is invalid for properties defined with `<percentage>` syntax.

This issue occurs with both the default and advanced profiles. The lite profile correctly preserves the percent sign.

Input:
```css
@property --ratio {
  syntax: "<percentage>";
  inherits: false;
  initial-value: 0%;
}
```

Output:
```css
@property --ratio{syntax:"<percentage>";inherits:false;initial-value:0}
```

### Expected behaviour

The percent sign should be preserved in the minified output:
```css
@property --ratio{syntax:"<percentage>";inherits:false;initial-value:0%}
```

### Steps to reproduce

1. Open the [sniped in the online playground](https://cssnano.github.io/cssnano/playground/#eyJpbnB1dCI6IkBwcm9wZXJ0eSAtLXJhdGlvIHtcbiAgc3ludGF4OiBcIjxwZXJjZW50YWdlPlwiO1xuICBpbmhlcml0czogZmFsc2U7XG4gIGluaXRpYWwtdmFsdWU6IDAlO1xufSIsImNvbmZpZyI6ImNzc25hbm8tcHJlc2V0LWRlZmF1bHQifQ==).
2. Minify using the default profile.

### Version

7.0.6

### Preset

default

### Environment

```shell
System:
  OS: macOS 15.4.1
  CPU: (8) arm64 Apple M2
  Memory: 152.61 MB / 24.00 GB
  Shell: 5.9 - /bin/zsh
```

### Package details

```shell
├─┬ cssnano@7.0.6
│ ├─┬ cssnano-preset-default@7.0.6
│ │ ├─┬ css-declaration-sorter@7.2.0
│ │ │ └── postcss@8.5.3 deduped
│ │ ├─┬ cssnano-utils@5.0.0
│ │ │ └── postcss@8.5.3 deduped
│ │ ├─┬ postcss-calc@10.1.1
│ │ │ └── postcss@8.5.3 deduped
│ │ ├─┬ postcss-colormin@7.0.2
│ │ │ └── postcss@8.5.3 deduped
│ │ ├─┬ postcss-convert-values@7.0.4
│ │ │ └── postcss@8.5.3 deduped
│ │ ├─┬ postcss-discard-comments@7.0.3
[...]

PR #1695, opened 2025-05-05
fix(postcss-convert-values): preserve percent sign in percentage values in at-rules with double quotes 

aramikuto:
Fixes https://github.com/cssnano/cssnano/issues/1694

The `shouldKeepZeroUnit` check previously expected the syntax value `<percentage>` to be enclosed only in single quotes. However, in CSS, string values can also be surrounded by double quotes. I've added an additional check to handle both quote styles and updated the test cases to include the double-quoted variant.

NOTE: I didn't change any logic in `packages/postcss-minify-selectors/test/index.js`, but the file was automatically updated after running the `pnpm fixlint` command.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -72,3 +72,4 @@ Thanks goes to these wonderful people:
 - Xiaojun Zhou
 - Seth Falco
 - Seiya
+- Aleksandr Kondrashov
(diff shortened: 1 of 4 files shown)


alexander-akait (approved):

alexander-akait:
@ludofischer Thank you

PR #1700, opened 2025-05-10
chore: add `packageManager` field & move pnpm setting

btea:
To ease management and avoid conflicts with npm, pnpm supports moving all configurations to the pnpm-workspace.yaml file. For more [github.com/orgs/pnpm/discussions/9037](https://github.com/orgs/pnpm/discussions/9037)

Add the `packageManager` field to facilitate users to automatically switch to the specified version of pnpm when installing dependencies locally to prevent unnecessary conflicts.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,5 +34,6 @@
     "production": [
       "IE 11 and last 2 versions"
     ]
-  }
+  },
+  "packageManager": "pnpm@10.10.0"
 }
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -4,7 +4,7 @@ permissions:
 on:
   release:
     types: [published]
-    
+
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -34,5 +34,6 @@
     "production": [
       "IE 11 and last 2 versions"
     ]
-  }
+  },
+  "packageManager": "pnpm@10.10.0"
```
ludofischer, line 38: What would happen if the user as the wrong pnpm installed globally?

btea: When the user executes `pnpm install`, pnpm will switch to the pnpm version specified by packageManager field.

btea: ref https://pnpm.io/settings#managepackagemanagerversions. https://github.com/pnpm/pnpm/pull/8363

ludofischer: Don't we also need to change the pnpm action in the GitHub workflows then so that it automatically detect the pnpm version?

btea: Yes, please take a look at the documentation. https://github.com/pnpm/action-setup#inputs

ludofischer (approved):

PR #1701, opened 2025-05-10
ci: use Node.js 24

btea:
[nodejs/node@v24.0.0 (release)](https://github.com/nodejs/node/releases/tag/v24.0.0)


ludofischer (approved):
Thank you

PR #1702, opened 2025-05-12
fix(postcss-convert-values): preserve percent sign in length-percentage syntax

Multiply:
Currently there's a check to preserve the percent sign for `syntax: "<percentage>"`, this simply adds `<length-percentage>` as well. ([MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/@property/syntax#length-percentage) reference)

While debugging https://github.com/tailwindlabs/tailwindcss/issues/17977 I found https://github.com/cssnano/cssnano/pull/1695 which I initially thought would fix my underlying issue, until I noticed that I am having the issue with `<length-percentage>` and not `<percentage>`.


ludofischer (approved):
Thanks!

Issue #1719, opened 2025-08-20
[Bug]: CSS unit stripped after build in the`linear()` timing function causes it to break


cernymatej:
### Describe the bug

I noticed that after transformation, the `%` unit in CSS gets stripped from all values that are `0`, which causes the `linear()` timing function to break, because it always expects a percentage on the right side.

```css
.example {
  transition-timing-function: linear(0 0%, 1 100%);
}
```

<img width="341" height="46" alt="Image" src="https://github.com/user-attachments/assets/4c4773e8-25c4-4c88-ab33-9697c9d685a5" />

### Expected behaviour

I would expect it to either keep the `0%` or transform the argument to a single `0` -> `linear(0, 1 100%)`

### Steps to reproduce

transform the following code and see that the `%` unit gets stripped

```css
.example {
  transition-timing-function: linear(0 0%, 1 100%);
}
```

### Version

7.1.0

### Preset

(no preset)

### Environment

```shell
System:
    OS: macOS 15.6
    CPU: (12) arm64 Apple M2 Max
    Memory: 1.12 GB / 96.00 GB
    Shell: 5.9 - /bin/zsh
```

### Package details

```shell
nuxt 4.0.3
└─┬ @nuxt/vite-builder 4.0.3
  └── cssnano 7.1.0
```

### Additional context

Following up https://github.com/nuxt/nuxt/issues/33014

PR #1720, opened 2025-08-20
fix(convert-values): exclude `linear()` from stripping `%` from value 0

cernymatej:
fix https://github.com/cssnano/cssnano/issues/1719

This fixes a problem where the `%` was stripped from 0 values in the `linear()` function.
I also added a test for it.


ludofischer:
> @ludofischer Can you do a patch release, thank you

Done!

Issue #1488, opened 2023-04-07
[Bug]: 


meduzen:
### Describe the bug

Leading zero is not stripped from `cubic-bezier` functions assigned to custom properties.

Maybe there’s some proximity with #1456.

### Expected behaviour

The following

```css

.a { animation-timing-function: cubic-bezier(0.485,0.045,0.545,0.97); }
.b { --cp: cubic-bezier(0.485,0.045,0.545,0.97); }
.c { --gr: rgb(0, 0, 0); }
```

minifies to

```css
.a{animation-timing-function:cubic-bezier(.485,.045,.545,.97)}.b{--cp:cubic-bezier(0.485,0.045,0.545,0.97)}.c{--gr:#000}
```

We see that `animation-timing-function` allows `cubic-bezier` coordinates to be minified, while the same value assigned to a custom property isn’t.

```css
animation-timing-function:cubic-bezier(.485,.045,.545,.97); /* leading zero removed ✅ */
--cp:cubic-bezier(0.485,0.045,0.545,0.97); /* leading zero not removed ❌ */
```

My expectation for assigning such a timing-function to a custom property would be that it should be minified.

### Steps to reproduce

[CSS Nano playground link](https://cssnano.co/playground/#eyJpbnB1dCI6IlxuLmEgeyBhbmltYXRpb24tdGltaW5nLWZ1bmN0aW9uOiBjdWJpYy1iZXppZXIoMC40ODUsMC4wNDUsMC41NDUsMC45Nyk7IH1cbi5iIHsgLS1jcDogY3ViaWMtYmV6aWVyKDAuNDg1LDAuMDQ1LDAuNTQ1LDAuOTcpOyB9XG4uYyB7IC0tZ3I6IHJnYigwLCAwLCAwKTsgfSIsImNvbmZpZyI6ImNzc25hbm8tcHJlc2V0LWFkdmFuY2VkIn0=)

### Version

6.0.0

### Preset

(no preset)

### Environment

```shell
System:
    OS: macOS 13.2.1
    CPU: no, it’s connected to hamsters
    Memory: -
    Shell: -
```

### Package details

```shell
├─┬ cssnano@6.0.0
│ └── postcss@8.4.21
└─┬ vite@4.2.1
  └── postcss@8.4.21
```

ludofischer:
[...]

PR #1767, opened 2026-04-02
fix(postcss-convert-values): add transformCustomProperties option

dkryaklin:
`postcss-convert-values` currently skips all custom property declarations to avoid unsafe transformations. However, some users explicitly want value conversion inside custom properties (e.g. `500ms` → `.5s`, `0.485` → `.485`).

Adds a `transformCustomProperties` option (default: `false` to preserve existing behavior). Set to `true` to enable value conversions inside custom property declarations.

```js
cssnano({ preset: ['default', { convertValues: { transformCustomProperties: true } }] })
```

Relates to #1488.


dkryaklin:
Reopening from a clean single-commit branch.

PR #1768, opened 2026-04-02
fix(postcss-colormin): add transformCustomProperties option

### LINKED ISSUE #1456: [Bug]: cssnano transformations are unsafe in custom property values
### Describe the bug

In cssnano compilation modes that seek to guarantee exact behavioral equivalence between the input and output CSS, transformations on custom property values should be substantially limited relative to what they are now. Custom properties aren't just used for configuring values elsewhere in the stylesheet; they may also be used to communicate information from the stylesheet to JavaScript (as in Polymer.js for example), and so the details of their JS serialization is semantically relevant. cssnano currently modifies these values like any other property, which can change the serialization and thus the semantics of the stylesheet.

There are two broad categories of issue here:

1. The existence of whitespace tokens is visible to JS serialization, so `--foo: bar` and `--foo:bar` are meaningfully different. cssnano will convert the former to the latter.

2. There's no guarantee that a custom property's value will be parsed in a context where a CSS color is meaningful, so `--foo: rgb(0 0 0)` and `--foo: #000` are meaningfully different. cssnano will convert the former to the latter.

The second issue covers essentially all value-level cssnano optimizations, almost none of which are safe to perform on a custom property value.

### Expected behaviour

There should at least be a way to tell cssnano not to make changes that could affect the semantics of the compiled CSS—which is to say, to disable essentially all value-level conversions for custom properties unless they don't affect the way those properties are serialized to JS values. Ideally, this configuration would be enabled in `cssnano-preset-default`, since that advertises itself as making no assumptions about how the CSS will be used.

### Steps to reproduce

Copy/paste this into https://cssnano.co/playground/:

```css
a {
  --foo: rgb(0 0 0);
}
```

### Version

5.1.14

### Preset

default

### Environment

```shell
cssnano playground
```

### Package details

```shell
cssnano playground
```

### Additional context

_No response_

dkryaklin:
Custom property values may be read by JavaScript (e.g. via `getComputedStyle`), where `rgb(0 0 0)` and `#000` are not equivalent. By default, `postcss-colormin` transforms colors inside custom properties, which can break such use cases.

Adds a `transformCustomProperties` option (default: `true` to preserve existing behavior). Set to `false` to skip color minification inside custom property declarations.

```js
cssnano({ preset: ['default', { colormin: { transformCustomProperties: false } }] })
```

Fixes #1456.

### DIFF
--- a/packages/postcss-colormin/src/index.js
+++ b/packages/postcss-colormin/src/index.js
@@ -108,6 +108,7 @@ function addPluginDefaults(options, browsers) {
  * @property {boolean} [hsl]
  * @property {boolean} [name]
  * @property {boolean} [transparent]
+ * @property {boolean} [transformCustomProperties] Whether to minify colors inside custom property values (default: true)
  */
 
 /**
@@ -147,6 +148,10 @@ function pluginCreator(config = {}) {
                 decl.prop
               )
             ) {
+              return;
+            }
+
+            if (config.transformCustomProperties === false && decl.prop.startsWith('--')) {
               return;
             }
 
--- a/packages/postcss-colormin/test/index.js
+++ b/packages/postcss-colormin/test/index.js
@@ -231,6 +231,16 @@ test(
 );
 
 test(
+  'should minify colors inside custom properties by default',
+  processCSS('a{--foo:rgb(0 0 0)}', 'a{--foo:#000}')
+);
+
+test(
+  'should not minify colors inside custom properties when transformCustomProperties is false',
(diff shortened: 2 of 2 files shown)


ludofischer (changes requested):
Seems fine. but it does not completely fix the linked issue, which asks also that whitespace be preserved.

ludofischer (approved):

dkryaklin:
@ludofischer thanks!

PR #1770, opened 2026-04-02
Move css-declaration-sorter from "default" to "advanced"

### LINKED ISSUE #1764: [Feature Request]: remove cssDeclarationSorter from the default presets
### What should be improved?

`cssnano` default presets include the `cssDeclarationSorter` optimization. This optimization is great to ensure that properties are sorted in a way that makes compression algorithms more efficient.

https://cssnano.github.io/cssnano/docs/optimisations/cssdeclarationsorter/

However, this optimization is "dangerous" and can change how a page is rendered.

I recently found 2 bugs in the optimization's repo:
- https://github.com/Siilwyn/css-declaration-sorter/pull/506
- https://github.com/Siilwyn/css-declaration-sorter/pull/505

Although those PRs solve the immediate issues, they highlight a problem with the optimization: **css-declaration-sorter is not future-proof** (this isn't a problem with the repo, but a problem with how the CSS spec changes)

For example, if the CSS spec changes once again, and a new property "conflicts" with existing properties, then cssnano will begin breaking styles until the optimization mappings get updated.

(here, "conflict" means that two separate shorthand properties have intersecting longhand properties)

### Describe the solution you would like

I think the optimization is great to have in the "advanced" preset. I just don't think it should be enabled by default due to its possibility to break styles in the future.

### Possible alternatives

_No response_

### Additional context

enough context was added above, but happy to answer questions.

If you agree, I can make this configuration change

### Are you willing to work on this?

- [x] Yes, I would like to help

Goodwine:
Fixes #1764


ludofischer (changes requested):
Colud you add a changeset that describes what you did (run `pnpm changeset` in the repository)?

packages/cssnano-preset-default/src/index.js
```diff
@@ -128,9 +126,6 @@ function configurePlugins(plugins, opts = {}) {
     reduceInitial: {
       ...sharedProps,
     },
-    cssDeclarationSorter: {
```
ludofischer, line 131: Specifically, I think the removal of the option in the `default-preset` should appear in the release notes, at least it might help someone who might be confused by the output change.

Goodwine:
I ran the command, but I'm confused by the output. It created a `.changeset` directory with a file. However this directory isn't tracked in the repo. Is this intended? am I missing something? I don't think I can publish the changes myself 🤔 

Is this what you were looking for?

(new file: `.changeset/pink-maps-remain.md`

ludofischer:
> Is this what you were looking for?

Yes, this is how it works.

.changeset/pink-maps-remain.md
```diff
@@ -0,0 +1,18 @@
+---
+'cssnano-preset-advanced': major
+'cssnano-preset-default': major
```
ludofischer, line 3: This is going to cause major update of cssnano for this change? Do we really want that? Technically the types for `cssnano-preset-default` would break, but I think we did similar changes in the past without releasing a major.

packages/cssnano-preset-default/README.md
```diff
@@ -96,18 +95,6 @@ Note that you may wish to publish your own preset to npm for reusability, should
 
 ## Plugins
 
-### [`css-declaration-sorter`](https://github.com/Siilwyn/css-declaration-sorter) (external)
```
ludofischer, line 99: Should this be moved to the README of `cssnano-preset-advanced`?

ludofischer (changes requested):

packages/cssnano-preset-default/README.md
Goodwine: Sorry for the delay, I got sidetracked with other projects.

`cssnano-preset-advanced` already had this doc. It was dupped across both

.changeset/pink-maps-remain.md
Goodwine: I believe the diff output will be sufficiently different to justify calling this a major change, but ultimately it is up to you.

Let me know how you want to proceed.

ludofischer: Yeah, I think ultimately it might be worth a major, since we could also drop some older Nodejs versions at the same time (at least Node.js 18)

ludofischer (approved):

ludofischer:
Solved the conflicts and merged in https://github.com/cssnano/cssnano/pull/1793

PR #1775, opened 2026-04-18
feat(postcss-minify-selectors): fold selector lists into :is() (#1703)

### LINKED ISSUE #1703: [Feature Request]: Optionally use :is()
### What should be improved?

https://developer.mozilla.org/en-US/docs/Web/CSS/:is#simplifying_list_selectors

### Describe the solution you would like

Use `:is()`,

### Possible alternatives

_No response_

### Additional context

--

### Are you willing to work on this?

- [ ] Yes, I would like to help

dkryaklin:
Closes #1703.

## What it solves

Generated CSS (SCSS `@extend`, Tailwind, etc.) often produces long comma-separated selector lists with a shared prefix or suffix. Collapsing them with `:is()` cuts bytes:

```css
/* Before */
section h1, article h1, aside h1, nav h1 { font-size: 25px }

/* After */
:is(article,aside,nav,section) h1{font-size:25px}
```

## How it works

`postcss-minify-selectors` gets a new `convertToIs` pass (default on). For each rule it tokenises each selector into alternating compound/combinator pieces, finds the longest common prefix and suffix (snapped to combinator boundaries), and emits `<prefix>:is(<middles>)<suffix>`.

The rewrite only fires when:

- the result is strictly shorter;
- every divergent middle has the same CSS specificity (`:is()` takes the max, so mixing e.g. `.a` and `#x` would silently shift the cascade);
- no pseudo-element or nesting `&` appears inside the folded part;
- the browserslist target supports `css-matches-pseudo`.

Opt out with `convertToIs: false` on `postcss-minify-selectors` or the `minifySelectors` preset option.

## Byte savings on popular frameworks

Measured by running the repo's framework fixtures through the default preset with a modern browserslist target:

| Framework      | Before  | After   |       Saved |  % |
| -------------- | ------- | ------- | ----------- | -- |
| Bootstrap 3    | 119 959 | 115 726 |  −4 233 B   | −3.53% |
| Bulma          | 168 213 | 162 626 |  −5 587 B   | −3.32% |
| Schema         |  55 925 |  54 812 |  −1 113 B   | −1.99% |
| Foundation 6   | 115 879 | 113 655 |  −2 224 B   | −1.92% |
| Semantic UI    | 617 968 | 606 104 | −11 864 B   | −1.92% |
| Kube           |  30 661 |  30 100 |    −561 B   | −1.83% |
| Bootstrap 4    | 151 794 | 149 263 |  −2 531 B   | −1.67% |
| Hydrogen       |  14 520 |  14 335 |    −185 B   | −1.27% |
| UIKit 3        | 246 781 | 244 351 |  −2 430 B   | −0.98% |
| Vital          |  22 845 |  22 664 |    −181 B   | −0.79% |

Savings scale roughly with how SCSS-heavy a framework's generated output is; 1–3.5% is the realistic band for real-world projects.


ludofischer (approved):
Very impressive!

ludofischer:
@dkryaklin Your code seems to show more than average experience with processing CSS. Have you got any suggestions to update https://github.com/postcss/postcss-calc ? Its current parser is very out of date with the spec, which causes a lot of bug reports, but I am not sure what the best solution is (rewrite the parser by hand or use some existing library?)

dkryaklin:
> @dkryaklin Your code seems to show more than average experience with processing CSS. Have you got any suggestions to update https://github.com/postcss/postcss-calc ? Its current parser is very out of date with the spec, which causes a lot of bug reports, but I am not sure what the best solution is (rewrite the parser by hand or use some existing library?)

Thanks. Let me have a look at current postcss-calc code. I will share my thoughs 👍

dkryaklin:
I would suggest rewrite parser and use Pratt Parsing approach. It feels exactly what we need and easily extensible. There are few articles in the internet about it

https://matklad.github.io/2020/04/13/simple-but-powerful-pratt-parsing.html
https://www.crockford.com/javascript/tdop/tdop.html

Then we can add everything from the specs. That probably will require major release tho

ludofischer:
> I would suggest rewrite parser and use Pratt Parsing approach. It feels exactly what we need and easily extensible. There are few articles in the internet about it

Rewriting the parser by hand also seemed the best option when I searched for something (2 years ago!). Possibly https://github.com/csstools/postcss-plugins/tree/main/packages/css-calc could be used, but at the time the maintainer said it was not really suitable for a minifier https://github.com/csstools/postcss-plugins/discussions/1450

dkryaklin:
> > I would suggest rewrite parser and use Pratt Parsing approach. It feels exactly what we need and easily extensible. There are few articles in the internet about it
> 
> Rewriting the parser by hand also seemed the best option when I searched for something (2 years ago!). Possibly https://github.com/csstools/postcss-plugins/tree/main/packages/css-calc could be used, but at the time the maintainer said it was not really suitable for a minifier [csstools/postcss-plugins#1450](https://github.com/csstools/postcss-plugins/discussions/1450)

https://github.com/postcss/postcss-calc/pull/240
Here is early raw draft PR. I will continue progress over next weeks. But you can have a look early and provide some feedback about overall direction

PR #1777, opened 2026-04-20
chore: improve CI security

ludofischer:
Pin actions by hash and fix zizmor warnings

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,12 +12,13 @@ jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v6.0.2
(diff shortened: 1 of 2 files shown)


alexander-akait (approved):

.github/workflows/test.yml
```diff
@@ -65,6 +66,6 @@ jobs:
 
       - name: Publish code coverage report
         if: matrix.os == 'ubuntu-latest' && matrix.node-version == '24'
-        uses: codecov/codecov-action@v3.1.4
+        uses: codecov/codecov-action@v6.0.0
```
ludofischer, line 69: The codecov action fails to download (or fails to download the github-scripts) action when you try to pin it.

PR #1782, opened 2026-04-20
docs: update website for release

ludofischer:


ludofischer:
@alexander-akait codecov never completes now because the new version requires a token, but I  cannot add it because I don't have permission to configure the repository GitHub apps. If you want to keep Codecov you would need to configure it yourself. I've also noticed the Azure pipelines app is still installed but looks unused now, but I haven't got the permissions to remove it either.

Issue #1784, opened 2026-04-29
[Bug]: `:is()` rewrite is firing even with divergent CSS specificity in the middle segments


philipstarkey:
### Describe the bug

In #1775, the change to fold selector lists into `:is()` appears to be firing for me even when the middle segments have divergent specificity. 

### Expected behaviour

Given that #1775 stated

> The rewrite only fires when:
> - ...
> - every divergent middle has the same CSS specificity (:is() takes the max, so mixing e.g. .a and #x would silently shift the cascade);

I would have expected not to see this rewrite.

### Steps to reproduce

I don't have a _minimal_ reproducible example on hand because my CSS is defined in a Vue SFC and going through some layers of compilation, but in Vue I have:

```
.field-input.field-input-default :deep(input),
.field-input.field-input-default :deep(button[role="combobox"]),
.field-input.field-input-default :deep(textarea),
.field-input.field-input-default :deep(select) {
  ...
}
```

On v7.1.5, this would compile to:
```
.field-input.field-input-default[data-v-01b440c5] button[role=combobox],
.field-input.field-input-default[data-v-01b440c5] input,
.field-input.field-input-default[data-v-01b440c5] select,
.field-input.field-input-default[data-v-01b440c5] textarea {
  ...
}
```
where the `[data-v-...]` is injected by the Vue compiler to scope component CSS styles to specific Vue components.

With cssnano v7.1.7, this now produces:
```
.field-input.field-input-default[data-v-d8c645b5] :is(button[role=combobox], input, select, textarea) {
  ...
}
```

This increases the specificity of the `input`/`select`/`textarea` cases to `(0,4,1)` instead of `(0,3,1)` due to the specificity of `button[role=combobox]`. Similar issues apply if you use other selectors like `div[data-my-attr]` or `button[data-my-attr]`.

### Version

7.1.7

### Preset

default

### Environment

```shell
System:
    OS: Linux 6.19 Debian GNU/Linux 12 (bookworm) 12 (bookworm)
    CPU: (8) arm64 unknown
    Memory: 20.20 GB / 25.06 GB
    Container: Yes
    Shell: 5.2.15 - /bin/bash
```

### Package details

```shell
cssnano@7.1.7
postcss@8.5.10
```

### Additional context

Vue 3.5.33 but I'm pretty sure Vue is irrelevant to the issue beyond the general format of the scoped component style CSS

dkryaklin:
Hi!

[...]

PR #1785, opened 2026-04-29
fix(postcss-minify-selectors): reject :is() fold for :nth-child(... of S) to preserve cascade

dkryaklin:
The `:is()` fold was incorrectly folding `:nth-child(... of S)` selectors with mismatched `of` specificities, silently raising the cascade specificity of every middle to the max:

  ```css
  /* Before: folds and shifts cascade */
  .x :nth-child(2n of #a) y, .x :nth-child(2n of .b) y, ... { ... }
  /* After: leaves as-is */
  ```

  Reject the fold for any compound containing :nth-child(... of …) / :nth-last-child(... of …), since the parser flattens An+B and of S into one node and we can't recover S's specificity to compare. Plain :nth-child(2n+1) still folds.


alexander-akait (dismissed):
/cc @ludofischer can you take a look too, thanks

packages/postcss-minify-selectors/test/foldToIs.unit.js
```diff
@@ -0,0 +1,639 @@
...
+    n.rawSpaceBefore = '';
+    n.rawSpaceAfter = '';
+    const v = n.value.trim();
+    n.value = v.length ? v : ' ';
+  });
+  return root;
+};
+
+// ---------------------------------------------------------------------------
+// specificityOf — exhaustive per-branch coverage
```
ludofischer: Maybe nitpicking, but how do we know this coverage is really exhaustive? Also it does not look like branch coverage, but more testing different types of selectors (element, id, pseudo-element etc).

packages/postcss-minify-selectors/src/lib/foldToIs.js
```diff
@@ -306,3 +340,13 @@ function tryFold(root) {
 }
 
 module.exports = tryFold;
+module.exports._internal = {
```
ludofischer: I think this pattern of an `internal` object on exports. is not used anywhere else in the codebase. It is also not clear to me what `internal` means. It looks like it's supposed to be just used in tests. In that case, could we not split these functions in a different file? Or just add them to the exports?

packages/postcss-minify-selectors/test/foldToIs.unit.js
```diff
@@ -0,0 +1,639 @@
...
+});
+
+test('universal in middle — folds when all are spec 0', () => {
+  // Prefix `* `, middles diverge but all are tag → spec (0,0,1). Folds.
+  const out = tryFold(parseRoot('* a,* b,* c,* d'));
+  assert.equal(out, '* :is(a,b,c,d)');
+});
+
+test('universal-only middle (all `*`) — no divergent middle, rejects', () => {
+  // After dedup, only one unique middle remains → middleStrs.length < 2.
```
ludofischer: What is `middleStrs`? I think if these comments are necessary, these should be assertion messages instead of comments

packages/postcss-minify-selectors/test/foldToIs.unit.js
```diff
@@ -0,0 +1,639 @@
...
+
+test('tryFold: :is(.a) vs :is(#x) differ in spec — rejects', () => {
+  assert.equal(
+    tryFold(parseRoot('.x :is(.a) .y,.x :is(#z) .y,.x :is(.b) .y,.x :is(.c) .y')),
+    null
+  );
+});
+
+// ---------------------------------------------------------------------------
+// Edge-case coverage from review checklist
```
ludofischer: Where is the 'review checklist' this is referring to?

packages/postcss-minify-selectors/test/foldToIs.unit.js
```diff
@@ -0,0 +1,639 @@
...
+
+test('custom-property pseudo-classes (:--state) fold (same spec)', () => {
+  const out = tryFold(
+    parseRoot('.x :--open,.x :--closed,.x :--hover,.x :--focus')
+  );
+  assert.equal(out, '.x :is(:--open,:--closed,:--hover,:--focus)');
+});
+
+test('mixed: tag vs *|tag (namespaced wildcard) — same spec, folds', () => {
+  // Both contribute type=1; same spec.
```
ludofischer: I would write specificity in full, initially I thought this meant "specification".
```suggestion
  // Both contribute type=1; same specificity.
```

packages/postcss-minify-selectors/test/foldToIs.unit.js
```diff
@@ -0,0 +1,639 @@
...
+      '.x [xlink|href],.x [xlink|title],.x [xlink|role],.x [xlink|type]'
+    )
+  );
+  assert.equal(
+    out,
+    '.x :is([xlink|href],[xlink|title],[xlink|role],[xlink|type])'
+  );
+});
+
+test('custom-property pseudo-classes (:--state) fold (same spec)', () => {
```
ludofischer:
```suggestion
test('fold custom-property pseudo-classes (same specificity)', () => {
```

packages/postcss-minify-selectors/test/foldToIs.unit.js
```diff
@@ -0,0 +1,639 @@
...
+  assert.equal(out, '* :is(a,b,c,d)');
+});
+
+test('universal-only middle (all `*`) — no divergent middle, rejects', () => {
+  // After dedup, only one unique middle remains → middleStrs.length < 2.
+  assert.equal(tryFold(parseRoot('.a *,.b *,.c *,.d *')), ':is(.a,.b,.c,.d) *');
+});
+
+test('selector starting with combinator (nesting context)', () => {
+  // 4-entry version is exactly equal length so guard rejects; need 6+ to
```
ludofischer: I find the comment is unclear. It should not talk about rejecting if the fold happens.

packages/postcss-minify-selectors/test/foldToIs.unit.js
```diff
@@ -0,0 +1,639 @@
...
+  const out = tryFold(
+    parseRoot(
+      '.x :nth-child(2n of #a) y,.x :nth-child(2n of .b) y,' +
+        '.x :nth-child(2n of .c) y,.x :nth-child(2n of .d) y'
+    )
+  );
+  assert.equal(out, null);
+});
+
+test(':nth-child(2n+1) without `of` clause still folds normally', () => {
```
ludofischer: I suppose there is no 'abnormal' folding, so normally confuses me.
```suggestion
test(':nth-child(2n+1) without `of` clause still folds', () => {
```

ludofischer (changes requested):
Is this also supposed to fix #1784 ? It's nice that there are more tests, but I find that the test naming and description hard to follow. It's true that existing tests are not amazing either in that regard, but I have left some suggestions on how to improve things. Was any code coverage tool not included in the repository used to  determine which tests to generate. In that case I think it might be interesting to see if we can add it to the repository. Also if you have some successful agent instructions or test setup that works well with agents it could be interesting to add in in another PR.

packages/postcss-minify-selectors/src/lib/foldToIs.js
dkryaklin: Fair. Refactored

dkryaklin:
Hi @ludofischer !

First of all, this is not a fix for https://github.com/cssnano/cssnano/issues/1784. I wasn't able to reproduce it yet. So will create separate PR once know what is wrong.

Secondly, sorry, it was not very clean PR in terms of test names and comments. Hope now its better.
For agent promps/skills I don't use anything extra. Just my personal view/opinion. But this time probably had to spend more time on cleaning up.

So regarding testing techniques: I was exploring option with [stryker mutation testing](https://stryker-mutator.io/) This is why some comments were off. The comments were referencing context that wasn't actually in the PR.

I can provide separate PR for stryker if it's something interesting to you.

packages/postcss-minify-selectors/src/lib/foldToIsHelpers.js
```diff
@@ -0,0 +1,233 @@
...
+    const s = specificityOf(child.nodes);
+    if (compareSpecificity(s, best) > 0) {
+      best = s;
+    }
+  }
+  return best;
+}
+
+/**
+ * @param {Token[]} middle
```
ludofischer, line 170: Would it be possible to add to the comment what this function is doing and some description of the parameters? What is `middle`?

packages/postcss-minify-selectors/src/lib/foldToIsHelpers.js
```diff
@@ -0,0 +1,233 @@
...
+function compareSpecificity(a, b) {
+  return a[0] - b[0] || a[1] - b[1] || a[2] - b[2];
+}
+
+/**
+ * @param {Specificity} a
+ * @param {Specificity} b
+ * @return {boolean}
+ */
+function sameSpecificity(a, b) {
```
ludofischer: I think this should be `equal` instead of `same`.

packages/postcss-minify-selectors/test/foldToIs.test.js
```diff
@@ -2,8 +2,8 @@
 const { test } = require('node:test');
```
ludofischer, line 2: The file naming convention `xxx.unit.js` is not used anywhere else. Maybe rename this file.

ludofischer (commented):
Thank you for your updates. I find it less surprising now, but I have left some new change requests.

packages/postcss-minify-selectors/src/lib/foldToIsHelpers.js
dkryaklin: e.g. for selector list:

`.x .a .y, .x .b .y, .x .c .y`

The fold tokenises each selector into compound + combinator tokens, then finds the longest common edges:

shared prefix: [.x, ' ']
shared suffix: [' ', .y]
middles: [.a], [.b], [.c] one per selector

packages/postcss-minify-selectors/src/lib/foldToIsHelpers.js
```diff
@@ -0,0 +1,225 @@
+'use strict';
+
+/**
+ * @typedef {object} Token
+ * @property {'compound'|'combinator'} kind
+ * @property {string} str
+ * @property {import('postcss-selector-parser').Node[]} [nodes]
+ */
+
+/** @typedef {[number, number, number]} Specificity */
```
ludofischer, line 10: Since the number of elements in specificity is always 'id, cls, type', wouldn't it be better to make `Specificity`  an object `{id: number, cls: number, type: number}`? What do we gain with an array?

ludofischer (approved):

packages/postcss-minify-selectors/src/lib/foldToIsHelpers.js
dkryaklin: > Since the number of elements in specificity is always 'id, cls, type', wouldn't it be better to make `Specificity` an object `{id: number, cls: number, type: number}`? What do we gain with an array?

No much. Just spec notation is a tuple (A, B, C) and our [id, cls, type] mirrors that in some way

PR #1799, opened 2026-05-08
fix(postcss-minify-selectors): tighten :is() fold acceptance to a strict allowlist

dkryaklin:
Replaces the fold's denylist with a strict allowlist: class / id / plain tag / plain attribute (no namespace, no case flag), plus :hover, :focus, :active, :visited, :link. Anything else: namespaced selectors, attribute case flags, structural / form / modern / functional pseudos is rejected


packages/postcss-minify-selectors/test/foldToIs.test.js
```diff
@@ -283,94 +283,76 @@ test('hasNthChildOfClause: deeply nested is true', () => {
...
 
-test('hasUnknownPseudoWithArgs: combinator is false', () => {
-  assert.equal(
-    hasUnknownPseudoWithArgs({ kind: 'combinator', str: '>' }),
-    false
-  );
-});
-
-test('hasUnknownPseudoWithArgs: plain compound is false', () => {
-  assert.equal(hasUnknownPseudoWithArgs(compoundOf('div.foo')), false);
```
ludofischer, line 294: Did the test for `div.foo` disappear?

ludofischer (approved):
Thanks!

packages/postcss-minify-selectors/test/foldToIs.test.js
dkryaklin: > Did the test for `div.foo` disappear?

It was reshaped into other tests

PR #1810, opened 2026-05-24
perf: speed up default preset by ~35% on framework corpus

dkryaklin:


packages/postcss-colormin/src/index.js
```diff
@@ -161,18 +161,20 @@ function pluginCreator(config = {}) {
...
             }
 
-            const cacheKey = JSON.stringify({ value, options, browsers });
-
-            if (cache.has(cacheKey)) {
-              decl.value = cache.get(cacheKey);
+            // options and browsers are pass-invariant (set once in
+            // prepare()), so the per-decl cache key only needs to be the
+            // value itself. The prior JSON.stringify allocated a key string
+            // per declaration to no benefit.
```
ludofischer, line 167: Stringification likely happened because the cache used to be an object instad of a map, so the keys at the time could only be strings or symbols.

packages/postcss-discard-duplicates/src/index.js
```diff
@@ -12,7 +12,13 @@ function trimValue(value) {
  * @return {boolean}
  */
 function empty(node) {
-  return !node.nodes.filter((child) => child.type !== 'comment').length;
+  const nodes = node.nodes;
```
ludofischer, line 15: Does this change cause an observable speedup? Anyway I prefer the new version because I think it is more readable.

packages/postcss-discard-duplicates/src/index.js
```diff
@@ -83,72 +89,97 @@ function equals(nodeA, nodeB) {
...
  * @return {void}
  */
-function dedupeNode(last, nodes) {
-  let index = nodes.includes(last) ? nodes.indexOf(last) - 1 : nodes.length - 1;
-
-  while (index >= 0) {
-    const node = nodes[index--];
-    if (node && equals(node, last)) {
-      node.remove();
+function dedupeFrom(node, targetNodes) {
```
ludofischer, line 96: Same as above, does rewriting the loop measurably improve performance? It could be more readable, but right now I cannot understand what either the old version or the new version is supposed to be doing. I would suggest to make sure these functions are tested and that the behaviour matches the old one. Since you are renaming the functions and the arguments, you could also try to make it more meaningful. I don't think `dedupeFrom` is an improvement compared to `dedupeNode`.

packages/postcss-merge-rules/src/index.js
```diff
@@ -144,27 +153,33 @@ function isDeclaration(node) {
...
-  if (ruleMeta && rule) {
-    let meta = ruleMeta.get(rule);
-    if (!meta && rule.nodes) {
-      meta = {
-        selectors: rule.selectors,
-        declarations: rule.nodes.filter(isDeclaration),
-        dirty: false,
-      };
-      ruleMeta.set(rule, meta);
+  let meta = ruleMeta.get(rule);
```
ludofischer, line 160: Why is it not necessary anymore to check if `ruleMeta` is defined?

util/bench.mjs
```diff
@@ -0,0 +1,183 @@
+// Baseline benchmark harness for cssnano.
```
ludofischer, line 1: Could we just use something like hyperfine to benchmark instead of a custom script? As far as I can see this is not giving timing per plugin or code path either.

ludofischer (commented):
Interesting PR, I would suggest splitting stuff that are relatively straightfoward refactorings, like rewriting loops, in a separate PR.

dkryaklin:
@ludofischer Thanks for feedback! I will update in next couple days

packages/postcss-discard-duplicates/src/index.js
```diff
@@ -83,72 +89,97 @@ function equals(nodeA, nodeB) {
...
+      }
+    }
+  }
+
+  // Recurse-then-dispatch order preserves atrule equals() semantics: an earlier
+  // atrule must still see its (un-recursed) interior when compared at this level.
   let index = nodes.length - 1;
   while (index >= 0) {
-    let last = nodes[index--];
+    const last = nodes[index--];
```
ludofischer, line 140: Also good change, but probably would classify it as a  _style_ change rather than performance.

packages/postcss-discard-duplicates/src/index.js
```diff
@@ -83,72 +89,97 @@ function equals(nodeA, nodeB) {
...
       continue;
     }
-    dedupe(last);
+    if (last.type === 'rule' || last.type === 'atrule') {
+      dedupe(last);
+    }
+
     if (last.type === 'rule') {
-      dedupeRule(last, nodes);
+      const group = ruleGroups && ruleGroups.get(last.selector);
```
ludofischer, line 149: What happened here to `dedupeRule`? Has it been inlined? Is inlining supposed to speed up  execution?

packages/postcss-merge-longhand/src/index.js
```diff
@@ -10,10 +12,69 @@ function pluginCreator() {
...
       css.walkRules((rule) => {
-        processors.forEach((p) => {
-          p.explode(rule);
-          p.merge(rule);
-        });
+        // Most rules only touch one (or zero) of the four prop families this
+        // plugin handles, so pre-scan once and skip the processors that have
+        // nothing to do — saves their walkDecls + getDecls passes.
+        const nodes = rule.nodes;
+        let mask = 0;
```
ludofischer, line 19: Does replacing `mask` with 4 booleans (e.g. `processBorders, `processMargins`) not achieve a very similar speedup and avoid unusual bit manipulation (at least I rarely see this in JavaScript code)

packages/postcss-merge-longhand/src/index.js
```diff
@@ -10,10 +12,69 @@ function pluginCreator() {
...
+        for (let i = 0; i < nodes.length; i++) {
+          const node = nodes[i];
+          if (node.type !== 'decl') {
+            continue;
+          }
+          const prop = node.prop;
+          // Most CSS in the wild has lowercase props, so the
+          // case-sensitive startsWith short-circuit avoids the toLowerCase
+          // allocation on the common path; the case-insensitive fallback
+          // only runs when the first byte already mismatches.
```
ludofischer, line 29: I think you could achieve a similar result by doing something like

```javascript
const lowerCaseProp = node.prop.toLowerCase();
```
Then you do not need to perform the check into every `if` clause.

ludofischer (changes requested):

ludofischer:
> @ludofischer Thanks for feedback! I will update in next couple days

I assume this is LLM-assisted so I wonder if the LLM can use a profiler output to guide the suggested optimization (aparenty you can generate `.cpuprofile` and `.heapprofile` files from the `node` cli by passing some options.

ludofischer:
You could also look into how Lightningcss did their benchmark https://github.com/parcel-bundler/lightningcss#benchmarks

dkryaklin:
I've split this PR into 3 batches and added tooling as separate PR

PR #1812, opened 2026-06-08
chore(util): add benchmark + cpuprofile tooling

dkryaklin:
## What

Adds the benchmarking and CPU-profiling tooling used to drive and verify performance work on the default preset.

## Tools

- **`util/bench.mjs`** — runs the full default preset over a directory of CSS files for a fixed warmup + measured iteration count, and reports per-file and total **median** and **min** time. Writes a JSON snapshot per `--label` so two branches can be compared directly.
- **`util/bench-profile.mjs`** — runs the same workload under the V8 CPU profiler and emits a `.cpuprofile`.
- **`util/analyze-cpuprofile.mjs`** — aggregates a `.cpuprofile` by **self-time** per function/package, so the hot paths to optimize are obvious.

## Methodology

- **Corpus**: released CSS from popular frameworks (Bootstrap, Bulma, Foundation, …). `bench.mjs --dir=<path>` points at any folder of `.css` files.
- **Per run**: warmup iterations to let V8 settle, then measured iterations; report the **median** (robust to outliers) and the **min** (least noisy).
- **A/B**: compare the per-`--label` JSON snapshots; interleave baseline and branch runs across rounds to spread thermal/scheduler drift.
- **Attribution**: pair the end-to-end number with an isolated single-plugin run plus a `.cpuprofile` self-time delta on the touched function, so a change's effect is measured, not assumed.

## Reproduce

```sh
node util/bench.mjs --dir=<css-dir> --label=before
# (apply change / switch branch)
node util/bench.mjs --dir=<css-dir> --label=after
```

`bench-results/` and `lcov.info` are gitignored.


util/bench.mjs
```diff
@@ -0,0 +1,190 @@
...
+function fmtMs(ms) {
+  return ms.toFixed(2).padStart(8) + ' ms';
+}
+
+async function benchOne(name, source, processor, iters, warmup) {
+  // Warmup — let V8 optimize hot paths before we measure.
+  for (let i = 0; i < warmup; i++) {
+    const res = await processor.process(source, { from: undefined });
+    // Force stringification so plugins that defer work via .toResult() complete.
+    void res.css;
```
ludofischer, line 65: How does this force stringification? Is the `void` operator needed for that?

util/analyze-cpuprofile.mjs
```diff
@@ -0,0 +1,139 @@
+// Summarize a V8 .cpuprofile by hottest self-time and hottest inclusive (total)
+// time, plus a roll-up by package (postcss-* / postcss / node internals).
+//
+// Usage:
+//   node util/analyze-cpuprofile.mjs <path/to/profile.cpuprofile> [topN]
+
+import { readFileSync } from 'fs';
+
+const file = process.argv[2];
```
ludofischer: Would it be possible to rename these general variable names? For example make it clear that `file` contains the path to the CPU profile.

util/analyze-cpuprofile.mjs
```diff
@@ -0,0 +1,139 @@
...
+if (!file) {
+  console.error('usage: node util/analyze-cpuprofile.mjs <profile> [topN]');
+  process.exit(1);
+}
+const topN = Number(process.argv[3]) || 25;
```
ludofischer, line 15: Same, hard to tell at a glance what `topN` means.

util/analyze-cpuprofile.mjs
```diff
@@ -0,0 +1,139 @@
+// Summarize a V8 .cpuprofile by hottest self-time and hottest inclusive (total)
+// time, plus a roll-up by package (postcss-* / postcss / node internals).
+//
```
ludofischer, line 3: The commend should say what the `topN` argument means.

util/analyze-cpuprofile.mjs
```diff
@@ -0,0 +1,139 @@
...
+const byId = new Map(nodes.map((n) => [n.id, n]));
+
+// Build child->parent map (nodes only store children, not parents).
+const parentOf = new Map();
+for (const n of nodes) {
+  if (n.children) for (const c of n.children) parentOf.set(c, n.id);
+}
+
+function frameLabel(n) {
+  const cf = n.callFrame;
```
ludofischer: Rename `cf` to `callFrame`? `cf` reminds me of ColdFusion...

util/analyze-cpuprofile.mjs
```diff
@@ -0,0 +1,139 @@
...
+  }
+  // Prepend the package name so plugin OnceExit's are distinguishable, then
+  // show the in-package path (lib/foo.js or src/index.js).
+  const pkg = packageOf(n);
+  const inPkg = url.match(/\/(src|lib|dist|types)\/(.+)$/);
+  const tail = inPkg ? `${inPkg[1]}/${inPkg[2]}` : url.replace(/^.*\//, '');
+  return `${name} @ ${pkg}/${tail || '<?>'}`;
+}
+
+function packageOf(n) {
```
ludofischer: Rename the argument `n`.

util/analyze-cpuprofile.mjs
```diff
@@ -0,0 +1,139 @@
...
+  // pnpm layout: node_modules/.pnpm/<name>@<ver>/node_modules/<name>/...
+  const mpnpm = url.match(/node_modules\/\.pnpm\/([^/]+)\//);
+  if (mpnpm) return mpnpm[1].replace(/@[\d.][^@]*$/, '');
+  const m2 = url.match(/node_modules\/(@[^/]+\/[^/]+|[^/]+)\//);
+  if (m2) return m2[1];
+  return url.replace(/^.*\//, '');
+}
+
+// Self time per node id.
+const selfUs = new Map();
```
ludofischer: Instead of a comment, `selfUs` could be called `selfTimePerNodeId`. Maybe the agent needs an instruction to follow the refactorings rules.

util/analyze-cpuprofile.mjs
```diff
@@ -0,0 +1,139 @@
...
+  const dt = timeDeltas[i] || 0;
+  selfUs.set(id, (selfUs.get(id) || 0) + dt);
+}
+
+const totalUs = samples.length
+  ? prof.endTime - prof.startTime
+  : timeDeltas.reduce((a, b) => a + b, 0);
+
+// Inclusive time: walk from each sample leaf up to the root, adding the delta.
+const inclUs = new Map();
```
ludofischer: Same story, why is it called `inclUs` if means inclusive _time_?

util/analyze-cpuprofile.mjs
```diff
@@ -0,0 +1,139 @@
...
+    const fn = n.callFrame.functionName;
+    if (fn === '(garbage collector)') return '<gc>';
+    if (fn === '(program)' || fn === '(root)' || fn === '(idle)') return fn;
+    return '<internal>';
+  }
+  if (url.startsWith('node:')) return url;
+  const m = url.match(/packages\/([^/]+)\//);
+  if (m) return m[1];
+  // pnpm layout: node_modules/.pnpm/<name>@<ver>/node_modules/<name>/...
+  const mpnpm = url.match(/node_modules\/\.pnpm\/([^/]+)\//);
```
ludofischer: Does this break when we upgrade to pnpm 11?

ludofischer (changes requested):
Looks OK, I would prefer more descriptive variable names in practice. Did you check the output of `node --prof-process`? https://nodejs.org/docs/latest/api/cli.html#prof-process Could that replace some of the scripts here?

util/bench.mjs
dkryaklin: You're right I've double checked void doesn't force anything

util/analyze-cpuprofile.mjs
dkryaklin: Renamed file to profilePath

util/analyze-cpuprofile.mjs
dkryaklin: Added a comment

util/analyze-cpuprofile.mjs
dkryaklin: Done

util/analyze-cpuprofile.mjs
dkryaklin: Renamed to node in both frameLabel and packageOf

util/analyze-cpuprofile.mjs
dkryaklin: Renamed to selfTimePerNodeId

util/analyze-cpuprofile.mjs
dkryaklin: Renamed to inclusiveTimePerNodeId

util/analyze-cpuprofile.mjs
dkryaklin: Good catch. It no longer parses the .pnpm/<name>@<ver> key

dkryaklin:
> Looks OK, I would prefer more descriptive variable names in practice. Did you check the output of `node --prof-process`? https://nodejs.org/docs/latest/api/cli.html#prof-process Could that replace some of the scripts here?

On --prof-process: it covers the per-function view and even shows V8 builtins this script doesn't, but it doesn't give selftime rolled up per npm package

I did drop bench-profile.mjs. node --cpu-prof util/bench.mjs --only=<file> generates the profile without a separate runner

util/bench.mjs
ludofischer: Mmh, according to the documentation, `void` does evaluate the expression, so maybe the comment is incorrect https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/void

util/bench.mjs
```diff
@@ -0,0 +1,190 @@
...
+const __dirname = dirname(fileURLToPath(import.meta.url));
+// Default corpus directory. Override with `--dir=<path>` to point at any folder
+// of .css files (the corpus itself is not committed).
+const DEFAULT_DIR = join(__dirname, '..', 'frameworks');
+const RESULTS_DIR = join(__dirname, '..', 'bench-results');
+
+function parseArgs(argv) {
+  const out = { label: 'baseline', iters: 15, warmup: 3, only: null, dir: null };
+  for (const a of argv.slice(2)) {
+    const m = a.match(/^--([^=]+)=(.*)$/);
```
ludofischer: Could some of this be replaced by `util.parseArgs`? https://nodejs.org/api/util.html#utilparseargsconfig Seems odd to use a regex.

ludofischer (commented):
I've added some additional remarks. Let me know if you want to tackle those. I would definitely suggest looking at the [built-in Node command-line argument parser](https://nodejs.org/api/util.html#utilparseargsconfig), it's available since Node.js 18 and seems to do most of what the `parseArgs` function here does.

util/bench.mjs
dkryaklin: Switched parseArgs to util.parseArgs

ludofischer (approved):

PR #1817, opened 2026-06-09
Include postcss-discard-overridden types in package

itosa-kazu:
## Summary
- include the postcss-discard-overridden types directory in the published package files
- keep the existing types entrypoint usable for TypeScript consumers

## Verification
- npm pack --dry-run --json --ignore-scripts
- installed the locally packed tarball into a clean TypeScript project and ran npx tsc --noEmit


ludofischer (approved):
Thanks!

PR #1819, opened 2026-06-11
fix: preserve apostrophes in svg data uris

### LINKED ISSUE #1783: [Bug]: cssnano produces bad SVG when it contains `&apos;` within an attribute
### Describe the bug

I noticed that cssnano produces an invalid SVG when minifying with SVGO. I also confirmed this is _not_ an SVGO bug, but a bug in how cssnano processes SVGO's result.

Specifically:

```xml
attr="&apos;value&apos;"
```

Becomes:
```xml
attr=''value''
```

These repeated single quotes

### Expected behaviour

Either use double quotes like SVGO does:
```xml
attr="'value'"
```

Or keep html entities
Either use double quotes like SVGO does:
```xml
attr='&apos;value&apos;'
```

### Steps to reproduce

Specifically take this SVG:
```svg
<svg xmlns="http://www.w3.org/2000/svg">
  <text font-size="20" font-family="&apos;Arial&apos;" transform="translate(0,20)">?</text>
</svg>
```

Convert it to base64, and use it in CSS (before cssnano). You will notice this renders a question mark, as expected:
```html
<style>
.foo {
    height: 30px;
    width: 20px;
    background-color: #9fd;

    background-image: url("data:image/svg+xml,%3Csvg%20xmlns=%27http://www.w3.org/2000/svg%27%3E%0A%20%20%20%20%3Ctext%20font-size=%2720%27%20font-family=%27%26apos;Arial%26apos;%27%20transform=%27translate%280%2C20%29%27%3E?%3C/text%3E%0A%3C/svg%3E%0A");
}

</style>
<div class="foo"></div>
```

However, when you pass it to cssnano with default settings, you get the following:
```html
<style>
.foo {
    height: 30px;
    width: 20px;
    background-color: #9fd;

    background-image:url("data:image/svg+xml;charset=utf-8,%3Csvg xmlns='http://www.w3.org/2000/svg'%3E%3Ctext font-family=''Arial'' font-size='20' transform='translate(0 20)'%3E?%3C/text%3E%3C/svg%3E")
}

</style>
<div class="foo"></div>
```

Note `font-family=''Arial''` (double single-quotes, `''`)

This generates an SVG with invalid syntax.

### Version

7.1.2

### Preset

default

### Environment

```shell
cssnano playground
```

### Package details

```shell
cssnano@7.1.2

and whatever the playground uses today.
```

### Additional context

SVGO seems to use double quotes. I noticed the result goes through this path, so I believe the bug is around here:

https://github.com/cssnano/cssnano/blob/aed9ec0dc8ab914e857fcb794979bf746abdd30d/packages/postcss-svgo/src/index.js#L88-L96

When I force `"'"` for quotes, I actually get the results as I expect. But I'm not sure if it's okay to force that value.

pupuking723:
Fixes #1783.

This keeps apostrophes inside optimized SVG data URI content as XML entities before the existing URL encoding step converts double-quoted SVG attributes to single quotes. Without this, SVGO output such as `font-family="'Arial'"` becomes invalid XML after encoding as `font-family=''Arial''`.

Changes:

- encode apostrophes as `&apos;` before percent-encoding SVG data URI output
- add a regression test for an encoded SVG attribute containing `&apos;Arial&apos;`
- add a patch changeset for `postcss-svgo`

Verification:

- `pnpm test:only packages/postcss-svgo/test/index.js --test-name-pattern "preserve apostrophes"`
- `pnpm test:only packages/postcss-svgo/test/index.js`
- `pnpm lint`
- `pnpm types`
- `pnpm test:only`
- `pnpm exec prettier --check .changeset/quiet-buckets-encode.md packages/postcss-svgo/src/lib/url.js packages/postcss-svgo/test/index.js`
- `git diff --check`

### DIFF
--- a/.changeset/quiet-buckets-encode.md
+++ b/.changeset/quiet-buckets-encode.md
@@ -1,0 +1,5 @@
+---
+"postcss-svgo": patch
+---
+
+Encode apostrophes as XML entities in SVG data URIs so optimized attributes remain valid when SVGO emits values containing apostrophes.
--- a/packages/postcss-svgo/src/lib/url.js
+++ b/packages/postcss-svgo/src/lib/url.js
@@ -5,6 +5,7 @@
  */
 function encode(data) {
   return data
+    .replace(/'/g, '&apos;')
     .replace(/"/g, "'")
     .replace(/%/g, '%25')
     .replace(/</g, '%3C')
--- a/packages/postcss-svgo/test/index.js
+++ b/packages/postcss-svgo/test/index.js
@@ -105,6 +105,14 @@ test(
 );
 
 test(
+  'should preserve apostrophes in uri-encoded svg attributes',
+  processCSS(
+    'h1{background:url("data:image/svg+xml,%3Csvg%20xmlns=%27http://www.w3.org/2000/svg%27%3E%3Ctext%20font-size=%2720%27%20font-family=%27%26apos;Arial%26apos;%27%20transform=%27translate%280%2C20%29%27%3E?%3C/text%3E%3C/svg%3E")}',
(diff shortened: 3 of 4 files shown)


ludofischer (approved):
Thanks!

PR #1824, opened 2026-06-15
Add regression test for border custom property merging

wyf027:
Refs cssnano/cssnano#1000.

## Summary
- Add an integration regression test for border shorthands with custom property widths followed by `border-width`.
- Assert cssnano's current full optimization pipeline preserves declaration order so the optimized output keeps the effective `1px` border width.
- This PR does not change optimizer implementation code; it codifies the current fixed behavior so the #1000 regression is caught if it comes back.

## Verification
- `pnpm install --frozen-lockfile`
- `pnpm exec node --test packages/cssnano/test/issue1000.js`
- `pnpm exec node --test packages/postcss-merge-longhand/test/borders.js`
- `git diff --check`

Note: the local Node version is v25.9.0, so pnpm reported the repository engine warning (expected ^22.11 || ^24.11 || >=26.0), but the focused checks above passed.

### DIFF
--- a/packages/cssnano/test/issue1000.js
+++ b/packages/cssnano/test/issue1000.js
@@ -1,0 +1,23 @@
+'use strict';
+const { test } = require('node:test');
+const processCss = require('./_processCss');
+
+test(
+  'preserves border width when merging border shorthands with custom properties (#1000)',
+  processCss(
+    `:root {
+      --border--width--thick: 10px;
+    }
+
+    .a {
+      border-top: var(--border--width--thick) solid blue;
+      border-width: 1px;
+    }
+
+    .a > .b {
+      border-left: var(--border--width--thick) solid blue;
+      border-width: 1px;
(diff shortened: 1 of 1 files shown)


ludofischer:
This looks like it is just a test, how does it solve the issue?

wyf027:
Thanks for calling that out. You're right: this PR does not change the optimizer implementation.

I rechecked the current code path with the full `cssnano` integration pipeline, and the #1000 fixture now produces the expected output where `border-width:1px` remains after the side shorthand. So this PR is intended as regression coverage for the already-fixed current behavior, not as a runtime fix.

I updated the PR body from `Closes` to `Refs` and made that explicit. Validation from this branch:

- `pnpm exec node --test packages/cssnano/test/issue1000.js`
- `pnpm exec node --test packages/postcss-merge-longhand/test/borders.js`
- `git diff --check`

PR #1826, opened 2026-06-29
fix(postcss-reduce-idents): only rename a referenced counter style

mahirhir:
`postcss-reduce-idents` keeps a per-instance cache of idents so names stay unique when one plugin instance processes several files. The keyframes reducer renames an `@keyframes` at-rule only when its name was actually referenced in the current file. The counter-style reducer is missing that check: it renames an `@counter-style` whenever the cached reference count is above zero, and that count is never reset between files.

So when the same plugin instance runs over more than one file, a counter style that was referenced in an earlier file still has a count above zero, and an unrelated `@counter-style` with the same name in a later file gets renamed even though nothing references it there.

### Reproduction

```js
const postcss = require('postcss');
const reduceIdents = require('postcss-reduce-idents');

const instance = postcss([reduceIdents()]);

instance.process(
  '@counter-style custom{system:extends decimal;suffix:"> "}ol{list-style:custom}',
  { from: undefined }
).css;
// @counter-style a{...}ol{list-style:a}   (correct, custom is referenced)

instance.process(
  '@counter-style custom{system:extends decimal;suffix:"> "}',
  { from: undefined }
).css;
// before: @counter-style a{...}            (wrong: custom is not referenced here)
// after:  @counter-style custom{...}       (left untouched)
```

The fix mirrors the guard the keyframes reducer already uses, added in `d5b929106` ("Prevent name collision when parsing multiple files"): collect the names referenced in the current file and only rename the at-rule when its name is in that set.

I added a regression test next to the existing keyframes reuse test. The package test suite passes (50 tests).

### DIFF
--- a/.changeset/counter-style-reused-instance.md
+++ b/.changeset/counter-style-reused-instance.md
@@ -1,0 +1,5 @@
+---
+"postcss-reduce-idents": patch
+---
+
+Do not rename an unreferenced `@counter-style` when the plugin instance is reused across files. The counter-style reducer now tracks references per file like the keyframes reducer, so a stale reference count from an earlier file no longer renames an unreferenced counter style in a later one.
--- a/packages/postcss-reduce-idents/src/lib/counter-style.js
+++ b/packages/postcss-reduce-idents/src/lib/counter-style.js
@@ -97,11 +97,17 @@ module.exports = function () {
     },
 
     transform() {
+      const referenced = new Set();
+
       // Iterate each property and change their names
       decls.forEach((decl) => {
         decl.value = valueParser(decl.value)
           .walk((node) => {
             if (node.type === 'word' && node.value in cache) {
+              if (!referenced.has(node.value)) {
+                referenced.add(node.value);
+              }
+
               cache[node.value].count++;
 
               node.value = cache[node.value].ident;
@@ -114,7 +120,7 @@ module.exports = function () {
       atRules.forEach((rule) => {
         const cached = cache[rule.params];
 
-        if (cached && cached.count > 0) {
+        if (cached && cached.count > 0 && referenced.has(rule.params)) {
           rule.params = cached.ident;
         }
       });
--- a/packages/postcss-reduce-idents/test/index.js
+++ b/packages/postcss-reduce-idents/test/index.js
@@ -569,6 +569,30 @@ test('should not generate same ident when plugin instance is reused', () => {
   });
 });
 
+test('should not rename an unreferenced counter style when plugin instance is reused', () => {
+  const instance = postcss(plugin);
+
+  return Promise.all([
+    instance.process(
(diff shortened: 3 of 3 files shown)


.changeset/counter-style-reused-instance.md
```diff
@@ -0,0 +1,5 @@
+---
+"postcss-reduce-idents": patch
+---
+
+Do not rename an unreferenced `@counter-style` when the plugin instance is reused across files. The counter-style reducer now tracks references per file like the keyframes reducer, so a stale reference count from an earlier file no longer renames an unreferenced counter style in a later one.
```
ludofischer, line 5: What does a "stale reference count" mean?

ludofischer (commented):
I'm afraid I don not understand the bug being fixed. If the counter style 'unreferenced', why does it matter if it gets renamed?

mahirhir:
The rename itself isn't miscomputed, it's just not scoped to the file being processed. `cached.count` is a per-instance counter that's never reset between `process()` calls, so a reference in an earlier file leaves `count > 0` when a later file runs through the same instance. If that later file has a `@counter-style custom{...}` that nothing in it actually references, the old check (`cached.count > 0`) still passed because of the earlier file's reference, so the declaration got renamed anyway even though nothing in that file's own output points at it.

That matters because the reducer's whole guarantee is: if you rename an ident, every reference to it gets renamed too. Renaming a declaration in isolation from the file that actually uses that ident breaks that guarantee, since the plugin has no way to know the earlier file (or anything outside its tracked declarations) still expects the original name. The keyframes reducer already avoids this by tracking which names were referenced per file (added in d5b929106 for the same cross-file leak); this PR gives the counter-style reducer the same per-file `referenced` set.

ludofischer:
> The rename itself isn't miscomputed, it's just not scoped to the file being processed. `cached.count` is a per-instance counter that's never reset between `process()` calls, so a reference in an earlier file leaves `count > 0` when a later file runs through the same instance. If that later file has a `@counter-style custom{...}` that nothing in it actually references, the old check (`cached.count > 0`) still passed because of the earlier file's reference, so the declaration got renamed anyway even though nothing in that file's own output points at it.
> 
> That matters because the reducer's whole guarantee is: if you rename an ident, every reference to it gets renamed too. Renaming a declaration in isolation from the file that actually uses that ident breaks that guarantee, since the plugin has no way to know the earlier file (or anything outside its tracked declarations) still expects the original name. The keyframes reducer already avoids this by tracking which names were referenced per file (added in [d5b9291](https://github.com/cssnano/cssnano/commit/d5b92910617ef3c0b79e6d2e54cefa7dd63d4d9b) for the same cross-file leak); this PR gives the counter-style reducer the same per-file `referenced` set.

Could you give a short example of how it breaks? The input CSS and an explanation of why the output is wrong.

ludofischer:
It's not clear to me whether this is fixing a bug. Re-open if you can show why the current behaviour causes an issue.

PR #1845, opened 2026-07-26
feat(postcss-normalize-url): add removeTrailingSlash option

### LINKED ISSUE #1705: [Bug]: `normalizeUrl` removes trailing slashes
### Describe the bug

The [normalizeUrl](https://cssnano.github.io/cssnano/docs/optimisations/normalizeurl/) optimization removes trailing slashes.  On servers that require trailing slashes, this optimization will break the url.

### Expected behaviour

Trailing slashes are preserved

### Steps to reproduce

```js
import postcss from "postcss";
import autoprefixer from "autoprefixer";
import cssnano from "cssnano";
import cssnanoPresetDefault from "cssnano-preset-default";

const processor = postcss([
  autoprefixer(),
  cssnano({
    preset: cssnanoPresetDefault(),
  }),
]);


const css = `
@font-face {
    font-family: "Inter";
    src: url("http://localhost:4321/api/woff2/inter.woff2/") format("woff2");
  }
`

const result = await processor.process(css, {
    // avoid warning https://stackoverflow.com/a/63193341
    from: undefined,
});
console.log(result);
```

### Version

7.0.7

### Preset

default

### Environment

```shell
System:
    OS: macOS 15.5
    CPU: (16) x64 Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
    Memory: 1.70 GB / 64.00 GB
    Shell: 5.9 - /bin/zsh
```

### Package details

```shell
├─┬ @astrojs/react@4.3.0
│ └─┬ vite@6.3.5
│   └── postcss@8.5.6 deduped
├─┬ autoprefixer@10.4.21
│ └── postcss@8.5.6 deduped
├─┬ cssnano-preset-default@7.0.7
│ ├─┬ css-declaration-sorter@7.2.0
│ │ └── postcss@8.5.6 deduped
│ ├─┬ cssnano-utils@5.0.1
│ │ └── postcss@8.5.6 deduped
│ ├─┬ postcss-calc@10.1.1
│ │ └── postcss@8.5.6 deduped
│ ├─┬ postcss-colormin@7.0.3
│ │ └── postcss@8.5.6 deduped
│ ├─┬ postcss-convert-values@7.0.5
│ │ └── postcss@8.5.6 deduped
│ ├─┬ postcss-discard-comments@7.0.4
│ │ └── postcss@8.5.6 deduped
│ ├─┬ postcss-discard-duplicates@7.0.2
│ │ └── postcss@8.5.6 deduped
│ ├─┬ postcss-discard-empty@7.0.1
│ │ └── postcss@8.5.6 deduped
│ ├─┬ postcss-discard-overridden@7.0.1
│ │ └── postcss@8.5.6 deduped
│ ├─┬ postcss-merge-longhand@7.0.5
│ │ ├── postcss@8.5.6 deduped
│ │ └─┬ stylehacks@7.0.5
│ │   └── postcss@8.5.6 deduped
│ ├─┬ postcss-merge-rules@7.0.5
│ │ └── postcss@8.5.6 deduped
│ ├─┬ postcss-minify-font-values@7.0.1
│ │ └── postcss@8.5.6 deduped
│ ├─┬ postcss-minify-gradients@7.0.1
│ │ └── postcss@8.5.6 deduped
│ ├─┬ postcss-minify-params@7.0.3
│ │ └── postcss@8.5.6 deduped
│ ├─┬ postcss-minify-selectors@7.0.5
│ │ └── postcss@8.5.6 deduped
│ ├─┬ postcss-normalize-charset@7.0.1
│ │ └── postcss@8.5.6 deduped
│ ├─┬ postcss-normalize-display-values@7.0.1
│ │ └── postcss@8.5.6 deduped
│ ├─┬ postcss-normalize-positions@7.0.1
│ │ └── postcss@8.5.6 deduped
│ ├─┬ postcss-normalize-repeat-style@7.0.1
│ │ └── postcss@8.5.6 deduped
│ ├─┬ postcss-normalize-string@7.0.1
│ │ └── postcss@8.5.6 deduped
│ ├─┬ postcss-normalize-timing-functions@7.0.1
│ │ └── postcss@8.5.6 deduped
│ ├─┬ postcss-normalize-unicode@7.0.3
│ │ └── postcss@8.5.6 deduped
│ ├─┬ postcss-normalize-url@7.0.1
│ │ └── postcss@8.5.6 deduped
│ ├─┬ postcss-normalize-whitespace@7.0.1
│ │ └── postcss@8.5.6 deduped
│ ├─┬ postcss-ordered-values@7.0.2
│ │ └── postcss@8.5.6 deduped
│ ├─┬ postcss-reduce-initial@7.0.3
│ │ └── postcss@8.5.6 deduped
│ ├─┬ postcss-reduce-transforms@7.0.1
│ │ └── postcss@8.5.6 deduped
│ ├─┬ postcss-svgo@7.0.2
│ │ └── postcss@8.5.6 deduped
│ ├─┬ postcss-unique-selectors@7.0.4
│ │ └── postcss@8.5.6 deduped
│ └── postcss@8.5.6 deduped
├─┬ cssnano@7.0.7
│ └── postcss@8.5.6 deduped
└── postcss@8.5.6
```

### Additional context

Disabling `normalizeUrl` will leave the trailing slash, but… the url will not be normalized.

Hashim1999164:
### Summary

Adds a `removeTrailingSlash` option to `postcss-normalize-url` so absolute URLs can keep a trailing slash when it is significant for the server. Default remains `true`, so existing minification behavior is unchanged.

### Motivation

Some hosts require a trailing slash (for example font files served under a path that ends with `/`). The current normalizer always strips it, which breaks those URLs. Maintainer feedback on #1705 was that an option to disable this is fine.

Usage:

```js
postcss([
  require('postcss-normalize-url')({ removeTrailingSlash: false }),
]);
```

Or via cssnano:

```js
cssnano({
  preset: ['default', { normalizeUrl: { removeTrailingSlash: false } }],
});
```

Closes #1705

- [ ] Yes
- [x] No

### Checklist

### Documentation

- [x] My proposed change require an update to the documentation
- [x] I have added or updated the relevant documentation

### Test plan

- [x] `node --test packages/postcss-normalize-url/test/*.js` (63 passing)
- [x] Default still strips trailing slashes
- [x] `removeTrailingSlash: false` preserves them while still normalizing other parts of the URL


ludofischer (commented):
I wonder if there should be an option to remove the trailing slash. According to https://github.com/cssnano/cssnano/issues/1705 removal might be unsafe in general and I can't think it can save many bytes. @alexander-akait what do you think?

packages/postcss-normalize-url/src/normalize.js
```diff
@@ -138,13 +144,19 @@ function normalizeUrl(urlString) {
     urlObject.hostname = urlObject.hostname.replace(trailingDotRegex, '');
   }
 
-  urlObject.pathname = urlObject.pathname.replace(trailingSlashRegex, '');
```
ludofischer, line 141: Only this line should be removed.

packages/postcss-normalize-url/src/normalize.js
```diff
@@ -138,13 +144,19 @@ function normalizeUrl(urlString) {
...
+  if (removeTrailingSlash) {
+    urlObject.pathname = urlObject.pathname.replace(trailingSlashRegex, '');
+  }
 
   // Take advantage of many of the Node `url` normalizations
   urlString = urlObject.toString();
 
-  // Remove ending `/`
-  if (urlObject.pathname === '/' && urlObject.hash === '') {
```
ludofischer, line 147: This seems to be correct, since it seems to be checking whether the URI is hostname-only.

ludofischer (changes requested):
From my interpretation of the specs (URI RFC https://www.rfc-editor.org/info/rfc3986/#section-6.2.2.3 referenced by HTTP RFChttps://www.rfc-editor.org/info/rfc9110/#section-4.2.3), removing the trailing slash does yield an equivalent URI in general. For HTTP, removing the trailing slash only yields an equivalent URI if the URI only contains the domain name (http://example.com and http://example.com/ are equivalent, but not http://example.com/path/ and http:// example.com/path). This is confirmed by the fact that many servers actively redirect from one variant to the other. I think that cssnano should keep the trailing slash, and it should be considered a bug fix, so no need to introduce an extra option to remove the trailing slash.

Hashim1999164:
Updated this to the focused bug fix requested: removed only the path trailing-slash normalization, kept bare-origin normalization intact, and updated the focused test. The broader option/API changes have been reverted.

ludofischer (approved):
Thanks!

PR #1958, opened 2026-08-31
fix: add main and types to package manifests

### LINKED ISSUE #1935: Package exports change gets flagged by eslint-plugin-import
This might be an issue with ESLint, but fyi that after https://github.com/cssnano/cssnano/pull/1863, https://github.com/import-js/eslint-plugin-import flags:

`error  Unable to resolve path to module 'cssnano'  import/no-unresolved`

When using `import cssnano from 'cssnano'` in the codebase, and it must be updated to `import cssnano from 'cssnano/src/index.js'` in order for the linter rule to pass. 

I would think the wildcard default at https://github.com/cssnano/cssnano/blob/4c4cd75cdda376e3f36f1f00143b61880b235042/packages/cssnano/package.json#L8 would satisfy it, so I'm not sure what the issue is, but I wanted to flag it here in case there's an issue in the package exports that I'm not seeing.

dkryaklin:
Fixes #1935

`main` and `types` were dropped in #1863 and first shipped in 8.0.3, so resolvers that predate `exports` (eslint-plugin-import, TypeScript `moduleResolution: node`) fail to resolve any package. 9.0 also dropped the ./package.json` subpath

### DIFF
--- a/util/checkPackaging.mjs
+++ b/util/checkPackaging.mjs
@@ -1,80 +1,49 @@
 import { existsSync, readFileSync, readdirSync } from 'node:fs';
+import { createRequire } from 'node:module';
 import { fileURLToPath } from 'node:url';
 import path from 'node:path';
 
 /**
- * Nothing in the workspace resolves a package through its manifest — tests
- * import the source by relative path — so a manifest can point nowhere and
- * every test still passes. That is how 8.0.3 shipped without `main`.
+ * Checks that every package can still be resolved by tools that predate
+ * `exports`, which read `main` and `types` instead.
  *
- * `exports` is read by Node and by modern bundlers. `main` and `types` are
- * read by resolvers that predate it, such as eslint-plugin-import and
- * TypeScript's classic `moduleResolution: node`. Both have to be present, and
- * every path they name has to exist and be published.
+ * Nothing else here covers that. The test suite imports the source by
+ * relative path, and Node resolves the workspace through `exports`, so a
+ * package can become unresolvable for those tools while everything passes.
+ * That is how 8.0.3 shipped without `main`.
+ *
+ * The resolving is done by `resolve`, the same package eslint-plugin-import
+ * uses, rather than by a copy of its algorithm kept here.
  */
 
-/**
- * Every string in a manifest value, however deeply nested.
- * @param {unknown} value
- * @param {string[]} found
- * @return {string[]}
- */
-function targets(value, found = []) {
-  if (typeof value === 'string') {
-    found.push(value);
-  } else if (value !== null && typeof value === 'object') {
-    for (const nested of Object.values(value)) {
-      targets(nested, found);
-    }
-  }
+const require = createRequire(import.meta.url);
+const resolve = require('resolve');
 
-  return found;
-}
-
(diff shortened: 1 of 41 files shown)


ludofischer (commented):
Did you check this fixes your problem? According to the TypeScript documentation, `moduleResolution: node` [only supports `require()`](https://www.typescriptlang.org/tsconfig/#moduleResolution), and for `eslint-plugin-import`, it sounds like a bug with the plugin (they've got an open issue since 2020 https://github.com/import-js/eslint-plugin-import/issues/1810). csstools [also only uses `exports` ](https://github.com/csstools/postcss-plugins/blob/main/packages/css-tokenizer/package.json)

util/checkPackaging.mjs
```diff
@@ -0,0 +1,94 @@
+import { existsSync, readFileSync, readdirSync } from 'node:fs';
```
ludofischer, line 1: I think it would be better if the test checked whether the modules can be resolved. From my cursory research, the eslint-plugin-import  is using https://github.com/browserify/resolve/, which has open issue https://github.com/browserify/resolve/issues/222, where it says it does not support ESM either, so in that case, even with `main`, it would not work.

dkryaklin:
> Did you check this fixes your problem? According to the TypeScript documentation, `moduleResolution: node` [only supports `require()`](https://www.typescriptlang.org/tsconfig/#moduleResolution), and for `eslint-plugin-import`, it sounds like a bug with the plugin (they've got an open issue since 2020 [import-js/eslint-plugin-import#1810](https://github.com/import-js/eslint-plugin-import/issues/1810)). csstools [also only uses `exports` ](https://github.com/csstools/postcss-plugins/blob/main/packages/css-tokenizer/package.json)

Made a small demo: https://github.com/dkryaklin/cssnano-resolution-demo

Two copies of cssnano 9.0.1, one adds main and types, nothing else differs. resolve and eslint-plugin-import fail on the first and work on the second, even though both copies are ESM.

util/checkPackaging.mjs
dkryaklin: Adressed in https://github.com/cssnano/cssnano/pull/1958/changes/dfa4befdf0439cec629e7a777b4df99d240af75a

ludofischer (approved):

ludofischer:
> Made a small demo: https://github.com/dkryaklin/cssnano-resolution-demo
> 
> Two copies of cssnano 9.0.1, one adds main and types, nothing else differs. resolve and eslint-plugin-import fail on the first and work on the second, even though both copies are ESM.

Is there a reason you need to use eslint-plugin-import? There  seem to be a few alternatives that try to implement more recent specifications.

dkryaklin:
> > Made a small demo: https://github.com/dkryaklin/cssnano-resolution-demo
> > Two copies of cssnano 9.0.1, one adds main and types, nothing else differs. resolve and eslint-plugin-import fail on the first and work on the second, even though both copies are ESM.
> 
> Is there a reason you need to use eslint-plugin-import? There seem to be a few alternatives that try to implement more recent specifications.

No reason. Just because #1935 mentioned it

PR #1969, opened 2026-09-07
Fix SvgoParserError on data URIs mixing percent-encoding with a bare %

### LINKED ISSUE #1961: [Bug]: Data URIs that include any percent-encoded chars and an unencoded % provoke SvgoParserError
### Describe the bug

Running `cssnano` on some CSS having an `url()` function that contains a data URI, can, in some instances, provoke the following error message:

```
⚠  SvgoParserError: <input>:1:1: Non-whitespace before first tag.

> 1 | %3csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 100 100' style='backgroun…
    | ^
 [postcss-svgo]
```

in cases where there is no white space before the first tag. `cssnano` appears to process the data URI regardless.

This appears to happen if the data URI percent encodes at least one character and also includes an unencoded `%` symbol.

### Expected behaviour

Either, `cssnano` should process the css without emitting an error, or the error emitted should relate to the root cause and explain what it is so it can be corrected.

### Steps to reproduce

Run `cssnano` on

```css
:root
{
  /*The SVG does not contain any percent encoded characters and includes a raw '%' character.*/
  --var-1: url("data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 100 100' style='background: rgb(0 0 0 / 80%);' ></svg>");

  /*The SVG contains one percent encoded character ('<' -> '%3c') and includes a raw '%' character */
  --var-2: url("data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 100 100' style='background: rgb(0 0 0 / 80%);' ></svg>");

  /*The SVG contains two percent encoded characters ('<' -> '%3c' and '%' => '%25') */
  --var-3: url("data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 100 100' style='background: rgb(0 0 0 / 80%25);' ></svg>");
}
```

Get an error message relating to the processing of the `--var-2` data URI only.

```
7:3     ⚠  SvgoParserError: <input>:1:1: Non-whitespace before first tag.

> 1 | %3csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 100 100' style='backgroun…
    | ^
 [postcss-svgo]
```

### Version

9.0.2

### Preset

(no preset)

### Environment

```shell
System:
  OS: Ubuntu 24.04.4 LTS
  CPU: (2) x64 Intel(R) Xeon(R) W-2295 CPU @ 3.00GHz
  Memory: 1.17 GB / 3.82 GB
  Shell: 5.2.21 - /bin/bash
```

### Package details

```shell
├─┬ autoprefixer@10.5.4
│ └── postcss@8.5.26 deduped
├─┬ cssnano@9.0.2
│ ├─┬ cssnano-preset-default@9.0.2
│ │ ├─┬ cssnano-utils@7.0.1
│ │ │ └── postcss@8.5.26 deduped
│ │ ├─┬ postcss-calc@11.0.3
│ │ │ └── postcss@8.5.26 deduped
│ │ ├─┬ postcss-colormin@9.0.1
│ │ │ └── postcss@8.5.26 deduped
│ │ ├─┬ postcss-convert-values@9.0.1
│ │ │ └── postcss@8.5.26 deduped
│ │ ├─┬ postcss-discard-comments@9.0.1
│ │ │ └── postcss@8.5.26 deduped
│ │ ├─┬ postcss-discard-duplicates@9.0.1
│ │ │ └── postcss@8.5.26 deduped
│ │ ├─┬ postcss-discard-empty@9.0.1
│ │ │ └── postcss@8.5.26 deduped
│ │ ├─┬ postcss-discard-overridden@9.0.1
│ │ │ └── postcss@8.5.26 deduped
│ │ ├─┬ postcss-merge-longhand@9.0.2
│ │ │ ├── postcss@8.5.26 deduped
│ │ │ └─┬ stylehacks@9.0.2
│ │ │   └── postcss@8.5.26 deduped
│ │ ├─┬ postcss-merge-rules@9.0.2
│ │ │ └── postcss@8.5.26 deduped
│ │ ├─┬ postcss-minify-font-values@9.0.1
│ │ │ └── postcss@8.5.26 deduped
│ │ ├─┬ postcss-minify-gradients@9.0.1
│ │ │ └── postcss@8.5.26 deduped
│ │ ├─┬ postcss-minify-params@9.0.1
│ │ │ └── postcss@8.5.26 deduped
│ │ ├─┬ postcss-minify-selectors@9.0.2
│ │ │ └── postcss@8.5.26 deduped
│ │ ├─┬ postcss-normalize-charset@9.0.1
│ │ │ └── postcss@8.5.26 deduped
│ │ ├─┬ postcss-normalize-display-values@9.0.1
│ │ │ └── postcss@8.5.26 deduped
│ │ ├─┬ postcss-normalize-positions@9.0.1
│ │ │ └── postcss@8.5.26 deduped
│ │ ├─┬ postcss-normalize-repeat-style@9.0.1
│ │ │ └── postcss@8.5.26 deduped
│ │ ├─┬ postcss-normalize-string@9.0.1
│ │ │ └── postcss@8.5.26 deduped
│ │ ├─┬ postcss-normalize-timing-functions@9.0.1
│ │ │ └── postcss@8.5.26 deduped
│ │ ├─┬ postcss-normalize-unicode@9.0.1
│ │ │ └── postcss@8.5.26 deduped
│ │ ├─┬ postcss-normalize-url@9.0.1
│ │ │ └── postcss@8.5.26 deduped
│ │ ├─┬ postcss-normalize-whitespace@9.0.1
│ │ │ └── postcss@8.5.26 deduped
│ │ ├─┬ postcss-ordered-values@9.0.1
│ │ │ └── postcss@8.5.26 deduped
│ │ ├─┬ postcss-reduce-initial@9.0.1
│ │ │ └── postcss@8.5.26 deduped
│ │ ├─┬ postcss-reduce-transforms@9.0.1
│ │ │ └── postcss@8.5.26 deduped
│ │ ├─┬ postcss-svgo@9.0.1
│ │ │ └── postcss@8.5.26 deduped
│ │ ├─┬ postcss-unique-selectors@9.0.1
│ │ │ └── postcss@8.5.26 deduped
│ │ └── postcss@8.5.26 deduped
│ └── postcss@8.5.26 deduped
├─┬ postcss-cli@11.0.1
│ ├─┬ postcss-load-config@5.1.0
│ │ └── postcss@8.5.26 deduped
│ ├─┬ postcss-reporter@7.1.0
│ │ └── postcss@8.5.26 deduped
│ └── postcss@8.5.26 deduped
└── postcss@8.5.26
```

### Additional context

_No response_

SamyakJ05:
## Summary

Fixes #1961.

`postcss-svgo`'s `decode` calls `decodeURIComponent` on the whole data URI string. `decodeURIComponent` throws `URI malformed` whenever a `%` sits next to characters that don't form a valid escape — which happens for entirely ordinary CSS like `rgb(0 0 0 / 80%)`. The `try/catch` in `minifySVG` treats that as "this string isn't percent-encoded" and passes it straight to svgo unchanged.

That's fine when the string truly isn't encoded at all. But it's wrong when the string genuinely *is* percent-encoded (e.g. contains `%3c`) and *also* happens to contain a bare `%` elsewhere (e.g. `80%`) — the whole decode fails, so the still-encoded string (starting with `%3c...`) gets handed to svgo as if it were raw SVG, and svgo throws `SvgoParserError: Non-whitespace before first tag`.

### Reproduction (from the issue)

```css
:root {
  /* no percent-encoding, works fine */
  --var-1: url("data:image/svg+xml,<svg ... style='background: rgb(0 0 0 / 80%);' ></svg>");

  /* one %3c + bare 80% -> currently throws SvgoParserError */
  --var-2: url("data:image/svg+xml,%3csvg ... style='background: rgb(0 0 0 / 80%);' ></svg>");

  /* fully encoded (80% -> %25), works fine */
  --var-3: url("data:image/svg+xml,%3csvg ... style='background: rgb(0 0 0 / 80%25);' ></svg>");
}
```

Only `--var-2` reproduces the error.

## Fix

`decode` now decodes each **contiguous run** of `%XX` escapes as a unit, instead of decoding the whole string in one `decodeURIComponent` call:

```diff
 const encode = encodeURIComponent;
-const decode = decodeURIComponent;
+function decode(str) {
+  return str.replace(/(?:%[0-9a-fA-F]{2})+/g, (run) => {
+    try {
+      return decodeURIComponent(run);
+    } catch {
+      return run;
+    }
+  });
+}
```

A bare `%` that isn't part of a `%XX` escape is never matched by the regex, so it's left untouched rather than failing the whole decode. Runs (not individual `%XX` tokens) are decoded together so multi-byte UTF-8 escapes like `%C3%A9` still round-trip correctly — decoding a lone continuation byte on its own throws, which a naive per-token fix would silently regress.

## Test plan

- Added tests to `packages/postcss-svgo/test/index.js` reproducing all three cases from the issue (unencoded, partially-encoded-with-bare-%, fully-encoded), plus direct unit tests on `decode` for the bare-% and multi-byte-UTF-8 cases.
- Full `postcss-svgo` suite: 39 passed (34 existing + 5 new), 0 failed.
- `oxlint` and `oxfmt --check` clean on changed files.
- Added a changeset (`patch` bump for `postcss-svgo`) per CONTRIBUTING.md.

### DIFF
--- a/packages/postcss-svgo/src/lib/url.js
+++ b/packages/postcss-svgo/src/lib/url.js
@@ -1,5 +1,19 @@
 const encode = encodeURIComponent;
 
-const decode = decodeURIComponent;
+// A run of one or more contiguous %XX escapes is decoded as a unit so
+// multi-byte UTF-8 sequences (e.g. %C3%A9) round-trip correctly. A run
+// that isn't valid percent-encoding (e.g. it sits next to a bare `%`
+// that isn't part of an escape, like `80%` in `rgb(0 0 0 / 80%)`) is
+// left untouched instead of failing the whole string.
+// See: https://github.com/cssnano/cssnano/issues/1961
+function decode(str) {
+  return str.replace(/(?:%[0-9a-fA-F]{2})+/g, (run) => {
+    try {
+      return decodeURIComponent(run);
+    } catch {
+      return run;
+    }
+  });
+}
 
 export { encode, decode };
--- a/packages/postcss-svgo/test/index.js
+++ b/packages/postcss-svgo/test/index.js
@@ -299,4 +299,36 @@ test('should not crash on malformed urls when encoded', () => {
   assert.doesNotThrow(() => decode(svg));
 });
 
+test('decode should leave a bare percent sign next to valid escapes untouched', () => {
+  assert.strictEqual(decode('rgb(0 0 0 / 80%)'), 'rgb(0 0 0 / 80%)');
+  assert.strictEqual(decode('%3crgb(0 0 0 / 80%)'), '<rgb(0 0 0 / 80%)');
+});
+
+test('decode should still correctly decode multi-byte UTF-8 sequences', () => {
+  assert.strictEqual(decode('caf%C3%A9'), 'café');
+});
+
 test('should use the postcss plugin api', usePostCSSPlugin(plugin()));
+
+// https://github.com/cssnano/cssnano/issues/1961
+test(
+  'should not warn on a partially-encoded svg containing a bare percent sign',
+  processCSS(
+    "h1{background:url(\"data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 100 100' style='background: rgb(0 0 0 / 80%);' ></svg>\")}",
+    'h1{background:url("data:image/svg+xml;charset=utf-8,%3Csvg%20xmlns%3D%22http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%22%20style%3D%22background%3Argb(0%200%200%2F80%25)%22%20viewBox%3D%220%200%20100%20100%22%2F%3E")}'
+  )
+);
+
+test('should not warn on an unencoded svg containing a bare percent sign', async () => {
+  const css =
+    "h1{background:url(\"data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 100 100' style='background: rgb(0 0 0 / 80%);' ></svg>\")}";
+  const result = await postcss(plugin()).process(css, { from: undefined });
+  assert.strictEqual(result.messages.length, 0);
+});
+
+test('should not warn on a fully-encoded svg with an encoded percent sign', async () => {
+  const css =
+    "h1{background:url(\"data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 100 100' style='background: rgb(0 0 0 / 80%25);' ></svg>\")}";
+  const result = await postcss(plugin()).process(css, { from: undefined });
+  assert.strictEqual(result.messages.length, 0);
+});
--- a/.changeset/fix-svgo-data-uri-bare-percent.md
+++ b/.changeset/fix-svgo-data-uri-bare-percent.md
@@ -1,0 +1,5 @@
+---
+"postcss-svgo": patch
+---
+
+Fix `SvgoParserError` on data URIs that mix percent-encoded characters with a bare, unencoded `%` (e.g. `rgb(0 0 0 / 80%)`). `decode` now decodes each contiguous run of `%XX` escapes as a unit instead of failing the whole string when it also contains a `%` that isn't part of a valid escape.


packages/postcss-svgo/src/lib/url.js
```diff
@@ -1,5 +1,19 @@
 const encode = encodeURIComponent;
 
-const decode = decodeURIComponent;
+// A run of one or more contiguous %XX escapes is decoded as a unit so
+// multi-byte UTF-8 sequences (e.g. %C3%A9) round-trip correctly. A run
+// that isn't valid percent-encoding (e.g. it sits next to a bare `%`
+// that isn't part of an escape, like `80%` in `rgb(0 0 0 / 80%)`) is
+// left untouched instead of failing the whole string.
+// See: https://github.com/cssnano/cssnano/issues/1961
+function decode(str) {
```
ludofischer, line 9: What I am wondering is if the input in https://github.com/cssnano/cssnano/issues/1961 is correctly encoded. If cssnano receives incorrectly encoded data or even ambiguous one, I think it should just pass it through instead of trying to "fix" it as the generated CSS then might behave differently than the original one. Also I would prefer not operating on encoded data with RegEx as that also introduced bugs in the past.

packages/postcss-svgo/test/index.js
```diff
@@ -299,4 +299,36 @@ test('should not crash on malformed urls when encoded', () => {
   assert.doesNotThrow(() => decode(svg));
 });
 
+test('decode should leave a bare percent sign next to valid escapes untouched', () => {
+  assert.strictEqual(decode('rgb(0 0 0 / 80%)'), 'rgb(0 0 0 / 80%)');
+  assert.strictEqual(decode('%3crgb(0 0 0 / 80%)'), '<rgb(0 0 0 / 80%)');
```
ludofischer, line 304: What kind of encoding is that? If it is not a correct URL encoding then it should be left untouched in my opinion. We cannot start guessing every possible encoding that the user might have wanted. Also there might be slight security implications as this kind of garbled text can be used to introduce unexpected content. Should not be too bad as usually the user controls the input for a css minifier, but it is something to keep in mind.

mind-bending-forks: There shouldn't be any guessing to be done. Best to follow the standards. I don't know what cssnano does at present, but I imagine it ought to work something like as follows:

First, cssnano parses the input as CSS, following [rules for quoting or not of the URL](https://drafts.csswg.org/css-values-3/#url), so that it can extract the raw URL from the CSS. It then needs to check that it is a form of URL permitted by the CSS standard.

The CSS standard linked to above indicates that it should be treated as an URL that conforms to the [WhatWG spec](https://url.spec.whatwg.org/), so that standard should be looked to when parsing and decoding the URL. So the URL can be parsed and checked for being in a form that CSS permits. At this point, it may be determined to be an data URI containing SVG data.

The data part should then be [decoded according to the WhatWG spec](https://url.spec.whatwg.org/#percent-encoded-bytes).

Finally, once the SVG part of the data URL has been extracted and decoded, it is sent to the postcss-svgo library as plain SVG for it to validate.

So, I think the instructions at https://url.spec.whatwg.org/#percent-encoded-bytes ought to be followed to decode it correctly. These instructions will already have been reviewed for security implications.

It may be the case that postcss-svgo has some inbuilt functionality for decoding data URIs, but if it is just calling `decodeURIComponent` on the whole data URI string, it may not be doing it correctly.

Regardless, if it is determined at any point that the URL is not valid, then the error message reported by cssnano should say that, rather than saying "Non-whitespace before first tag." when that's not the case.

ludofischer: Thank you for the WhattWG specs. I was looking at IETF

PR #1970, opened 2026-09-07
fix: update postcss-calc

ludofischer:


packages/cssnano/test/fixtures.js
```diff
@@ -355,5 +355,5 @@ test(
 
 test(
   'should remove leading zeroes from reduced calc values',
-  processCss(`.box { margin: calc(-.5 * 1rem); }`, `.box{margin:-.5rem}`)
+  processCss(`.box { margin: calc(-.5 * 1rem); }`, `.box{margin:calc(-.5rem)}`)
```
vincentbernat, line 358: This looks like a regression.

ludofischer: This is done on purpose, to avoid producing an invalid value where negative values are not allowed see the spec https://drafts.csswg.org/css-values/#example-6652ed9a . So for margin `calc(-.5rem)` and  `-.5rem` are equivalent, but for border-width `calc(-.5rem)` is supposed to be equivalent to `0` while `-.5rem` would be ignored because it is invalid. It is possible to maintain a list of all properties which allow negative values and only minify then, but it looked a bit brittle so for safety `postcss-calc` does not simplify a single negative value further, unless you pass the option https://github.com/postcss/postcss-calc#unwrapsinglenegativenumber-default-false

vincentbernat: Thanks for the precision!

